PR #1, opened 2016-06-02
Improve chromeext annotations

attfarhan:
WIP. Fix bugs related to chrome extension annotations.  

Implemented: 
- Adds refreshVCS and automatically getting the latest annotations.  
- Fixes bug where jump-to-def will automatically revert to master branch.  Jump-to-def will now stay within the currently selected branch. 
- Adds revision to the link to trigger builds when search 404s
- Adds checks to fix syntax highlighting issues when trying to solve the split up annotations in go imports because of quote mark span tags.
- Adds checks to ensure '<span>' appearing anywhere in comments or in the code won't break annotations. 

##### Reviewer tasks

##### Test plan
Build this locally and test the refresh and annotations update by pushing to a repo and seeing that broken annotations are not added.  Once the build on Sourcegraph is finished, see that correct annotations are added. 

Try jump-to-def on a branch that isn't master. https://github.com/alexo/wro4j/blob/1.7.x/wro4j-core/src/main/java/ro/isdc/wro/http/WroFilter.java#sourcegraph&def=JavaArtifact/ro.isdc.wro4j/wro4j-core/-/ro/isdc/wro/http/WroFilter:type/createConfiguration&L135-143 

View code that has <span> comments and see that anns don't break. https://github.com/attfarhan/mux/blob/master/route.go (build for this is queued, but should work when it's finished).  

Click on an external def from a branch that isn't master and make sure we don't 404, and we get to the def on the default branch of the repository.

### DIFF
--- a/client/browser-ext/app/containers/App.js
+++ b/client/browser-ext/app/containers/App.js
@@ -110,7 +110,7 @@ export default class App extends Component {
 							<h3 styleName="list-item-empty">Searching...</h3>
 						}
 						{srclibDataVersionFetch && srclibDataVersionFetch.response && srclibDataVersionFetch.response.status === 404 &&
-							<h3 styleName="list-item-empty">404 Not Found: This repository has not been indexed by Sourcegraph.<br/><a href={`https://sourcegraph.com/${this.props.repo}`}>Let Sourcegraph index this repository.</a></h3>
+							<h3 styleName="list-item-empty">404 Not Found: This repository has not been indexed by Sourcegraph.<br/><a href={`https://sourcegraph.com/${this.props.repo}@${this.props.rev}`}>Let Sourcegraph index this repository.</a></h3>
 						}
 						{srclibDataVersionFetch && srclibDataVersionFetch.response && srclibDataVersionFetch.response.status === 401 &&
 							<h3 styleName="list-item-empty">401 Unauthorized: Log in to Sourcegraph to search private code</h3>
--- a/client/browser-ext/app/containers/App.js
+++ b/client/browser-ext/app/containers/App.js
@@ -110,7 +110,7 @@ export default class App extends Component {
 							<h3 styleName="list-item-empty">Searching...</h3>
 						}
 						{srclibDataVersionFetch && srclibDataVersionFetch.response && srclibDataVersionFetch.response.status === 404 &&
-							<h3 styleName="list-item-empty">404 Not Found: This repository has not been indexed by Sourcegraph.<br/><a href={`https://sourcegraph.com/${this.props.repo}@${this.props.rev}`}>Let Sourcegraph index this repository.</a></h3>
+							<h3 styleName="list-item-empty">404 Not Found: This repository (or revision) has not been indexed by Sourcegraph.<br/><a href={`https://sourcegraph.com/${this.props.repo}@${this.props.rev}`}>Let Sourcegraph index this repository.</a></h3>
 						}
 						{srclibDataVersionFetch && srclibDataVersionFetch.response && srclibDataVersionFetch.response.status === 401 &&
 							<h3 styleName="list-item-empty">401 Unauthorized: Log in to Sourcegraph to search private code</h3>
--- a/client/browser-ext/chrome/extension/annotations.js
+++ b/client/browser-ext/chrome/extension/annotations.js
@@ -74,8 +74,23 @@ function traverseDOM(annsByStartByte, annsByEndByte){
 			if (children[j].nodeType === Node.TEXT_NODE){
 				childNodeChars = children[j].nodeValue.split("");
 			} else {
-				if (children[j].children.length > 1 && children[j].children[0].className === "pl-pds") {
-					children[j].innerHTML = _.escape(children[j].innerText);
+				// Quotation marks on GitHub are given their own span tags.
+				// This messes up anns for go imports. We handle this
+				// by combining all the child span tags when we see child spans
+				// with "pl-pds", the class for quotes, and there aren't any other
+				// tags between them (tags in between needed for highlighting).
+				if (children[j].children.length > 1) {
+					let goImport = true;
+					for (let k = 0; k < children[j].children.length; k++) {
+						if(children[j].children[k].className !== "pl-pds") {
+
+							goImport = false
+							break;
+						}
+					}
+					if (goImport) {
+						children[j].innerHTML = _.escape(children[j].innerText);
+					}
 				}
 				childNodeChars = (children[j].outerHTML).split("");
 			}
--- a/client/browser-ext/chrome/extension/annotations.js
+++ b/client/browser-ext/chrome/extension/annotations.js
@@ -75,24 +75,23 @@ function traverseDOM(annsByStartByte, annsByEndByte){
 				childNodeChars = children[j].nodeValue.split("");
 			} else {
 				// Quotation marks on GitHub are given their own span tags.
-				// This messes up anns for go imports. We handle this
-				// by combining all the child span tags when we see child spans
-				// with "pl-pds", the class for quotes, and there aren't any other
+				// This messes up anns for go imports and other strings with quote marks.
+				// We handle this by combining all the child span tags when we see child
+				// spans with "pl-pds", the class for quotes, and there aren't any other
 				// tags between them (tags in between needed for highlighting).
 				if (children[j].children.length > 1) {
-					let goImport = true;
+					let stringWithQuotes = true;
 					for (let k = 0; k < children[j].children.length; k++) {
-						if(children[j].children[k].className !== "pl-pds") {
-
-							goImport = false
+						if (children[j].children[k].className !== "pl-pds") {
+							stringWithQuotes = false
 							break;
 						}
 					}
-					if (goImport) {
+					if (stringWithQuotes) {
 						children[j].innerHTML = _.escape(children[j].innerText);
 					}
 				}
-				childNodeChars = (children[j].outerHTML).split("");
+				childNodeChars = children[j].outerHTML.split("");
 			}
 
 			let consumingSpan = false;
--- a/client/browser-ext/chrome/extension/annotations.js
+++ b/client/browser-ext/chrome/extension/annotations.js
@@ -70,15 +70,14 @@ function traverseDOM(annsByStartByte, annsByEndByte){
 
 		for (let j = 0; j < children.length; j++) {
 			let childNodeChars; // the "inner-stuff" of the code cell
-			if (children[j].className) {
-				if (annsByStartByte[startByte]) {
-					annsByStartByte[startByte].className = children[j].className;
-				}
-			}
+
 			if (children[j].nodeType === Node.TEXT_NODE){
 				childNodeChars = children[j].nodeValue.split("");
 			} else {
-				childNodeChars = _.unescape(children[j].outerHTML).split("");
+				if (children[j].children.length > 1 && children[j].children[0].className === "pl-pds") {
+					children[j].innerHTML = _.escape(children[j].innerText);
+				}
+				childNodeChars = (children[j].outerHTML).split("");
 			}
 
 			let consumingSpan = false;
@@ -89,14 +88,21 @@ function traverseDOM(annsByStartByte, annsByEndByte){
 				if (childNodeChars[k] === "<" && (childNodeChars.slice(k, k+5).join("") === "<span" || childNodeChars.slice(k, k+6).join("") === "</span")) {
 					consumingSpan = true;
 				}
-
 				if (!consumingSpan){
-					output += next(childNodeChars[k], startByte, annsByStartByte, annsByEndByte)
-					startByte += utf8.encode(childNodeChars[k]).length
-				}
-				else {
+					// Case to handle if < or > appears, so that we don't "consume" or make span tags in the code disappear
+					// This will not break if "&lt;" or "&gt;" appear because chars are escaped.
+					if (childNodeChars[k] === "&" && (((childNodeChars.slice(k, k+4).join("")) === ("&gt;")) || (childNodeChars.slice(k, k+4).join("") === ("&lt;")))) {
+						output += next(childNodeChars.slice(k, k+4).join(""), startByte, annsByStartByte, annsByEndByte);
+						k += childNodeChars.slice(k, k+4).join("").length-1;
+						startByte += utf8.encode(childNodeChars[k]).length;
+					}
+					else {
+						output += next(childNodeChars[k], startByte, annsByStartByte, annsByEndByte);
+						startByte += utf8.encode(childNodeChars[k]).length;
+					}
+				} else {
 					// when we are consuming the <span> element, don't increment startByte
-					output += childNodeChars[k]
+					output += childNodeChars[k];
 				}
 
 				if (childNodeChars[k] === ">" && consumingSpan) {
@@ -121,8 +127,6 @@ function traverseDOM(annsByStartByte, annsByEndByte){
 // into itself or wraps the character in a starting/ending anchor tag
 function next(c, byteCount, annsByStartByte, annsByEndByte) {
 	let matchDetails = annsByStartByte[byteCount];
-
-	c = _.escape(c); // IMPORTANT: escape all markup injected in HTML
 
 	// if there is a match
 	if (!annotating && matchDetails) {
--- a/client/browser-ext/chrome/extension/annotations.js
+++ b/client/browser-ext/chrome/extension/annotations.js
@@ -70,16 +70,14 @@ function traverseDOM(annsByStartByte, annsByEndByte){
 
 		for (let j = 0; j < children.length; j++) {
 			let childNodeChars; // the "inner-stuff" of the code cell
-
+			if (children[j].className) {
+				if (annsByStartByte[startByte]) {
+					annsByStartByte[startByte].className = children[j].className;
+				}
+			}
 			if (children[j].nodeType === Node.TEXT_NODE){
 				childNodeChars = children[j].nodeValue.split("");
 			} else {
-				if (children[j].children.length > 1) {
-					// HACK: combine children spans, e.g. for a quoted token
-					// there may be 3 spans, one for each quote and one
-					// for the token iteself
-					children[j].innerHTML = _.escape(children[j].innerText);
-				}
 				childNodeChars = _.unescape(children[j].outerHTML).split("");
 			}
 
@@ -134,10 +132,14 @@ function next(c, byteCount, annsByStartByte, annsByEndByte) {
 		const url = defIsOnGitHub ? urlToDef(matchDetails.URL) : `https://sourcegraph.com${matchDetails.URL}`;
 		if (!url) return c;
 
-		const insert = `<a href="${url}" ${defIsOnGitHub ? "data-sourcegraph-ref" : "target=tab"} data-src="https://sourcegraph.com${matchDetails.URL}" class=${styles.sgdef}>${c}`;
+		const insert = annsByStartByte[byteCount].className ? `<a href="${url}" ${defIsOnGitHub ? "data-sourcegraph-ref" : "target=tab"} data-src="https://sourcegraph.com${matchDetails.URL}" class=${styles.sgdef}><span class="${annsByStartByte[byteCount].className}">${c}`
+		: `<a href="${url}" ${defIsOnGitHub ? "data-sourcegraph-ref" : "target=tab"} data-src="https://sourcegraph.com${matchDetails.URL}" class=${styles.sgdef}>${c}`;
 
 		// off-by-one case
 		if (annsByStartByte[byteCount].EndByte - annsByStartByte[byteCount].StartByte === 1) {
+			if annsByStartByte[byteCount].className {
+				return `${insert}</span></a>`
+			}
 			return `${insert}</a>`;
 		}
 
--- a/client/browser-ext/chrome/extension/annotations.js
+++ b/client/browser-ext/chrome/extension/annotations.js
@@ -132,14 +132,10 @@ function next(c, byteCount, annsByStartByte, annsByEndByte) {
 		const url = defIsOnGitHub ? urlToDef(matchDetails.URL) : `https://sourcegraph.com${matchDetails.URL}`;
 		if (!url) return c;
 
-		const insert = annsByStartByte[byteCount].className ? `<a href="${url}" ${defIsOnGitHub ? "data-sourcegraph-ref" : "target=tab"} data-src="https://sourcegraph.com${matchDetails.URL}" class=${styles.sgdef}><span class="${annsByStartByte[byteCount].className}">${c}`
-		: `<a href="${url}" ${defIsOnGitHub ? "data-sourcegraph-ref" : "target=tab"} data-src="https://sourcegraph.com${matchDetails.URL}" class=${styles.sgdef}>${c}`;
+		const insert = `<a href="${url}" ${defIsOnGitHub ? "data-sourcegraph-ref" : "target=tab"} data-src="https://sourcegraph.com${matchDetails.URL}" class=${styles.sgdef}>${c}`;
 
 		// off-by-one case
 		if (annsByStartByte[byteCount].EndByte - annsByStartByte[byteCount].StartByte === 1) {
-			if annsByStartByte[byteCount].className {
-				return `${insert}</span></a>`
-			}
 			return `${insert}</a>`;
 		}
 
--- a/client/browser-ext/chrome/extension/annotations.js
+++ b/client/browser-ext/chrome/extension/annotations.js
@@ -172,16 +172,18 @@ function next(c, byteCount, annsByStartByte, annsByEndByte) {
 }
 
 export const defaultBranchCache = {};
+export const fetchingDefaultBranchCache = {};
 function cacheDefaultBranch(annURL) {
 	let annURLsplit = [annURL.split("/")[1], annURL.split("/")[2], annURL.split("/")[3]];
 	let repo = annURLsplit.join("/");
-	if (defaultBranchCache[repo]) {
-		console.log("CACHED")
+	if (fetchingDefaultBranchCache[repo]) {
 		return;
 	}
+	fetchingDefaultBranchCache[repo] = true;
 	fetch(`https://sourcegraph.com/.api/repos/${repo}`)
 		.then((response) => {
 			defaultBranchCache[repo] = response.DefaultBranch;
+			fetchingDefaultBranchCache[repo] = false;
 		})
 		.catch((err) => console.log("Error getting default branch"));
 }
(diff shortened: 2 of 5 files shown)


sqs:
@attfarhan Because the event listener is listening on click events on ALL child elements of document (https://github.com/sourcegraph/sourcegraph/blob/37d7727760b8a384f690c8bf1593c843c6c32b88/client/browser-ext/chrome/extension/inject.js#L80), `ev.target` and `ev.currentTarget` will be the deepest node that was clicked (basically).

So now that there can be spans inside the links, we need to do some node traversal.

Specifically, at https://github.com/sourcegraph/sourcegraph/blob/37d7727760b8a384f690c8bf1593c843c6c32b88/client/browser-ext/chrome/extension/inject.js#L110, we need to have some code that tries to climb the DOM tree to look for the `data-sourcegraph-ref` element. But since the deepest nesting underneath a `data-sourcegraph-ref` element is a single level, we can just look to see whether the event target or its parent has `data-sourcegraph-ref`; we don't need to iterate. So this should work:

```
	// ...
	_clickRef(ev) {
		if (typeof ev.target.dataset.sourcegraphRef !== "undefined" || (ev.target.parentNode && typeof ev.target.parentNode.dataset.sourcegraphRef !== "undefined")) {
	// ...
```

attfarhan:
This has been updated.  @sqs I didn't add the case above since I changed the way I handled the issue.

client/browser-ext/chrome/extension/inject.js
```diff
@@ -132,7 +132,9 @@ class InjectApp extends React.Component {
 		const urlsplit = loc.pathname.slice(1).split("/");
 		let user = urlsplit[0];
 		let repo = urlsplit[1]
-		let rev = "master"
+		// We scrape the current branch and set rev to it so we stay on the same branch when doing jump-to-def
+		let currBranch = document.getElementsByClassName('select-menu-button js-menu-target css-truncate')[0] ? document.getElementsByClassName('select-menu-button js-menu-target css-truncate')[0].title : "master"
```
sqs: Extract the `document.getElementsByClassName('select-menu-button js-menu-target css-truncate')` expression into a var, since it's quite long.

client/browser-ext/chrome/extension/inject.js
```diff
@@ -173,8 +175,19 @@ class InjectApp extends React.Component {
 		this.addSearchButton();
 
 		let {user, repo, rev, path, defPath} = this.parseURL();
+		// This scrapes the latest commit ID and updates rev to the latest commit so we are never injecting
+		// outdated annotations.  If there is a new commit, srclib-data-version will return a 404, but the
+		// refresh endpoint will update the version and the annotations will be up to date once the new build succeeds
+		let latestRev = document.getElementsByClassName("js-permalink-shortcut")[0] ? document.getElementsByClassName("js-permalink-shortcut")[0].href.split("/")[6] : rev
+		// TODO: Branches that are not built on Sourcegraph will not get annotations, need to trigger
+		let currBranch = document.getElementsByClassName('select-menu-button js-menu-target css-truncate')[0] ? document.getElementsByClassName('select-menu-button js-menu-target css-truncate')[0].title : "master"
```
sqs: It's dangerous to ever hardcode "master". If a repo's default branch is not master, then the hardcoding will always be incorrect. In what cases can we not determine the branch name and need to fall back to "master"?

client/browser-ext/chrome/extension/inject.js
```diff
@@ -132,7 +132,9 @@ class InjectApp extends React.Component {
 		const urlsplit = loc.pathname.slice(1).split("/");
 		let user = urlsplit[0];
 		let repo = urlsplit[1]
-		let rev = "master"
+		// We scrape the current branch and set rev to it so we stay on the same branch when doing jump-to-def
+		let currBranch = document.getElementsByClassName('select-menu-button js-menu-target css-truncate')[0] ? document.getElementsByClassName('select-menu-button js-menu-target css-truncate')[0].title : "master"
+		let rev = currBranch
```
sqs: semicolon

sqs:
Made a few comments. Overall looks great! Please get @rothfels to run through the test plan tomorrow before updating.

client/browser-ext/chrome/extension/inject.js
```diff
@@ -132,7 +132,10 @@ class InjectApp extends React.Component {
 		const urlsplit = loc.pathname.slice(1).split("/");
 		let user = urlsplit[0];
 		let repo = urlsplit[1]
-		let rev = "master"
+		// We scrape the current branch and set rev to it so we stay on the same branch when doing jump-to-def
+		let branchSelectorButton = document.getElementsByClassName('select-menu-button js-menu-target css-truncate')[0]
+		let currBranch = branchSelectorButton ? branchSelectorButton.title : "master";
```
attfarhan: @sqs I put "master" as a placeholder for when the user is is not on a blob or repo view on GitHub.  It'll show an error because it can't find property title of undefined, and this method is called in refreshState, which is called on every page movement.

rothfels:
The "Find usages" link needs to carry the branch (it does not): http://cl.ly/1G0E3Z100K2X

rothfels:
The first step of the test plan is broken, after I add new code: http://cl.ly/3x3e1F2M3v1c

rothfels:
Hmm...never mind on the latest comment, I think I was running two versions of the chrome extension simultaneously.

rothfels:
Ok @attfarhan test plan looks good, let's just update the find usages link to carry branch name.

client/browser-ext/app/containers/App.js
```diff
@@ -110,7 +110,7 @@ export default class App extends Component {
 							<h3 styleName="list-item-empty">Searching...</h3>
 						}
 						{srclibDataVersionFetch && srclibDataVersionFetch.response && srclibDataVersionFetch.response.status === 404 &&
-							<h3 styleName="list-item-empty">404 Not Found: This repository has not been indexed by Sourcegraph.<br/><a href={`https://sourcegraph.com/${this.props.repo}`}>Let Sourcegraph index this repository.</a></h3>
+							<h3 styleName="list-item-empty">404 Not Found: This repository has not been indexed by Sourcegraph.<br/><a href={`https://sourcegraph.com/${this.props.repo}@${this.props.rev}`}>Let Sourcegraph index this repository.</a></h3>
```
rothfels: Remove the @rev here, I don't think it's necessary.

client/browser-ext/chrome/extension/annotations.js
```diff
@@ -74,13 +74,25 @@ function traverseDOM(annsByStartByte, annsByEndByte){
 			if (children[j].nodeType === Node.TEXT_NODE){
 				childNodeChars = children[j].nodeValue.split("");
 			} else {
+				// Quotation marks on GitHub are given their own span tags.
+				// This messes up anns for go imports. We handle this
```
rothfels: Not just for go imports, for all kinds of things, right?

client/browser-ext/chrome/extension/annotations.js
```diff
@@ -74,13 +74,25 @@ function traverseDOM(annsByStartByte, annsByEndByte){
...
+				// with "pl-pds", the class for quotes, and there aren't any other
+				// tags between them (tags in between needed for highlighting).
 				if (children[j].children.length > 1) {
-					// HACK: combine children spans, e.g. for a quoted token
-					// there may be 3 spans, one for each quote and one
-					// for the token iteself
-					children[j].innerHTML = _.escape(children[j].innerText);
+					let goImport = true;
+					for (let k = 0; k < children[j].children.length; k++) {
+						if(children[j].children[k].className !== "pl-pds") {
```
rothfels: Space after `if`

client/browser-ext/chrome/extension/annotations.js
```diff
@@ -74,13 +74,25 @@ function traverseDOM(annsByStartByte, annsByEndByte){
...
+							goImport = false
+							break;
+						}
+					}
+					if (goImport) {
+						children[j].innerHTML = _.escape(children[j].innerText);
+					}
 				}
-				childNodeChars = _.unescape(children[j].outerHTML).split("");
+				childNodeChars = (children[j].outerHTML).split("");
```
rothfels: Remove parens around `children[j].outerHTML` (unnecessary)

client/browser-ext/chrome/extension/annotations.js
```diff
@@ -91,14 +103,21 @@ function traverseDOM(annsByStartByte, annsByEndByte){
...
 				}
-
 				if (!consumingSpan){
-					output += next(childNodeChars[k], startByte, annsByStartByte, annsByEndByte)
-					startByte += utf8.encode(childNodeChars[k]).length
-				}
-				else {
+					// Case to handle if < or > appears, so that we don't "consume" or make span tags in the code disappear
+					// This will not break if "&lt;" or "&gt;" appear because chars are escaped.
+					if (childNodeChars[k] === "&" && (((childNodeChars.slice(k, k+4).join("")) === ("&gt;")) || (childNodeChars.slice(k, k+4).join("") === ("&lt;")))) {
```
rothfels: The conditional is long and hard to parse. Can you decompose it into a private helper function?

client/browser-ext/app/containers/App.js
sqs: Going to the rev will trigger a build for that rev so it is necessary. But the text should make it clear that that specific revision/branch hasn't been indexed, not the whole repo.

client/browser-ext/chrome/extension/inject.js
rothfels: @attfarhan isn't the current branch state always in the URL? Why do we have to parse it from the rev selector dom element state? Is it because there can be an unspecified revision in the URL for the default branch? If so, a comment here would help.

client/browser-ext/chrome/extension/inject.js
```diff
@@ -132,7 +132,10 @@ class InjectApp extends React.Component {
 		const urlsplit = loc.pathname.slice(1).split("/");
 		let user = urlsplit[0];
 		let repo = urlsplit[1]
-		let rev = "master"
+		// We scrape the current branch and set rev to it so we stay on the same branch when doing jump-to-def
+		let branchSelectorButton = document.getElementsByClassName('select-menu-button js-menu-target css-truncate')[0]
```
rothfels: Use double quotes for all strings, like we have everywhere else.

client/browser-ext/chrome/extension/inject.js
```diff
@@ -173,8 +176,20 @@ class InjectApp extends React.Component {
...
 
 		let {user, repo, rev, path, defPath} = this.parseURL();
+		// This scrapes the latest commit ID and updates rev to the latest commit so we are never injecting
+		// outdated annotations.  If there is a new commit, srclib-data-version will return a 404, but the
+		// refresh endpoint will update the version and the annotations will be up to date once the new build succeeds
+		let latestRev = document.getElementsByClassName("js-permalink-shortcut")[0] ? document.getElementsByClassName("js-permalink-shortcut")[0].href.split("/")[6] : rev
+		// TODO: Branches that are not built on Sourcegraph will not get annotations, need to trigger
+		let branchSelectorButton = document.getElementsByClassName('select-menu-button js-menu-target css-truncate')[0];
+		let currBranch = branchSelectorButton ? branchSelectorButton.title : rev;
+		if (rev !== latestRev) rev = latestRev;
```
rothfels: I don't think you need the conditional `if`, do you?

client/browser-ext/chrome/extension/inject.js
```diff
@@ -173,8 +176,20 @@ class InjectApp extends React.Component {
 		this.addSearchButton();
 
 		let {user, repo, rev, path, defPath} = this.parseURL();
+		// This scrapes the latest commit ID and updates rev to the latest commit so we are never injecting
+		// outdated annotations.  If there is a new commit, srclib-data-version will return a 404, but the
+		// refresh endpoint will update the version and the annotations will be up to date once the new build succeeds
+		let latestRev = document.getElementsByClassName("js-permalink-shortcut")[0] ? document.getElementsByClassName("js-permalink-shortcut")[0].href.split("/")[6] : rev
+		// TODO: Branches that are not built on Sourcegraph will not get annotations, need to trigger
+		let branchSelectorButton = document.getElementsByClassName('select-menu-button js-menu-target css-truncate')[0];
```
rothfels: You have this code in two places; decompose into a function.

client/browser-ext/chrome/extension/inject.js
```diff
@@ -173,8 +176,20 @@ class InjectApp extends React.Component {
...
+		let currBranch = branchSelectorButton ? branchSelectorButton.title : rev;
+		if (rev !== latestRev) rev = latestRev;
+		if (currBranch !== "master") rev = currBranch;
```
rothfels: I'm confused by what all this code is supposed to be doing. Can you explain to me in person?

client/browser-ext/app/containers/App.js
rothfels: Ahh, got it. Thanks.

attfarhan:
@rothfels, this is updated.

client/browser-ext/app/containers/App.js
rothfels: @attfarhan can you clarify the copy here? `This repository (or revision) has not been indexed by Sourcegraph`

client/browser-ext/chrome/extension/annotations.js
```diff
@@ -119,13 +137,14 @@ function traverseDOM(annsByStartByte, annsByEndByte){
 	}
 }
 
+function isSpanTagString(childNodeChars, k) {
```
rothfels: k is a bit contextless here -- what is k?

I would change the name to `isStartOfSpanTag` and name the second arg `idx`

client/browser-ext/chrome/extension/annotations.js
```diff
@@ -119,13 +137,14 @@ function traverseDOM(annsByStartByte, annsByEndByte){
...
+	return (childNodeChars[k] === "&" && (((childNodeChars.slice(k, k+4).join("")) === ("&gt;")) || (childNodeChars.slice(k, k+4).join("") === ("&lt;"))))
+}
+
 // next is a helper method for traverseDOM which transforms a character
 // into itself or wraps the character in a starting/ending anchor tag
 function next(c, byteCount, annsByStartByte, annsByEndByte) {
 	let matchDetails = annsByStartByte[byteCount];
-
-	c = _.escape(c); // IMPORTANT: escape all markup injected in HTML
```
rothfels: Why did this get removed?

client/browser-ext/chrome/extension/annotations.js
```diff
@@ -140,20 +159,35 @@ function next(c, byteCount, annsByStartByte, annsByEndByte) {
...
 	if (annotating && annsByEndByte[byteCount + 1]) {
 		annotating = false;
 		return `${c}</a>`;
 	}
-
 	return c;
 }
 
+export const defaultBranchCache = {};
+export const fetchingDefaultBranchCache = {};
```
rothfels: Can you leave a comment about what purpose this serves?

client/browser-ext/chrome/extension/annotations.js
```diff
@@ -140,20 +159,35 @@ function next(c, byteCount, annsByStartByte, annsByEndByte) {
...
+function cacheDefaultBranch(annURL) {
+	let annURLsplit = [annURL.split("/")[1], annURL.split("/")[2], annURL.split("/")[3]];
```
rothfels: A couple issues:
- You don't need to call `annURL.split("/")` 3 times; once will suffice
- What if the annURL doesn't have the form `github.com/user/repo/`? Maybe just leave a comment in here that we assume this is what structure the URL has so the next developer isn't confused. (And that if we can't fetch the default branch, we'll use "master" as a default so it's not the end of the world.

client/browser-ext/chrome/extension/annotations.js
```diff
@@ -140,20 +159,35 @@ function next(c, byteCount, annsByStartByte, annsByEndByte) {
...
+	if (fetchingDefaultBranchCache[repo]) {
+		return;
+	}
+	fetchingDefaultBranchCache[repo] = true;
+	fetch(`https://sourcegraph.com/.api/repos/${repo}`)
+		.then((response) => {
+			defaultBranchCache[repo] = response.DefaultBranch;
+			fetchingDefaultBranchCache[repo] = false;
+		})
+		.catch((err) => console.log("Error getting default branch"));
```
rothfels: This is an unreachable code path; fetch will never throw an error, the only way to get here would be if the `.then` chain did a `throw err`.

client/browser-ext/chrome/extension/annotations.js
```diff
@@ -140,20 +159,35 @@ function next(c, byteCount, annsByStartByte, annsByEndByte) {
...
+function cacheDefaultBranch(annURL) {
+	let annURLsplit = [annURL.split("/")[1], annURL.split("/")[2], annURL.split("/")[3]];
+	let repo = annURLsplit.join("/");
+	if (fetchingDefaultBranchCache[repo]) {
+		return;
+	}
+	fetchingDefaultBranchCache[repo] = true;
+	fetch(`https://sourcegraph.com/.api/repos/${repo}`)
+		.then((response) => {
+			defaultBranchCache[repo] = response.DefaultBranch;
```
rothfels: You only want to execute this logic if the status code is 200.

client/browser-ext/chrome/extension/annotations.js
```diff
@@ -140,20 +159,35 @@ function next(c, byteCount, annsByStartByte, annsByEndByte) {
...
+			fetchingDefaultBranchCache[repo] = false;
```
rothfels: You always want to set this to false (regardless of the response status code).

client/browser-ext/chrome/extension/annotations.js
rothfels: Do you want to just store "master" here if there's a non 200 status code?

client/browser-ext/chrome/extension/annotations.js
```diff
@@ -140,20 +159,35 @@ function next(c, byteCount, annsByStartByte, annsByEndByte) {
...
+export const defaultBranchCache = {};
+export const fetchingDefaultBranchCache = {};
+function cacheDefaultBranch(annURL) {
+	let annURLsplit = [annURL.split("/")[1], annURL.split("/")[2], annURL.split("/")[3]];
+	let repo = annURLsplit.join("/");
+	if (fetchingDefaultBranchCache[repo]) {
+		return;
+	}
+	fetchingDefaultBranchCache[repo] = true;
+	fetch(`https://sourcegraph.com/.api/repos/${repo}`)
```
rothfels: You're shouldn't fetch if the default branch has already been fetched (i.e. `defaultBranchCache[repo]` is defined).

client/browser-ext/chrome/extension/annotations.js
rothfels: Not saying you should make it reachable, just that you should never check in code that is unreachable. :)

Besides, you don't need to log the error, since chrome will automatically log network request errors (non-20x status codes)

client/browser-ext/chrome/extension/inject.js
```diff
@@ -111,10 +113,13 @@ class InjectApp extends React.Component {
 		if (typeof ev.target.dataset.sourcegraphRef !== "undefined") {
 			let urlProps = this.parseURL({pathname: ev.target.pathname, hash: ev.target.hash});
 			urlProps.repo = `github.com/${urlProps.user}/${urlProps.repo}`;
+			// HACK: We need the default branch for external defs. This gets it for us, but we'd
```
rothfels: I don't think this comment is relevant or necessary, you can remove it.

client/browser-ext/chrome/extension/inject.js
```diff
@@ -132,7 +137,9 @@ class InjectApp extends React.Component {
 		const urlsplit = loc.pathname.slice(1).split("/");
 		let user = urlsplit[0];
 		let repo = urlsplit[1]
-		let rev = "master"
+		// We scrape the current branch and set rev to it so we stay on the same branch when doing jump-to-def
```
rothfels: This is only true if we are if we are navigating within a repo, right? In which case, can't the rev be parsed out of the URL? And if we're navigating to an external repo, don't we want to use the `defaultBranchCache`?

client/browser-ext/chrome/extension/annotations.js
attfarhan: We no longer unescape in traverseDOM, we do this so we can read html in the code or comments as their character entities so we can make sure <span> tags in the code don't mess up the annotations.

client/browser-ext/chrome/extension/inject.js
```diff
@@ -155,6 +162,10 @@ class InjectApp extends React.Component {
 		return info;
 	}
 
+	branchSelectorButton() {
```
rothfels: Name this function `getBranchSelectorButton`.

client/browser-ext/chrome/extension/inject.js
```diff
@@ -173,8 +184,17 @@ class InjectApp extends React.Component {
 		this.addSearchButton();
 
 		let {user, repo, rev, path, defPath} = this.parseURL();
+		// This scrapes the latest commit ID and updates rev to the latest commit so we are never injecting
+		// outdated annotations.  If there is a new commit, srclib-data-version will return a 404, but the
+		// refresh endpoint will update the version and the annotations will be up to date once the new build succeeds
+		let latestRev = document.getElementsByClassName("js-permalink-shortcut")[0] ? document.getElementsByClassName("js-permalink-shortcut")[0].href.split("/")[6] : rev;
+		// TODO: Branches that are not built on Sourcegraph will not get annotations, need to trigger
+		if (rev !== latestRev) rev = latestRev;
```
rothfels: Why is the `if` necessary -- does it add anything?

client/browser-ext/chrome/extension/inject.js
```diff
@@ -173,8 +184,17 @@ class InjectApp extends React.Component {
 		this.addSearchButton();
 
 		let {user, repo, rev, path, defPath} = this.parseURL();
+		// This scrapes the latest commit ID and updates rev to the latest commit so we are never injecting
+		// outdated annotations.  If there is a new commit, srclib-data-version will return a 404, but the
+		// refresh endpoint will update the version and the annotations will be up to date once the new build succeeds
+		let latestRev = document.getElementsByClassName("js-permalink-shortcut")[0] ? document.getElementsByClassName("js-permalink-shortcut")[0].href.split("/")[6] : rev;
```
rothfels: Just making sure I understand this correctly -- is it replacing a branch name with a 64char CommitID?

client/browser-ext/chrome/extension/inject.js
```diff
@@ -217,11 +237,12 @@ class InjectApp extends React.Component {
 	_directURLToDef({repo, rev, defPath, def}) {
 		const defObj = def ? def.content[keyFor(repo, rev, defPath)] : null;
 		if (defObj) {
-			const pathname = `/${repo.replace("github.com/", "")}/blob/${rev}/${defObj.File}`;
+			let defBranch = defaultBranchCache[repo]
+			const pathname = (repo === this.props.repo) ? `/${repo.replace("github.com/", "")}/blob/${rev}/${defObj.File}` : `/${repo.replace("github.com/", "")}/blob/${defBranch}/${defObj.File}`;
```
rothfels: This ternary expression is long and hard to read. How about keeping the code as it was before, but subbing this line:

`if (repo !== this.props.repo) rev = defBranch || "master"`

client/browser-ext/chrome/extension/inject.js
```diff
@@ -217,11 +237,12 @@ class InjectApp extends React.Component {
 	_directURLToDef({repo, rev, defPath, def}) {
 		const defObj = def ? def.content[keyFor(repo, rev, defPath)] : null;
 		if (defObj) {
-			const pathname = `/${repo.replace("github.com/", "")}/blob/${rev}/${defObj.File}`;
+			let defBranch = defaultBranchCache[repo]
```
rothfels: You need a default, like `defaultBranchCache[repo] || "master"`

client/browser-ext/chrome/extension/annotations.js
rothfels: Ok

client/browser-ext/chrome/extension/inject.js
attfarhan: Yes.  If we leave the branch, the srclib-data-version endpoint will not necessarily return a 404 since we do have a srclib-data-version for that branch.  If we sub it with the commit ID, it will return a 404 and we don't allow broken annotations to get injected.

client/browser-ext/chrome/extension/annotations.js
attfarhan: It actually does get reached when this fetch returns a 404.  Also, when I remove this, I get a bunch of unresolved promise errors.

client/browser-ext/chrome/extension/inject.js
attfarhan: Yeah, this makes sense.  Made the switch, but to `if (repo !== this.props.repo) rev = defaultBranchCache[repo] || "master"`

client/browser-ext/chrome/extension/annotations.js
```diff
@@ -140,20 +159,46 @@ function next(c, byteCount, annsByStartByte, annsByEndByte) {
...
+	let annRepo = [annURLsplit[1], annURLsplit[2], annURLsplit[3]]
+	let repo = annRepo.join("/");
+	if (fetchingDefaultBranchCache[repo]) {
+		return;
+	}
+	if (!defaultBranchCache[repo]) {
+		fetchingDefaultBranchCache[repo] = true;
+		fetch(`https://sourcegraph.com/.api/repos/${repo}`)
+			.then((response) => {
+				if (response.ok) {
```
rothfels: you can condense the if/else block into the following:

```
defaultBranchCache[repo] = response.ok ? response.DefaultBranch : "master";
fetchingDefaultBranchCache[repo] = false;
```

^^ cleaner amirite? :)

client/browser-ext/chrome/extension/annotations.js
```diff
@@ -140,20 +159,46 @@ function next(c, byteCount, annsByStartByte, annsByEndByte) {
...
+					defaultBranchCache[repo] = response.DefaultBranch;
+					fetchingDefaultBranchCache[repo] = false;
+				} else {
+					// Fall back onto master if we can't get the default branch
+					defaultBranchCache[repo] = "master";
+					fetchingDefaultBranchCache[repo] = false;
+				}
+			})
+			.catch((err) => console.log("Error getting default branch"))
```
rothfels: This is still unreachable; you should remove it.

attfarhan: @rothfels it does get reached when the fetch returns a 404. (screenshot: http://cl.ly/1d3N1y0W1G0R from here: https://github.com/alexo/wro4j/blob/1.7.x/wro4j-core/src/main/java/ro/isdc/wro/http/WroFilter.java) 

When I remove it, I get a bunch of unhandled promise rejections.

rothfels:
@attfarhan this LGTM, can you have @uforic take a final QA pass before we merge this?

uforic:
Went through test plan, tried out this morning in a variety of use cases, LGTM.

PR #2, opened 2016-06-02
global_refs: Prevent duplicate work for indexing

keegancsmith:
See commit messages. TL;DR we have a new table which stores which commit we have indexed in `global_refs`. We prevent concurrent and unnecessary indexing with it.

Fixes https://app.asana.com/0/37478073567611/138332225969213 and https://app.asana.com/0/37478073567611/138332225969215


services/backend/internal/localstore/global_refs.go
```diff
@@ -382,6 +398,11 @@ ON COMMIT DROP;`
 	GROUP BY def_key_id, repo, file;`
 
 	return dbutil.Transact(graphDBH(ctx), func(tx gorp.SqlExecutor) error {
+		// Update and lock version row
+		if _, err := tx.Exec(updateVersionSQL, commitID, repo.URI); err != nil {
```
neelance: What are the SQL semantics here? Does the SELECT query below block until the transaction is complete?

neelance: Does it make any difference if we move the initial SELECT into the transaction?

keegancsmith: I may have done this incorrectly, I'm researching this properly now + will do a little local test.

As for moving the initial select in: Between starting the transaction and the initial select we read from the graphstore + read/write to the `def_keys` table. I'd like to keep those outside of the transaction to reduce the amount of time we are in the transaction. As mentioned in the commit message having it outside of the transaction does mean we still have race conditions. What I can do instead is repeat the `oldCommitID` check.

neelance: What about doing both? A check where it currently is and another one at the beginning of the transaction?

neelance: Hmm, but what is bad in expanding the transaction to this earlier SELECT? It should only lock the version row, right? And that one we WANT to have locked if some concurrent import happens.

keegancsmith: yup, you are correct. I'll do that.

keegancsmith:
@neelance PTAL. Will squash fixup commits into the `Prevent duplicate work...` commit.

neelance:
LGTM

beyang:
@keegancsmith Now if someone triggers a rebuild from the UI and the commit already exists in `global_defs`, it won't get reindexed. Should that action also set the `Force = true` parameter?

keegancsmith:
@beyang yeah we should. Filed https://app.asana.com/0/37478073567611/139929701713598

PR #6, opened 2016-06-02
UX: Tools page expanded popup support

KattMingMing:
##### Description

Tools improvements to provide additional context and usage information up front for users before installing a plugin or browser extension. With the number of these growing, the current tools page will become cluttered with very similar looking plugins and has a high probability of causing confusion. The solution is to have generic components on the tools page that discuss the high level goal of what the user is about to commit to using. For example: Browser extension. Seeing a panel that gives high level context and then clicking to view more information results in the ability to show multiple extensions on a page that is devoted to explaining the purpose of the extension, how it works, which extensions we support and additional steps to get started if necessary. It also reduces complexity for managing state of these extensions because it more about tools we support and are supporting in the future and less about a status of your Sourcegraph account. These changes also allow for direct linking to a specific tool while keeping the user on Sourcegraph and providing enough information up front about what the user is seeing. There is also the ability to opt in for email notifications if a user does not see a supported integration such as Firefox.  See screenshots below:

screenshot 2016-06-02 15 11 40
screenshot 2016-06-02 15 11 49
screenshot 2016-06-02 15 14 46

##### Testing Information
Available for testing on staging4

##### People

@uforic - Can you take a look at this since these changes reflect your SFYE updates


chexee:
Are these available logged out?

chexee:
What does "x" do if these are not true modals?

chexee:
Imo, I think it makes more sense to add more context to the tools page itself than it does to expand these to individual pages. It looks like we're preparing for complexity that doesn't seem like it's going to exist any time soon.

KattMingMing:
@chexee  
1) Still talking through that, but it feels like the right decision. What do you think?
2) The X Brings them back to the tools page, these can easily be moved to just modals, but the reason they weren't was because a lot of the implementation details and tech details would live on these.

KattMingMing:
As for complexity, we currently have 2 browser extensions, 2 types of GitHub authing levels and 2 editors. These could very well expand into a larger view on the page. The current tools page lacks a lot of the education around how to use many of these features and also why they are important. What do you suggest @chexee?

chexee:
If we do these, it makes sense to make them available both logged in and logged out. 

Overall this experience seems a bit hacky and I'm not positive what problem we're trying to solve.

chexee:
My recommendation here would be to redesign the tools page to add room for additional context, rather than create three very sparse pages. The grouping you have here makes sense, but the singular pages could actually just live on the tools page as bigger sections.

The new page is a bit disorienting (Am I still on the tools page? How do I get back? What happens if I dismiss this?) – and imo, the real problem is that the current tools page may just be too simplistic.

KattMingMing:
What are your thoughts on something like this? User clicks learn more and instead of taking the user elsewhere this happens.
https://www.dropbox.com/s/ysp27p6uo3mdeso/Screenshot%202016-06-02%2015.05.39.png?dl=0

chexee:
Again, the extra click still seems unnecessary and seems like it obscures the functionality and information that's important. If seeing install/auth status and additional actions are important, we shouldn't obscure them behind a click when there's plenty of room an the page to make these sections the user can scan through.

KattMingMing:
Sounds good. I'll continue to rethink this and if you have any mocks you'd like to share or thoughts ping me and we can continue to discuss.

KattMingMing:
@chexee Thanks for all the design and UX feedback. I'll have this updated this morning with something that is less navigation and incorporates the minor fixes.

app/web_modules/sourcegraph/home/ToolComponent.js
```diff
@@ -0,0 +1,61 @@
...
+					<div styleName="tool-item">
+						<Panel hoverLevel="high">
+							<span style={{display: "block", textAlign: "left"}}>
+							<Button onClick={this._backCTAClicked.bind(this)} color="white" style={{textAlign: "left"}}>
+								<TriangleLeftIcon className={base.pt2} />&nbsp;
+							</Button>
+							</span>
+							<div style={{display: "flex"}}>
+								<span><img styleName="tool-img" className={base.pt5} src={`${this.context.siteConfig.assetsRoot}${this.props.supportedTool.hero.img}`}></img></span>
+								<div style={{display: "block", paddingLeft: "15px"}}>
```
rothfels: Why inlining style?

app/web_modules/sourcegraph/home/ToolComponent.js
```diff
@@ -0,0 +1,61 @@
...
+									<Heading align="left" level="2" className={base.pt5}>{this.props.supportedTool.hero.title}</Heading>
+									<div styleName="tool-item-paragraph" className={base.pt2}>
+										{this.props.supportedTool.hero.subtitle}
+										<br/><br/>
+										{this.props.supportedTool.hero.paragraph}
+									</div>
+								</div>
+							</div>
+							{this.props.supportedTool.gif && <div styleName="tool-gif-container" className={base.pv3}>
```
rothfels: It's weird to me that we have 3 different ways of styling in here: `className`, `styleName`, and `style`. Is there a reason?

app/web_modules/sourcegraph/home/ToolComponent.js
```diff
@@ -0,0 +1,61 @@
...
+
+	render() {
+		return (
+			<div>
+				<div styleName="tool-container">
+					<div styleName="tool-item">
+						<Panel hoverLevel="high">
+							<span style={{display: "block", textAlign: "left"}}>
+							<Button onClick={this._backCTAClicked.bind(this)} color="white" style={{textAlign: "left"}}>
+								<TriangleLeftIcon className={base.pt2} />&nbsp;
```
rothfels: Instead of using `&nbsp;` can you just have the style have right padding?

app/web_modules/sourcegraph/home/ToolsContainer.js
```diff
@@ -1,107 +1,23 @@
...
 import CSSModules from "react-css-modules";
 import styles from "./styles/Tools.css";
-import base from "sourcegraph/components/styles/_base.css";
-import {Heading, Hero, Panel, Button} from "sourcegraph/components";
-import redirectForUnauthedUser from "sourcegraph/user/redirectForUnauthedUser";
+import ToolsHomeComponent from "./ToolsHomeComponent";
+import Component from "sourcegraph/Component";
 
-class ToolsContainer extends React.Component {
+class ToolsContainer extends Component {
```
rothfels: This component looks like it doesn't do anything anymore. Can we remove it?

rothfels: As in, it's literally just a wrapper around a single component now. Can we just eliminate the wrapper?

app/web_modules/sourcegraph/home/ToolsHomeComponent.js
```diff
@@ -0,0 +1,241 @@
...
+					img: "/img/Dashboard/SourcegraphLiveAsset.svg",
+				},
+				primaryButton: this._syfeCTA(),
+				gif: "/img/Dashboard/LiveEditorExampleGif.gif",
+			},
+		};
+	}
+
+	componentDidMount() {
+		this.timeout = setTimeout(() => this.setState({
```
rothfels: Is this timeout necessary? Did you see what would happen without it?

app/web_modules/sourcegraph/home/ToolsHomeComponent.js
```diff
@@ -0,0 +1,241 @@
...
+			showChromeExtensionCTA: !document.getElementById("sourcegraph-app-bootstrap") && !window.localStorage["installed_chrome_extension"],
+		}), 1);
+	}
+
+	componentWillUnmount() {
+		clearTimeout(this.timeout);
```
rothfels: Is this necessary? Usually only relevant for intervals.

app/web_modules/sourcegraph/home/ToolsHomeComponent.js
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	reconcileState(state, props, context) {
+		Object.assign(state, props);
+		state.githubToken = context.githubToken;
+
+		this.renderComponent = this.supportedTools[state.location.query.tool] ? <ToolComponent supportedTool={this.supportedTools[state.location.query.tool]} location={state.location}/> : null;
```
rothfels: `renderComponent` sounds like a function, not an element.

app/web_modules/sourcegraph/home/ToolsHomeComponent.js
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	_toolClicked(toolType) {
+		this.context.eventLogger.logEventForPage("ToolCTAClicked", "DashboardTools", {toolType: toolType});
+		this.context.router.replace({...this.props.location, query: {tool: toolType}});
+	}
+
+
+	/* Chrome CTA and handlers */
```
rothfels: Is this comment and all the whitespace necessary?

app/web_modules/sourcegraph/home/ToolsHomeComponent.js
```diff
@@ -0,0 +1,241 @@
...
+		}
+
+		return (
+			<Button color="purple" onClick={this._browserLearnMoreCTAClicked.bind(this)}>View on the chrome web store</Button>
+		);
+	}
+
+	_browserLearnMoreCTAClicked() {
+		this.context.eventLogger.logEventForPage("ChromeExtensionStoreCTAClicked", "DashboardTools");
+		window.location.assign("https://chrome.google.com/webstore/detail/sourcegraph-for-github/dgjhfomjieaadpoljlnidmbgkdffpack?hl=en");
```
rothfels: `window.location.href = ...`

rothfels:
A few minor comments for code cleanup (mostly around styling). Feel free to merge when that's cleaned up.

app/web_modules/sourcegraph/home/ToolsContainer.js
KattMingMing: Leaving per our discussion offline. Recap: This is useful for routing as it will separate a routed component vs an experimental presentation.

uforic:
Usability stuff:
Thoughts on making click anywhere on the information box open the lightbox? Since the whole box interacts (/ grows and stuff), my expectation would be that clicking anywhere on it will work. Otherwise, seems weird to have it grow like that.

The image is 9.8mb, so takes a while to load, and the div is small then pops out when its done. Is there a way to load a filler object that takes the same amount of space, to make it less jittery?

Moving action buttons up to the top of the windows would be solid. Perhaps listing them vertically, with a header like:

Editors supported:
Emacs
Sublime

Will work on the GitHub copy.

Other than the above, LGTM!

uforic:
GitHub copy:
Sourcegraph for your repositories
"Start searching, browsing, and cross-referencing your code, with IDE-like capabilities in your browser."

Expanded text:
<incoming>
 



2) for the Sourcegraph for your repositories, thoughts on making the "Connect" button "just work" ? Also, thoughts on making install and connect "just work, but the learn more expand out the lightbox? Inconsistent, but more action oriented

Last thing: on the details page, let's make it:
<Title> (same as tools page title, right now it's different)
<Subtext> bolded, same as tools page

PR #8, opened 2016-06-03
WIP: Try grpc-gateway to replace one httpapi endpoint.

shurcooL:
DO NOT MERGE.

This is just a WIP PR to demonstrate the diff to add grpc-gateway support, and replace one manually created httpapi endpoint (repo branches) with a generated one.

It requires the HTTP URL to be specified in sourcegraph.proto file.

### TODOs

- [ ] Confirm support for HTTP caching headers in response from grpc-gateway handlers.


sqs:
Is it easy/possible for us to set HTTP caching headers on the response?

shurcooL:
That's a great question, I'll make sure we have an answer to it before proceeding.

shurcooL:
This WIP PR has served its purpose of communicating and displaying the diff it'll take to add support for grpc-gateway powered endpoints, so I'll close it. This PR can still be viewed for reference after it's closed.

We may still try grpc-gateway in the future, but that will be a separate PR.

PR #9, opened 2016-06-03
Update repository meta on browser visit

Unknwon:
Update repository description and fork meta when user hits `rebuild` button.

##### Test plan

We used to set `fork` field in `repo` table to be `false`, and did not update this field in the future, which will always appear in the global repo search results. After trigger new build, forked repositories should no longer be listed anymore.

cc @beyang


services/backend/internal/localstore/repos_lg_test.go
```diff
@@ -105,7 +105,7 @@ func TestRepos_CreateDeleteWorks_lg(t *testing.T) {
 			New: &sourcegraph.ReposCreateOp_NewRepo{
 				URI:      repo,
 				CloneURL: fsServer.Config.Serve.AppURL + "/myrepo",
-				Mirror:   true,
+				// Mirror:   true,
```
beyang: removing this line is okay

services/backend/repos.go
```diff
@@ -82,6 +82,24 @@ func (s *repos) Get(ctx context.Context, repoSpec *sourcegraph.RepoSpec) (*sourc
...
+	ghrepo, err := github.ReposFromContext(ctx).Get(ctx, repoSpec.URI)
+	if err != nil {
+		return nil, err
+	}
+	if err = store.ReposFromContext(ctx).Update(ctx, store.RepoUpdate{
+		ReposUpdateOp: &sourcegraph.ReposUpdateOp{
+			Repo:        repoSpec.URI,
+			Description: ghrepo.Description,
+		},
+		Fork: &ghrepo.Fork,
```
beyang: Move `Fork` into the `sourcegraph.ReposUpdateOp` struct.

services/backend/repos.go
```diff
@@ -82,6 +82,24 @@ func (s *repos) Get(ctx context.Context, repoSpec *sourcegraph.RepoSpec) (*sourc
 	if repo.Blocked {
 		return nil, grpc.Errorf(codes.FailedPrecondition, "repo %s is blocked", repo)
 	}
+
+	// No need to fetch meta for non-mirror repositories.
+	if !repo.Mirror {
+		return repo, nil
+	}
+	ghrepo, err := github.ReposFromContext(ctx).Get(ctx, repoSpec.URI)
```
beyang: Add a comment here like "Sync metadata from GitHub to DB"

services/backend/repos.go
```diff
@@ -82,6 +82,24 @@ func (s *repos) Get(ctx context.Context, repoSpec *sourcegraph.RepoSpec) (*sourc
...
+	if err != nil {
```
beyang: Add a unit test to check that `Update` is called on every `Get`.

beyang:
3 comments, otherwise LGTM.

services/backend/internal/localstore/repos.go
```diff
@@ -549,6 +549,10 @@ func (s *repos) Update(ctx context.Context, op store.RepoUpdate) error {
 			return err
 		}
 	}
+	_, err := appDBH(ctx).Exec(`UPDATE repo SET "fork"=$1 WHERE uri=$2`, op.Fork, op.Repo)
```
beyang: should you check nil here?

beyang:
1 minor comment, otherwise LGTM :ship: :ship: :ship:

pkg/store/mockstore/repos_mock_helpers.go
```diff
@@ -22,6 +23,32 @@ func (s *Repos) MockGet(t *testing.T, wantRepo string) (called *bool) {
 	return
 }
 
+func (s *Repos) MockGetRepo(t *testing.T, wantRepo *sourcegraph.Repo) (called *bool) {
```
keegancsmith: this already exists, it is the function below the updates called `MockGet_Returns` (which is also the pattern we use for mock helpers that return what we want)

services/backend/repos.go
```diff
@@ -82,6 +82,25 @@ func (s *repos) Get(ctx context.Context, repoSpec *sourcegraph.RepoSpec) (*sourc
 	if repo.Blocked {
 		return nil, grpc.Errorf(codes.FailedPrecondition, "repo %s is blocked", repo)
 	}
+
+	// No need to fetch meta for non-mirror repositories.
+	if !repo.Mirror {
+		return repo, nil
+	}
+	// Sync metadata from GitHub to DB.
+	ghrepo, err := github.ReposFromContext(ctx).Get(ctx, repoSpec.URI)
```
keegancsmith: doesn't this mean for every call we do to `Get` we will hit github? This will cause us to run out of rate limits very quickly and get a P0

keegancsmith:
The improvement to using the ctx instead of a harcoded github object is awesome! I've got one inline comment which makes me worried about how often we hit github. I actually have a change from a little while ago which I never finished. Instead of updating on every get, I hook into `repoSetFromRemote` call we do. That call has logic to use the cache.

services/backend/internal/localstore/repos.go
keegancsmith: You should also follow the same pattern as above where `op.Fork` is a pointer. Otherwise someone will do a `store.RepoUpdate` without specifying `Fork` which leads to `Fork` becoming false.

keegancsmith:
@Unknwon anything I can do to help unblock this. I really want to get us into a state where we don't hit github as aggressively as we currently do soon. I'm keen to even take a look at extending this PR tomorrow to achieve those goals if that is alright with you?

keegancsmith:
@Unknwon yeah this won't actually cause us to hit the API more (unless the repo is private). The caching lives inside of the github client, which I was mistaken about (I think it used to live somewhere else). Sorry about the false alarm :)

services/backend/internal/localstore/repos.go
Unknwon: @keegancsmith do you have any examples to let `protoc` generate `*bool` type from `*.proto` file to `*.go`? Seems could not find examples in existing files.

Unknwon:
I've updated code to fix addressed problems, please help review again, thanks!

services/backend/internal/localstore/repos_db_test.go
```diff
@@ -181,34 +181,38 @@ func TestRepos_List_URIs(t *testing.T) {
 	}
 }
 
-type RepoGetterMockPublicRepo struct{}
-
-func (r *RepoGetterMockPublicRepo) Get(ctx context.Context, uri string) (*sourcegraph.RemoteRepo, error) {
-	return &sourcegraph.RemoteRepo{Private: false}, nil
+type mockGitHubRepoGetter struct {
```
keegancsmith: Three suggestions:
* Make this live inside of the `github` pkg that we use when we call `github.WithRepos`. (Possibly in a subpackage called mocks)
* Add a helper to this struct called `MockGet_Return`
* Update `testContext` to always set `github.WithRepos` to a mock.

Those suggestions should make the github mock usable by other package (ie `backend` would have uses for it) + would reduce boilerplate inside of the test updates you do.

services/backend/repos.go
```diff
@@ -78,6 +78,32 @@ func (s *repos) Get(ctx context.Context, repoSpec *sourcegraph.RepoSpec) (*sourc
 	if repo.Blocked {
 		return nil, grpc.Errorf(codes.FailedPrecondition, "repo %s is blocked", repo)
 	}
+
+	// No need to fetch meta for non-mirror repositories.
```
keegancsmith: nice, I assume this was often a wasted github request?

services/backend/repos.go
```diff
@@ -78,6 +78,32 @@ func (s *repos) Get(ctx context.Context, repoSpec *sourcegraph.RepoSpec) (*sourc
...
+	if !repo.Mirror {
+		return repo, nil
+	}
+	// Sync metadata from GitHub to DB.
+	ghrepo, err := github.ReposFromContext(ctx).Get(ctx, repo.URI)
```
keegancsmith: This is going to cause issues in production. We have lots of repos that are mirrors whose `repo.URI` will not work on github.

```
# select uri, http_clone_url from repo where mirror = true and uri not like '%github%' limit 10;
               uri               |               http_clone_url
---------------------------------+---------------------------------------------
 gopkg.in/rs/xmux.v0             | https://gopkg.in/rs/xmux.v0
 gopkg.in/go-tomb/tomb.v1        | https://gopkg.in/go-tomb/tomb.v1
 azul3d.org/semver.v1            | https://azul3d.org/semver.v1
 gopkg.in/zboralski/tweetlib.v3  | https://gopkg.in/zboralski/tweetlib.v3
 gopkg.in/stripe/stripe-go.v3    | https://gopkg.in/stripe/stripe-go.v3
 gopkg.in/simplexml.v0           | https://gopkg.in/simplexml.v0
 bitbucket.org/dvirsky/boilerdb  | https://bitbucket.org/dvirsky/boilerdb.git
 gopkg.in/datianshi/crypto.v1    | https://gopkg.in/datianshi/crypto.v1
 gopkg.in/bugsnag/bugsnag-go.v1  | https://gopkg.in/bugsnag/bugsnag-go.v1
 bitbucket.org/SamWhited/sysexit | https://bitbucket.org/SamWhited/sysexit.git
(10 rows)

# select count(1) from repo where mirror = true and uri not like '%github%';
 count
-------
  1986
(1 row)
```

You can check if `githubutil.SplitRepoURI` returns an error to know if it should work. We should also consider doing the metadata update asynchronously and just log errors.

services/backend/repos.go
```diff
@@ -78,6 +78,32 @@ func (s *repos) Get(ctx context.Context, repoSpec *sourcegraph.RepoSpec) (*sourc
...
+		forkOp = sourcegraph.ReposUpdateOp_TRUE
+	} else {
+		forkOp = sourcegraph.ReposUpdateOp_FALSE
+	}
+	if err = store.ReposFromContext(ctx).Update(ctx, store.RepoUpdate{
+		ReposUpdateOp: &sourcegraph.ReposUpdateOp{
+			Repo:        repoSpec.ID,
+			Description: ghrepo.Description,
+			Fork:        forkOp,
+		},
```
keegancsmith: This is turning every read we do into a write. We already have the `repo` object, can we not only update if the values change?

Also a request: Can you also update `DefaultBranch`, `Language` and `Private`?

services/backend/repos.go
```diff
@@ -78,6 +78,32 @@ func (s *repos) Get(ctx context.Context, repoSpec *sourcegraph.RepoSpec) (*sourc
 	if repo.Blocked {
 		return nil, grpc.Errorf(codes.FailedPrecondition, "repo %s is blocked", repo)
 	}
+
```
keegancsmith: Also note that just above this we do `setRepoFieldsFromRemote`, which is also doing a `github.ReposFromContext(ctx).Get(ctx, repo.URI)`. FYI I have this WIP commit when I looked into this

```
commit a6300a513b6e9b93e72639a94f703430e3cacaf6
Author: Keegan Carruthers-Smith <keegan.csmith@gmail.com>
Date:   Wed May 25 16:35:27 2016 +0200

    github: Record repo metadata in db

diff --git a/pkg/store/repos.go b/pkg/store/repos.go
index 6ee3263..aaba4d4 100644
--- a/pkg/store/repos.go
+++ b/pkg/store/repos.go
@@ -37,6 +37,8 @@ type Repos interface {
 type RepoUpdate struct {
 	*sourcegraph.ReposUpdateOp
 
+	Fork      bool
+	Private   bool
 	UpdatedAt *time.Time
 	PushedAt  *time.Time
 }
diff --git a/services/backend/repos.go b/services/backend/repos.go
index 3d78a8f..ad3a6ed1 100644
--- a/services/backend/repos.go
+++ b/services/backend/repos.go
@@ -105,6 +105,7 @@ func (s *repos) List(ctx context.Context, opt *sourcegraph.RepoListOptions) (*so
 }
 
 func (s *repos) setRepoFieldsFromRemote(ctx context.Context, repo *sourcegraph.Repo) error {
+	var update *store.RepoUpdate
 	repo.HTMLURL = conf.AppURL(ctx).ResolveReference(app_router.Rel.URLToRepo(repo.URI)).String()
 	// Fetch latest metadata from GitHub (we don't even try to keep
 	// our cache up to date).
@@ -113,20 +114,84 @@ func (s *repos) setRepoFieldsFromRemote(ctx context.Context, repo *sourcegraph.R
 		if err != nil {
 			return err
 		}
-		repoSetFromRemote(repo, ghrepo)
+		update = repoSetFromRemote(repo, ghrepo)
+	}
+
+	if update != nil {
+		log15.Debug("Updating repo metadata from remote", "repo", repo.URI)
+		if err := store.ReposFromContext(ctx).Update(ctx, *update); err != nil {
+			log15.Error("Failed updating repo metadata from remote", "repo", repo.URI, "error", err)
+		}
 	}
 
 	return nil
 }
 
-func repoSetFromRemote(repo *sourcegraph.Repo, ghrepo *sourcegraph.RemoteRepo) {
-	repo.Description = ghrepo.Description
-	repo.Language = ghrepo.Language
-	repo.DefaultBranch = ghrepo.DefaultBranch
-	repo.Fork = ghrepo.Fork
-	repo.Private = ghrepo.Private
-	repo.UpdatedAt = ghrepo.UpdatedAt
-	repo.PushedAt = ghrepo.PushedAt
+// repoSetFromRemote updates repo with fields from ghrepo that are
+// different. If any fields are changed a non-nil store.RepoUpdate is returned
+// representing the update.
+func repoSetFromRemote(repo *sourcegraph.Repo, ghrepo *sourcegraph.RemoteRepo) *store.RepoUpdate {
+	u := store.RepoUpdate{
+		ReposUpdateOp: &sourcegraph.ReposUpdateOp{
+			Repo: repo.RepoSpec(),
+		},
+	}
+	updated := false
+	if repo.Description != ghrepo.Description {
+		repo.Description = ghrepo.Description
+		u.Description = ghrepo.Description
+		updated = true
+	}
+	if repo.Language != ghrepo.Language {
+		repo.Language = ghrepo.Language
+		u.Language = ghrepo.Language
+		updated = true
+	}
+	if repo.DefaultBranch != ghrepo.DefaultBranch {
+		repo.DefaultBranch = ghrepo.DefaultBranch
+		u.DefaultBranch = ghrepo.DefaultBranch
+		updated = true
+	}
+	if repo.Fork != ghrepo.Fork {
+		repo.Fork = ghrepo.Fork
+		u.Fork = ghrepo.Fork
+		updated = true
+	}
+	if repo.Private != ghrepo.Private {
+		repo.Private = ghrepo.Private
+		u.Private = ghrepo.Private
+		updated = true
+	}
+	if !timestampEqual(repo.UpdatedAt, ghrepo.UpdatedAt) {
+		repo.UpdatedAt = ghrepo.UpdatedAt
+		if ghrepo.UpdatedAt != nil {
+			t := ghrepo.UpdatedAt.Time()
+			u.UpdatedAt = &t
+		}
+		updated = true
+	}
+	if !timestampEqual(repo.PushedAt, ghrepo.PushedAt) {
+		repo.PushedAt = ghrepo.PushedAt
+		if ghrepo.PushedAt != nil {
+			t := ghrepo.PushedAt.Time()
+			u.PushedAt = &t
+		}
+		updated = true
+	}
+	if !updated {
+		return nil
+	}
+	return &u
+}
+
+func timestampEqual(a, b *pbtypes.Timestamp) bool {
+	if a == b {
+		return true
+	}
+	if a == nil || b == nil {
+		return false
+	}
+	return a.Seconds == b.Seconds && a.Nanos == b.Nanos
 }
 
 func (s *repos) Create(ctx context.Context, op *sourcegraph.ReposCreateOp) (repo *sourcegraph.Repo, err error) {
```

keegancsmith:
@Unknwon quite a few inline comments, sorry I didn't review properly before :) Looking forward to this change shipping! Will make several parts of our codebase much better / help use move towards not hitting github as much.

services/backend/repos.go
Unknwon: Hmm... what am I supposed to do with this patch?

services/backend/repos.go
Unknwon: > This is turning every read we do into a write. We already have the repo object, can we not only update if the values change?

I think this has the same problem as we may send a lot GitHub API request which just few lines above, but in production, we have cache so this write won't be much as well?

services/backend/internal/localstore/repos_db_test.go
Unknwon: >Update testContext to always set github.WithRepos to a mock.

This can't be done because you don't know what return results you expect when setting up test context.

Unknwon: >Add a helper to this struct called MockGet_Return

Not sure what's the primary purpose of adding this, for the current (newly added) code, in terms of lines of code, there is no reduction.

services/backend/repos.go
Unknwon: https://github.com/sourcegraph/sourcegraph/pull/9/commits/ae6961f9c1050a84aaae0314fde91986406acfe7

services/backend/internal/localstore/repos_db_test.go
keegancsmith: `testContext` returns a mock struct. See the implementation, we set mocks for servers and stores. We can also set a mock for github there. With a helper it will change lines like

```
ctx, _, done := textContext()
defer done()
ctx = github.WithRepos(ctx, mockGitHubRepoGetter{
        Get_: func(context.Context, string) (*sourcegraph.RemoteRepo, error) {
                return &sourcegraph.RemoteRepo{Private: false}, nil
        },
})
```
to
```
ctx, mock, done := textContext()
defer done()
mock.githubRepo.MockGet_Return(ctx, &sourcegraph.RemoteRepo{Private: false})
```

note this is not blocking this landing. Just a suggestion to make it consistent with our other mocks + cleaner.

services/backend/repos.go
keegancsmith: it is some base code so you don't have to do a DB write if nothing has actually changed between what is in the DB and what github returns.

services/backend/repos.go
keegancsmith: there is a cache, so we don't have to worry about doing the same GH API request twice in a row, etc. By doing a write I mean a write to our pgsql database, which this will do for every Get since you call Update.

services/backend/repos.go
beyang: Joe and I chatted about this and it seems that checking the last modified date would still incur a DB read (to read the previous modified date). Given that, we don't think it's worth doing the check. Is there something we're missing?

Unknwon:
Changes for using mocks in `*_test.go` has been made.

services/backend/repos.go
```diff
@@ -93,22 +93,27 @@ func (s *repos) Get(ctx context.Context, repoSpec *sourcegraph.RepoSpec) (*sourc
...
-	}
-	if err = store.ReposFromContext(ctx).Update(ctx, store.RepoUpdate{
-		ReposUpdateOp: &sourcegraph.ReposUpdateOp{
-			Repo:        repoSpec.ID,
-			Description: ghrepo.Description,
-			Fork:        forkOp,
-		},
-	}); err != nil {
-		return nil, err
+	repoSetFromRemote(repo, ghrepo)
```
shurcooL: `repoSetFromRemote` does, among other things:

```
repo.Description = ghrepo.Description
repo.Fork = ghrepo.Fork
```

Your next `if` statement comparison will never be true, since it happens right after:

```Go
if repo.Description != ghrepo.Description || repo.Fork != ghrepo.Fork {
```

Or am I missing something?

beyang: good catch

services/backend/repos.go
keegancsmith: We don't need to read the DB again, we have already read it (the repo struct is from the db). You just need to compare what is in that struct with what github thinks. If there is a difference only do a write then. Otherwise we will be doing a shittone more writes to our DB. I'm not sure if postgres will be performant at doing writes which are no-ops, but at the end of the day it will likely still need to grab a write lock.

keegancsmith:
LGTM

I'd love to sync all the fields we have in the DB (especially DefaultBranch and Private). We can follow up if necessary.

services/backend/repos_test.go
```diff
@@ -33,8 +33,9 @@ func TestReposService_Get(t *testing.T) {
 	if !*calledGet {
 		t.Error("!calledGet")
 	}
-	if !*calledUpdate {
-		t.Error("!calledUpdate")
+	// Should not be called because mock GitHub has same data as mock DB.
+	if *calledUpdate {
+		t.Error("calledUpdate")
```
keegancsmith: ooh, maybe this is a real bug?

Do we have a test to ensure we call update?

Unknwon: Indeed, it turns out to be a real problem.

Before we check if meta needs to be updated, we already fetch meta from GitHub and set latest values to `repo` varaible by `s.setRepoFieldsFromRemote`: https://sourcegraph.com/sourcegraph/sourcegraph/-/blob/services/backend/repos.go#L68-73

Therefore, the `repo` and `ghrepo` are always having the same value, and we're calling GitHub API twice every time for the same reason.

Any thoughts?

Unknwon: The ideal solution I can think of is putting update DB code (`store.ReposFromContext(ctx).Update...`) inside `s.setRepoFieldsFromRemote` method, and add an argument to indicate whether it should do the update or not, because not only `Get` method is calling it, but also `List` and others. This way, we can remain calling GitHub API only once and compare the values first before assign/update them to the `repo` variable.

keegancsmith: Updating in `setRepoFieldsFromRemote` was one of my original suggestions (it is what that patch I pasted in was starting to do). I think we can avoid the boolean argument and just always update the repo if the gh metadata data has changed (even if we are just doing a list).

Luckily calling the ghrepo repeatedly is not that bad, since we have a cache with a 10min TTL. The only case we don't use the cache is in the case of private repos, but in that case I think we may hit an HTTP cache (namespaced per user) + that request doesn't affect our rate limit since the request will be authed with the user's credentials. However, multiple calls are clearly wasteful. So I agree, just doing it inside the `setRepoFieldsFromRemote` is the best place and we can remove most specific calls to ghrepo from everywhere but in that function.

Unknwon:
@keegancsmith I have made corresponding changes based on your patch, and added a test to make sure when GitHub meta changes, DB update operation happens.

PR #12, opened 2016-06-03
srclib-yaml coverage

attfarhan:
Changes to add srclib-YAML toolchain coverage metrics.

##### Reviewer tasks

### DIFF
--- a/pkg/coverageutil/yaml.go
+++ b/pkg/coverageutil/yaml.go
@@ -1,41 +1,92 @@
 package coverageutil
 
 import (
-	"bytes"
-	"text/scanner"
+	"fmt"
+	"github.com/attfarhan/yaml"
 )
 
 type yamlTokenizer struct {
-	scanner *scanner.Scanner
+	values     []string
+	startBytes []int
+	pointer    int
 }
 
+type T struct {
+	value  []string
+	line   []int
+	column []int
+}
+
+// Initializes text scanner that extracts only idents
 func (s *yamlTokenizer) Init(src []byte) {
-	s.scanner = &scanner.Scanner{}
-	s.scanner.Error = func(s *scanner.Scanner, msg string) {}
-	s.scanner.Init(bytes.NewReader(src))
-}
-
-func (s *yamlTokenizer) Done() {
-}
-
-func (s *yamlTokenizer) Next() *Token {
-	for {
-		r := s.scanner.Scan()
-		if r == scanner.EOF {
-			return nil
-		}
-		if r != scanner.Ident {
-			continue
-		}
-		text := s.scanner.TokenText()
-		p := s.scanner.Pos()
-		return &Token{uint32(p.Offset - len([]byte(text))), text}
+	p := yaml.NewParser(src)
+	node := p.Parse()
+	var nodeList []*yaml.Node
+	tokenList := yaml.Explore(node, nodeList)
+	out := &T{}
+	getLineAndColumn(tokenList, string(src), out)
+	for i, _ := range out.value {
+		start, _, value := findOffsets(string(src), out.line[i], out.column[i], out.value[i])
+		s.values = append(s.values, value)
+		s.startBytes = append(s.startBytes, start)
 	}
+
+	s.pointer = 0
 }
+
+// Next returns idents that are not Java keywords
+func (s *yamlTokenizer) Next() *Token {
+	if s.pointer >= len(s.values) {
+		return nil
+	}
+	out := &Token{}
+	out.Offset = uint32(s.startBytes[s.pointer])
+	out.Text = s.values[s.pointer]
+	s.pointer++
+	fmt.Println(out)
+
+	return out
+}
+
+func (s *yamlTokenizer) Done() {}
 
 func init() {
 	var factory = func() Tokenizer {
-		return &javaTokenizer{}
+		return &yamlTokenizer{}
 	}
 	newExtensionBasedLookup("YAML", []string{".yml"}, factory)
+}
+
+func getLineAndColumn(tokenList []*yaml.Node, fileString string, out *T) {
+	for _, token := range tokenList {
+		out.value = append(out.value, token.Value)
+		out.line = append(out.line, token.Line)
+		out.column = append(out.column, token.Column)
+		// a, b := findOffsets(data, token.Line, token.Column, token.Value)
+		// fmt.Println("start: ", a, "End: ", b)
+		getLineAndColumn(token.Children, fileString, out)
+	}
+}
+
+func findOffsets(fileText string, line, column int, token string) (start, end int, value string) {
+
+	// we count our current line and column position.
+	currentCol := 0
+	currentLine := 0
+	for offset, ch := range fileText {
+		// see if we found where we wanted to go to.
+		if currentLine == line && currentCol == column {
+			end = offset + len([]byte(token))
+			return offset, end, token
+		}
+
+		// line break - increment the line counter and reset the column.
+		if ch == '\n' {
+			currentLine++
+			currentCol = 0
+		} else {
+			currentCol++
+		}
+	}
+	return -1, -1, token // not found.
 }
--- a/pkg/coverageutil/yaml.go
+++ b/pkg/coverageutil/yaml.go
@@ -2,54 +2,54 @@ package coverageutil
 
 import (
 	"github.com/attfarhan/yaml"
-	// "log"
 )
 
 type yamlTokenizer struct {
-	// String values of tokens in the tokenizer
+	// values are the tokens returned by the tokenizer
 	values     []string
 	startBytes []int
-	line       []int
-	// Pointer represents the index in each
-	// field as we traverse each field.
+	lines      []int
+	// pointer is the index we're currently viewing in each of the arrays
 	pointer int
 }
 
-// Initializes text scanner that extracts only idents
 func (s *yamlTokenizer) Init(src []byte) {
-	// construct a new parser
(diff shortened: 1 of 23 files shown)


pkg/coverageutil/yaml.go
```diff
@@ -0,0 +1,62 @@
+package coverageutil
+
+import (
+	"github.com/attfarhan/yaml"
+	// "log"
```
rothfels: Remove.

pkg/coverageutil/yaml.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+type yamlTokenizer struct {
+	// String values of tokens in the tokenizer
```
rothfels: use format `values are the tokens returned by the tokenizer` with matching case

pkg/coverageutil/yaml.go
```diff
@@ -0,0 +1,62 @@
...
+	values     []string
+	startBytes []int
+	line       []int
+	// Pointer represents the index in each
```
rothfels: `pointer represents...`, etc

pkg/coverageutil/yaml.go
```diff
@@ -0,0 +1,62 @@
...
+	// field as we traverse each field.
```
rothfels: Comment doesn't make sense. What's a field? What is the "index in each field while we traverse each field" mean?

pkg/coverageutil/yaml.go
```diff
@@ -0,0 +1,62 @@
...
+	pointer int
+}
+
+// Initializes text scanner that extracts only idents
```
rothfels: `Init initializes...`

pkg/coverageutil/yaml.go
```diff
@@ -0,0 +1,62 @@
...
+func (s *yamlTokenizer) Init(src []byte) {
+	// construct a new parser
```
rothfels: This comment is unnecessary, the code on the next line is pretty easy to understand.

pkg/coverageutil/yaml.go
```diff
@@ -0,0 +1,62 @@
...
+	p := yaml.NewParser(src)
+	// node is the tree of tokens returned by the parser.
+	node := p.Parse()
+
+	fileString := string(src)
+
+	// List of nodes representing tokens.  Remove the first because YAML
+	// always starts with an empty token, and begins any sequence with
```
rothfels: The second clause (after the comma) seems to just be repeating the first clause.

pkg/coverageutil/yaml.go
```diff
@@ -0,0 +1,62 @@
...
+
+	for _, token := range tokenList {
+		s.line = append(s.line, token.Line)
+		s.values = append(s.values, token.Value)
+		s.startBytes = append(s.startBytes, token.StartByte)
+	}
+	s.pointer = 0
+}
+
+// Next returns YAML idents
```
rothfels: I don't think you need to comment these interface implementations; the Tokenizer interface has already commented what these functions do. Remove.

pkg/coverageutil/yaml.go
```diff
@@ -0,0 +1,62 @@
...
+	// node is the tree of tokens returned by the parser.
+	node := p.Parse()
+
+	fileString := string(src)
+
+	// List of nodes representing tokens.  Remove the first because YAML
+	// always starts with an empty token, and begins any sequence with
+	// an empty token (both considered starting at byte 0). If we don't remove
+	// it, we will get a duplicate ref key make failure for every file.
+	tokenList := yaml.Explore(fileString, node)[1:]
```
rothfels: Can `Explore` return a nil array (length 0)? If so this will error.

pkg/coverageutil/yaml.go
```diff
@@ -0,0 +1,62 @@
+package coverageutil
+
+import (
+	"github.com/attfarhan/yaml"
+	// "log"
+)
+
+type yamlTokenizer struct {
+	// String values of tokens in the tokenizer
+	values     []string
```
rothfels: Why not just call them tokens? Feels like we have 3-4 words here all meaning the same thing.

pkg/coverageutil/yaml.go
```diff
@@ -0,0 +1,62 @@
...
+	startBytes []int
+	line       []int
```
rothfels: Should be plural `lines`

pkg/coverageutil/yaml.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	s.pointer = 0
+}
+
+// Next returns YAML idents
+func (s *yamlTokenizer) Next() *Token {
+	if s.pointer >= len(s.values) {
+		return nil
+	}
+	out := &Token{}
```
rothfels: You can just inline the values, like this:

```
return &Token{
  Offset: ...,
}
```

pkg/coverageutil/yaml.go
```diff
@@ -0,0 +1,62 @@
...
+	out.Offset = uint32(s.startBytes[s.pointer])
+	out.Text = s.values[s.pointer]
+	out.Line = s.line[s.pointer]
+	s.pointer++
```
rothfels: You can `defer s.pointer++` at the top of this method and it will always remember to execute that expression when the function returns.

pkg/coverageutil/yaml_test.go
```diff
@@ -0,0 +1,42 @@
...
+		actual := make([]*Token, 0)
+		for {
+			tok := tokenizer.Next()
+			if tok == nil {
+				break
+			}
+			actual = append(actual, tok)
+		}
+		if len(actual) != len(t.expected) {
+			testing.Fatalf("%s: Expected %d tokens, got %d instead %v", t.name, len(t.expected), len(actual))
```
rothfels: Can you rebase and refactor this to use the new coverage test helper that Alex wrote?

services/worker/plan/srclib_images.go
```diff
@@ -16,6 +16,7 @@ var (
 	droneSrclibCSharpImage     = "sourcegraph/srclib-csharp@sha256:e5c112fc5ccb0551a09289cd732b00d038c4bc366f853e552826e36bcd903507"
 	droneSrclibCSSImage        = "sourcegraph/srclib-css@sha256:5dea2ffe7183d2fb3f3f3d6a43790189f1c9a82ef46d8d605d87a99830ca9fbd"
 	droneSrclibPythonImage     = "sourcegraph/srclib-python@sha256:aeb6ecf215d75936a539ca72770ef0dd996cd2ebd0e6130a4deb32a4edf6a48a"
+	droneSrclibYAMLImage       = "sourcegraph/srclib-yaml"
```
rothfels: This needs to have an image SHA before it is merged.

pkg/coverageutil/yaml.go
```diff
@@ -0,0 +1,66 @@
+package coverageutil
+
+import (
+	"github.com/attfarhan/yaml"
+)
+
+type yamlTokenizer struct {
+	// values are the tokens returned by the tokenizer
+	values     []string
```
rothfels: Instead of calling this values, why not just call it tokens?

pkg/coverageutil/yaml.go
```diff
@@ -0,0 +1,66 @@
...
+	startBytes []int
+	lines      []int
```
rothfels: Rather than keeping 3 separate, parallel slices, how about a secondary structure for the tokens? Then `yamlTokenizer` only needs to keep one slice. Could you use `[]*coverageUtil.Token`? That's what you want to return anyway, and it has the same 3 fields that are encoded here.

pkg/coverageutil/yaml.go
```diff
@@ -0,0 +1,66 @@
...
+	} else {
+		tokenList = yaml.Explore(fileString, node)[1:]
+	}
+
+	for _, token := range tokenList {
+		s.lines = append(s.lines, token.Line)
+		s.values = append(s.values, token.Value)
+		s.startBytes = append(s.startBytes, token.StartByte)
+	}
+	s.pointer = 0
```
rothfels: Unnecessary; the starting (empty) value for an int (in a struct and otherwise) is 0.,

services/worker/plan/srclib_images.go
```diff
@@ -16,6 +16,7 @@ var (
 	droneSrclibCSharpImage     = "sourcegraph/srclib-csharp@sha256:e5c112fc5ccb0551a09289cd732b00d038c4bc366f853e552826e36bcd903507"
 	droneSrclibCSSImage        = "sourcegraph/srclib-css@sha256:5dea2ffe7183d2fb3f3f3d6a43790189f1c9a82ef46d8d605d87a99830ca9fbd"
 	droneSrclibPythonImage     = "sourcegraph/srclib-python@sha256:c15cb9f11f63e1cfd3b3ca7b9e188bb74fae2faf5a90c38700490187068d88d1"
+	droneSrclibYAMLImage       = "sourcegraph/srclib-yaml"
```
rothfels: Another friendly reminder, the toolchain needs to be deployed and this needs to have a proper `@sha265` before it gets merged.

rothfels:
A few last comments for code cleanup/quality. Otherwise can you get the new srclib-yaml image deployed to dockerhub, get this branch on a staging tier, and queue a build on a repo then send me a link to an annotated YAML file?

rothfels:
@attfarhan closing this, we can pick it up later if it becomes a priority.

PR #13, opened 2016-06-03
app/web_modules: Don't navigate to selected search result on Right Arrow key.

shurcooL:
Including `this._onSelection();` for ArrowRight case was likely unintended; it was meant only for the Enter key case immediately below.

Fixes https://app.asana.com/0/87040567695724/139365933257941/f.

##### Reviewer tasks

- [x] OPTIONAL LOCAL TESTING: reviewer can confirm this fixes the issue

##### Test plan

Should probably be tested. I have not tested it at all yet. But it's trivial enough to have pretty high confidence.


KattMingMing:
lgtm ++

PR #14, opened 2016-06-03
[DO NOT MERGE] Automatically refresh the "Cloning this repository" screen once finished.

slimsag:
Before, when viewing a repository that wasn't cloned you would be asked to refresh the page: 
(and if you refreshed too early, code intelligence would never load until you refresh yet again).

After, the page automatically refreshes and always shows code intelligence:

##### Reviewer tasks

In addition to the below, please confirm my JS code is on-point; nit picky review is very much appreciated!

### DIFF
--- a/app/web_modules/sourcegraph/components/Header.js
+++ b/app/web_modules/sourcegraph/components/Header.js
@@ -3,18 +3,20 @@
 import React from "react";
 import CSSModules from "react-css-modules";
 import styles from "./styles/header.css";
+import Loader from "./Loader";
 
 class Header extends React.Component {
 	static propTypes = {
 		title: React.PropTypes.string.isRequired,
 		subtitle: React.PropTypes.string,
+		loading: React.PropTypes.bool,
 	};
 
 	render() {
 		return (
 			<div styleName="container">
 				<div styleName="cloning-title">{this.props.title}</div>
-				<div styleName="cloning-subtext">{this.props.subtitle}</div>
+				<div styleName="cloning-subtext">{this.props.loading ? <Loader /> : this.props.subtitle}</div>
 			</div>
 		);
 	}
--- a/app/web_modules/sourcegraph/repo/withResolvedRepoRev.js
+++ b/app/web_modules/sourcegraph/repo/withResolvedRepoRev.js
@@ -31,6 +31,11 @@ export default function withResolvedRepoRev(Component: ReactClass, isMainCompone
 			return [RepoStore];
 		}
 
+		componentWillUnmount() {
+			if (super.componentWillUnmount) super.componentWillUnmount();
+			if (this._cloningInterval) clearInterval(this._cloningInterval);
+		}
+
 		reconcileState(state, props) {
 			Object.assign(state, props);
 
@@ -97,6 +102,33 @@ export default function withResolvedRepoRev(Component: ReactClass, isMainCompone
 				if (!nextState.inventory && nextState.commitID && nextState.repoObj && !nextState.repoObj.Error && !nextState.isCloning) {
 					Dispatcher.Backends.dispatch(new RepoActions.WantInventory(nextState.repo, nextState.commitID));
 				}
+			}
+
+			// If the repository is cloning, poll against the server for an
+			// update periodically.
+			if (isMainComponent && nextState.isCloning && !this._cloningInterval && !this._cloningTimeout) {
+				// If the cloning would be quick, we don't want to flicker the
+				// loading screen, so display the screen for at least 1s.
+				const displayForAtLeast = 500;
+				const pollInterval = 500;
+				const maxAttempts = 10000 / pollInterval; // 10s / 20 times
+
+				this._cloningTimeout = true;
+				let attempt = 0;
+				setTimeout(() => {
+					attempt++;
+					if (attempt > maxAttempts) {
+						clearInterval(this._cloningInterval);
+						this._cloningInterval = null;
+					}
+					this._cloningTimeout = false;
+					this._cloningInterval = setInterval(() => {
+						Dispatcher.Backends.dispatch(new RepoActions.WantResolveRev(nextState.repo, nextState.rev, true));
+					}, pollInterval);
+				}, displayForAtLeast);
+			} else if (!nextState.isCloning && this._cloningInterval) {
+				clearInterval(this._cloningInterval);
+				this._cloningInterval = null;
 			}
 		}
 
--- a/app/web_modules/sourcegraph/tree/TreeSearch.js
+++ b/app/web_modules/sourcegraph/tree/TreeSearch.js
@@ -181,17 +181,16 @@ class TreeSearch extends Container {
 			}
 		}
 
-		// If there is no srclib version, i.e. the repository has not been
-		// built at least once, then poll against the server for an update
-		// periodically.
-		const noSrclibVersion = !nextState.srclibDataVersion || Object.keys(nextState.srclibDataVersion).length === 0;
-		if (noSrclibVersion && nextState.commitID && !this._srclibBuildingInterval) {
-			const pollRate = 500;
+		// If there was previously a response from the server but no srclib
+		// data version, i.e., if the repository has not been built recently
+		// then poll against the server for an update periodically.
+		const pollSrclibVersion = nextState.srclibDataVersion && !nextState.srclibDataVersion.CommitID;
+		if (pollSrclibVersion && nextState.commitID && !this._srclibBuildingInterval) {
+			const pollInterval = 500;
 			this._srclibBuildingInterval = setInterval(() => {
 				Dispatcher.Backends.dispatch(new TreeActions.WantSrclibDataVersion(nextState.repo, nextState.commitID, null, true));
-				Dispatcher.Backends.dispatch(new TreeActions.WantFileList(nextState.repo, nextState.commitID));
-			}, pollRate);
-		} else if (!noSrclibVersion && this._srclibBuildingInterval) {
+			}, pollInterval);
+		} else if (!pollSrclibVersion && this._srclibBuildingInterval) {
 			clearInterval(this._srclibBuildingInterval);
 			this._srclibBuildingInterval = null;
 		}
--- a/app/web_modules/sourcegraph/tree/TreeSearch.js
+++ b/app/web_modules/sourcegraph/tree/TreeSearch.js
@@ -141,6 +141,7 @@ class TreeSearch extends Container {
 		if (global.document) {
 			document.removeEventListener("keydown", this._handleKeyDown);
 		}
+		if (this._srclibBuildingInterval) clearInterval(this._srclibBuildingInterval);
 	}
 
 	stores(): Array<Object> { return [TreeStore, DefStore, SearchStore]; }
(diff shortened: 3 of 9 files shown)


app/web_modules/sourcegraph/components/Header.js
```diff
@@ -3,18 +3,21 @@
...
 
 class Header extends React.Component {
 	static propTypes = {
 		title: React.PropTypes.string.isRequired,
 		subtitle: React.PropTypes.string,
+		loading: React.PropTypes.bool,
 	};
 
 	render() {
+		let loading = this.props.loading;
```
sqs: unnecessary var, just refer to `this.props.loading` below.

app/web_modules/sourcegraph/repo/withResolvedRepoRev.js
```diff
@@ -31,6 +31,13 @@ export default function withResolvedRepoRev(Component: ReactClass, isMainCompone
 			return [RepoStore];
 		}
 
+		componentWillUnmount() {
+			if (super.componentWillUnmount) super.componentWillUnmount();
+
+			if (!isMainComponent) return;
```
sqs: If someone adds something to the end of this componentWillUnmount method, they may not realize that it won't be executed if it's the non-main component.

I think it would be clearer to have:

```
if (isMainComponent) {
  if (this._cloningInterval) clearInterval(this._cloningInterval);
}
```

(Also note the omission of curly braces for one-line if-statements.)

Actually, I think it's unnecessary to check `if (isMainComponent)`. If `this._cloningInterval` is truthy and you DON'T clear the interval before unmounting, then you have a leak. So you always want to clear it if it's set. So just this is good:

```
if (this._cloningInterval) clearTimeout(this._cloningInterval);
```

app/web_modules/sourcegraph/repo/withResolvedRepoRev.js
```diff
@@ -98,6 +105,27 @@ export default function withResolvedRepoRev(Component: ReactClass, isMainCompone
 					Dispatcher.Backends.dispatch(new RepoActions.WantInventory(nextState.repo, nextState.commitID));
 				}
 			}
+
+			// If the repository is cloning, poll against the server for an
+			// update periodically.
+			if (nextState.isCloning && !this._cloningInterval && !this._cloningTimeout) {
```
sqs: Should this only occur for the main component as well?

app/web_modules/sourcegraph/repo/withResolvedRepoRev.js
```diff
@@ -98,6 +105,27 @@ export default function withResolvedRepoRev(Component: ReactClass, isMainCompone
...
+				// If the cloning would be quick, we don't want to flicker the
+				// loading screen, so display the screen for at least 1s.
+				const displayForAtLeast = 500;
+				const pollRate = 500;
+
+				this._cloningTimeout = true;
+				let self = this;
```
sqs: Using `self` is unnecessary. ECMAScript 6's new function literal syntax preserves `this` from the outer scope.

app/web_modules/sourcegraph/components/Header.js
slimsag: Thanks for the mention, it was here from when this code was more complex but I found this simpler solution so it's certainly not needed now.

app/web_modules/sourcegraph/repo/withResolvedRepoRev.js
slimsag: Good point; fixed in 99cd23a06be298699e1dcde34fe4766be01e03a8

app/web_modules/sourcegraph/repo/withResolvedRepoRev.js
```diff
@@ -98,6 +105,27 @@ export default function withResolvedRepoRev(Component: ReactClass, isMainCompone
...
 				}
 			}
+
+			// If the repository is cloning, poll against the server for an
+			// update periodically.
+			if (nextState.isCloning && !this._cloningInterval && !this._cloningTimeout) {
+				// If the cloning would be quick, we don't want to flicker the
+				// loading screen, so display the screen for at least 1s.
+				const displayForAtLeast = 500;
+				const pollRate = 500;
```
sqs: For safety and to ensure this won't trip any future rate limiting we add, can you add a max number of attempts or time limit (like 5 min)? And then stop polling after that.

app/web_modules/sourcegraph/repo/withResolvedRepoRev.js
slimsag: Yeah. Fixed by 3db3619370ea4aad76a29460ed0a7ce39e8c97b9

app/web_modules/sourcegraph/repo/withResolvedRepoRev.js
```diff
@@ -31,6 +31,11 @@ export default function withResolvedRepoRev(Component: ReactClass, isMainCompone
 			return [RepoStore];
 		}
 
+		componentWillUnmount() {
+			if (super.componentWillUnmount) super.componentWillUnmount();
+			if (this._cloningInterval) { clearInterval(this._cloningInterval); }
```
sqs: remove curly braces for consistency etc.

app/web_modules/sourcegraph/repo/withResolvedRepoRev.js
slimsag: :happydance: that's perfect. Done in cef8f1a75e1d3c733d6a6bed189e0c233b6469c5

app/web_modules/sourcegraph/tree/TreeSearch.js
```diff
@@ -141,6 +141,7 @@ class TreeSearch extends Container {
 		if (global.document) {
 			document.removeEventListener("keydown", this._handleKeyDown);
 		}
+		if (this._srclibBuildingInterval) { clearInterval(this._srclibBuildingInterval); }
```
sqs: remove curly braces for consistency etc.

app/web_modules/sourcegraph/repo/withResolvedRepoRev.js
slimsag: done

app/web_modules/sourcegraph/tree/TreeSearch.js
slimsag: done

app/web_modules/sourcegraph/tree/TreeSearch.js
```diff
@@ -180,6 +181,21 @@ class TreeSearch extends Container {
 			}
 		}
 
+		// If there is no srclib version, i.e. the repository has not been
+		// built at least once, then poll against the server for an update
```
sqs: Not totally accurate. There could be a build, but it might be older than the commit lookback threshold (250 commits currently; set on the server). `s/at least once/recently/` suffices.

app/web_modules/sourcegraph/tree/TreeSearch.js
```diff
@@ -180,6 +181,21 @@ class TreeSearch extends Container {
...
+		// periodically.
+		const noSrclibVersion = !nextState.srclibDataVersion || Object.keys(nextState.srclibDataVersion).length === 0;
```
sqs: Why the `Object.keys` check? Why not check for `!nextState.srclibDataVersion.CommitID`?

 Also, it's unclear if it an HTTP error in the WantSrclibDataVersion fetch is considered "no srclib version." If there's an HTTP error, then `nextState.srclibDataVersion.Error` will be truthy. I don't see you handling that case.

app/web_modules/sourcegraph/repo/withResolvedRepoRev.js
```diff
@@ -98,6 +103,27 @@ export default function withResolvedRepoRev(Component: ReactClass, isMainCompone
...
 				}
 			}
+
+			// If the repository is cloning, poll against the server for an
+			// update periodically.
+			if (isMainComponent && nextState.isCloning && !this._cloningInterval && !this._cloningTimeout) {
+				// If the cloning would be quick, we don't want to flicker the
+				// loading screen, so display the screen for at least 1s.
+				const displayForAtLeast = 500;
+				const pollRate = 500;
```
sqs: The word "rate" to me makes me think "this will poll 500 times per second" or some other time interval. Obviously I would figure it out quickly, but I think `pollInterval` is a better name.

app/web_modules/sourcegraph/tree/TreeSearch.js
```diff
@@ -180,6 +181,21 @@ class TreeSearch extends Container {
...
+		if (noSrclibVersion && nextState.commitID && !this._srclibBuildingInterval) {
+			const pollRate = 500;
```
sqs: For the same reason described for the other pollRate, I think this should be pollinterval.

app/web_modules/sourcegraph/tree/TreeSearch.js
sqs: While the page is still loading, before the srclib data version HTTP response has been received, this `noSrclibVersion` will be true. So, this will always trigger an interval and then quickly cancel it, even if there is a srclib version that is received quickly. That would cause unnecessary HTTP requests and could slightly slow down the client.

A better way IMO would be to only start the interval if `nextState.srclibDataVersion && !nextState.srclibDataVersion.CommitID`. That means that the server response WAS received and there is no srclib data version.

app/web_modules/sourcegraph/tree/TreeSearch.js
```diff
@@ -180,6 +181,21 @@ class TreeSearch extends Container {
...
+			this._srclibBuildingInterval = setInterval(() => {
+				Dispatcher.Backends.dispatch(new TreeActions.WantSrclibDataVersion(nextState.repo, nextState.commitID, null, true));
+				Dispatcher.Backends.dispatch(new TreeActions.WantFileList(nextState.repo, nextState.commitID));
```
sqs: I don't think you need/want to dispatch WantFileList here. You can just dispatch it when other parts of the state change.

sqs:
Overall looks great, can't wait to have this feature! Go ahead and merge when you've addressed the comments.

app/web_modules/sourcegraph/tree/TreeSearch.js
slimsag: @sqs do you think I need a limit here on this interval / dispatch too? Builds can run for a while.. hmm

app/web_modules/sourcegraph/repo/withResolvedRepoRev.js
slimsag: fixed by 73ec5a2981b42a575d6b630c50be168e18c6a90f with a limit of 20 attempts / 10s in total

app/web_modules/sourcegraph/tree/TreeSearch.js
slimsag: Good point; fixed in da1fd04d2eb4726b0dade9d7453edc3d59cceab0

app/web_modules/sourcegraph/repo/withResolvedRepoRev.js
slimsag: done

app/web_modules/sourcegraph/tree/TreeSearch.js
sqs: Yep.

You can back off in the poll interval.

app/web_modules/sourcegraph/tree/TreeSearch.js
slimsag: you're right, I meant to remove. In fact the first request for this is always valid because the repo is cloned / it doesn't depend on srclibDataVersion. Fixed in eb495d0bc91dece00b54723b0ee8f7f339815cc1

app/web_modules/sourcegraph/tree/TreeSearch.js
slimsag: Very good points, I must have missed `.Error` and `.CommitID`. Done exactly as you said in 3e819c21cb70cf8304bb7b12563ec3821b4ee20e

slimsag:
Squashed all follow-up commits; I will merge myself tomorrow morning once CI passes (and I'd like to demo the comparison :) )

KattMingMing:
lgtm. Awesome addition, thanks for getting this out so quickly

PR #15, opened 2016-06-03
simplify repo-related methods: cut down on RepoSpec usage

sqs:
Many places in the API (protobuf and Go) that were using RepoSpec did
not actually need it; a `repo string` param would suffice. 

Initially we created RepoSpec because we thought we'd want API methods
to accept EITHER a repo URI or numeric ID as a parameter. We've since
learned that it's much better for the gRPC API methods to accept a
single parameter. (See also the removal of RepoRevSpec.Rev.) So, the
potential for RepoSpec adding a numeric ID field in the future is
extremely low. So, let's clean it up!

Note: This lays the groundwork for switching from string repo
identifiers ("URIs") to numeric repo identifiers. Even if we don't
choose to use numeric repo IDs, this change is still worthwhile
because it simplifies the code. 

Test plan:

* Navigate to code view, def info, and builds pages. These are the UI pages that are affected. They should work exactly as they did before.
* Review the changes to sourcegraph.proto. No need to review all of the individual changes; they are rote.


sqs:
This touches a lot of the code in a small way, so let's get this merged ASAP and reduce merge conflicts.

@neelance Feel free to merge this if it looks good to you, or make small tweaks you think are necessary to get it merged more quickly. If there are big blockers and you aren't able to address them, that's fine; this is not urgent (it can wait).

neelance:
LGTM
@keegancsmith suggested that we may want to use a type alias on `string`. It's a good idea, although not seen very often in other Go libs.

shurcooL:
> @keegancsmith suggested that we may want to use a type alias on `string`. It's a good idea, although not seen very often in other Go libs.

I used to think that too, I've encountered some evidence to the contrary. It can become quite hard to find a package low level enough to contain that type. It has to be a low level package so that there are no import cycles created, since many things may need to reference that type. So, `builtin` and its `string` type work quite well, as long as you document it (and give it a good name).

sqs:
We used to use a RepoURI type and it was much more of a hassle than it was worth.

PR #17, opened 2016-06-03
Add Definfo filters

renfredxh:
This change introduces three categories on the def info page for refs. "Top" will show up to three usage example snippets from different repositories.

I also split out the backend/front-end stack for serving and displaying refs and examples into two separate paths. These are similar now, but will diverge over time, and trying to keep them the same was resulting in various hacks and bugs. 

##### Test plan

Local testing and on staging.sourcegraph.com

image


api/sourcegraph/sourcegraph.proto
```diff
@@ -1465,6 +1465,9 @@ message DefListRefLocationsOptions {
 	// ListOptions specifies options for paginating
 	// the result.
 	ListOptions ListOptions = 3 [(gogoproto.nullable) = false, (gogoproto.embed) = true, (gogoproto.jsontag) = ""];
+
+	// Temporary: optionally specify a specific ranking.
+	string Sorting = 4;
```
sqs: elsewhere we call this `Sort` not `Sorting`

app/web_modules/sourcegraph/def/DefInfo.js
```diff
@@ -70,44 +68,34 @@ class DefInfo extends Container {
...
-
 		if (prevState.defCommitID !== nextState.defCommitID && nextState.defCommitID) {
 			if (this.context.features.Authors) {
 				Dispatcher.Backends.dispatch(new DefActions.WantDefAuthors(nextState.repo, nextState.defCommitID, nextState.def));
 			}
 		}
 	}
 
+	_mergeQuery(param: string, value: any) {
+		return Object.assign({}, this.props.location.query, {[param]: value});
```
sqs: `return {...this.props.location.query, [param]: value};`

ES6 FTW

sqs: And in fact given that it's so simple in ES6, might not make sense to extract this to a separate method.

app/web_modules/sourcegraph/def/DefInfo.js
```diff
@@ -178,7 +160,6 @@ class DefInfo extends Container {
 								onItemClick={(val) => this._onTranslateDefInfo(val)}
 								items={[
 									{name: "English", value: "en"},
-									{name: "简体中文", value: "zh-CN"},
```
renfredxh: TODO: add this line back in

app/web_modules/sourcegraph/def/RefsContainer.js
```diff
@@ -180,12 +163,8 @@ export default class RefsContainer extends Container {
 	}
 
 	onStateTransition(prevState, nextState) {
-		// optimization: since multiple RefContainers may be shown on a page, fetching refs for every container
-		// when the component is mounted will cause unnecessary re-render cycles across components.
-		// Instead, lazily fetch ref data on mouseover.
 		const refPropsUpdated = prevState.repo !== nextState.repo || prevState.rev !== nextState.rev || prevState.def !== nextState.def || prevState.refRepo !== nextState.refRepo;
-		const initialLoad = !prevState.repo && !prevState.rev && !prevState.commitID && !prevState.def && !prevState.refRepo;
-		if ((initialLoad && nextState.prefetch) || (refPropsUpdated && !initialLoad) || (nextState.mouseover && !prevState.mouseover)) {
```
renfredxh: Note: removed this because it was causing a bug, and we don't really need it anymore.

shurcooL:
On https://staging.sourcegraph.com I see:

image

Has that been fixed already in code?

renfredxh:
@shurcooL yes that has been fixed

api/sourcegraph/sourcegraph.proto
```diff
@@ -1480,7 +1480,7 @@ message RefList {
 message DefListRefLocationsOptions {
 	// Repos is the list of repos to restrict the results to.
 	// If empty, all repos are searched for references.
-	repeated int32 Repos = 1 [(gogoproto.moretags) = "url:\",omitempty\""];
+	repeated string Repos = 1 [(gogoproto.moretags) = "url:\",omitempty\""];
```
shurcooL: Question, why is this type changed from `int32` to `string`?

api/sourcegraph/sourcegraph.proto
```diff
@@ -1747,6 +1757,9 @@ service Defs {
 	// ListRefLocations lists repos and files that refer to def.
 	rpc ListRefLocations(DefsListRefLocationsOp) returns (RefLocationsList);
 
+	// ListExamples lists usage examples for a given def.
```
shurcooL: After reading that description, it's not very clear to me how this endpoint differs from similar ones above.

```
ListRefLocations lists repos and files that refer to def.

ListExamples lists usage examples for a given def.
```

What is the difference between an example for a def vs repos/files that refer to def?

pkg/store/global_graph.go
```diff
@@ -38,6 +38,12 @@ type GlobalRefs interface {
 	Update(ctx context.Context, op *sourcegraph.DefsRefreshIndexOp) error
 }
 
+// Examples defines the interface for getting and listing def usage examples.
```
shurcooL: `Examples` -> `DefExamples`. That's the name of the interface.

services/httpapi/def_refs.go
```diff
@@ -119,56 +118,40 @@ func serveDefRefLocations(w http.ResponseWriter, r *http.Request) error {
...
-			lastIdx := len(refLocations.RepoRefs) - 1
-			refLocations.RepoRefs[0], refLocations.RepoRefs[lastIdx] = refLocations.RepoRefs[lastIdx], refLocations.RepoRefs[0]
-		}
+	refLocations, err := cl.Defs.ListExamples(ctx, &opt)
+	if err != nil {
+		return err
 	}
 
-	return writeJSON(w, refLocations)
+	return json.NewEncoder(w).Encode(refLocations)
```
shurcooL: Why did you stop using `writeJSON` here?

api/sourcegraph/sourcegraph.proto
renfredxh: It was recently changed from `string` to `int32` when we moved the DB from using URI-based primary keys to numeric keys. This is not ideal for the graph DB API because we're still referring to things by repo URI there. If we ever move the graph DB to using numeric IDs for repos in the future this could change, but for now this is the simplest option. I already chatted with @sqs about this to make sure it made sense.

api/sourcegraph/sourcegraph.proto
renfredxh: Related to my comments in the PR description:


> I also split out the backend/front-end stack for serving and displaying refs and examples into two separate paths. These are similar now, but will diverge over time, and trying to keep them the same was resulting in various hacks and bugs.

So right now, there's not much of a difference in the format of the data that is returned. The main difference is that the usage examples will be a ranked/auto-curated selection of ref locations, where ref locations is just a generic, unordered list (with filtering options). Not sure of a good way to communicate this in the API docs that wouldn't make it more confusing, but as the way we present examples change over time the difference will become more obvious.

services/httpapi/def_refs.go
renfredxh: It wasn't being using in the endpoint above, will change this to make it consistent.

shurcooL: I'd be happy to stop using `writeJSON`, btw, as long as it's a conscious decision with some rationale.

api/sourcegraph/sourcegraph.proto
shurcooL: Gotcha. How about something like:

```Go
// ListExamples lists usage examples for a given def,
// which is a ranked/auto-curated subset of ref locations.
```

shurcooL:
Thanks, you've addressed/answered all my questions.

LGTM, I've reviewed this at a high/medium level and see no issues that would prevent this from being merged from my side. Feel free to merge after you test it out and resolve anything that's left on your side.

renfredxh:
Squashed and merged this manually

PR #19, opened 2016-06-03
Analytics: Filter out E2E and other bots from analytics based on user agent

KattMingMing:
##### Description

It looks like we only implemented checks for `userAgentIsBot` when calling `logEvent` to Amplitude instead of performing that check at logEvent we can set a flag up front that tells amplitude not to send the events. We also shouldn't be logging any events to intercom for e2e tests as well, but intercom looks like it doesn't have a flag for this so we will have to opt out manually. 

@slimsag could you look this over to see if this would affect the e2e tests at all?

### DIFF
--- a/app/web_modules/sourcegraph/util/EventLogger.js
+++ b/app/web_modules/sourcegraph/util/EventLogger.js
@@ -86,6 +86,11 @@ export class EventLogger {
 		}
 
 		this.userAgentIsBot = Boolean(context.userAgentIsBot);
+
+		// Opt out of Amplitude events if the user agent is a bot.
+		if (this.userAgentIsBot) {
+			this._amplitude.setOptOut(true);
+		}
 	}
 
 	// User data from the previous call to _updateUser.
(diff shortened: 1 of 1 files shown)


app/web_modules/sourcegraph/util/EventLogger.js
```diff
@@ -178,12 +179,12 @@ export class EventLogger {
 
 	// sets current user's property value
 	setIntercomProperty(property, value) {
-		if (this._intercom) this._intercomSettings[property] = value;
+		if (this._intercom && !this.userAgentIsBot) this._intercomSettings[property] = value;
```
slimsag: This is just setting a property within the `_intercomSettings` map/dictionary/object, right? Surely it'd be harmless to do this, unless someone else is actively using/watching this map?

slimsag:
One minor comment inline, otherwise this change looks alright to me. Nothing here will affect e2etest (everything there will continue to work fine).

PR #20, opened 2016-06-04
Enable gitserver tests, remove unnecessary cases

sqs:
See full commit messages.

Test plan: manually check CI logs to ensure that app/internal/gitserver's tests were run.

PR #21, opened 2016-06-04
Simplify Go build tags used for testing

sqs:
Addresses the tech debt item "clean up test entrypoints, remove
mdtest/lgtest targets, and remove Go test build tags". /cc +1ers
@shurcool @renfredxh @keegancsmith

Now, instead of needing to specify Go build tags in "go test -tags
..." or needing to run the confusing "make mdtest" (etc.) targets, you
can just run "go test" (with the list of pkgs, of course). **The
default is to run all tests.**

As a secret bonus dev goodie, those who are familiar with "go test
-short" can use that to run an abbreviated set of tests. But that
won't be documented in any of our Sourcegraph READMEs, since it's a Go
feature, and since you should know what you're doing (and what
failures it might miss) if you choose to run it. 

This PR depends on https://github.com/sourcegraph/sourcegraph/pull/20,
which must be merged first.


dev/circle-ci-run-tests.sh
```diff
@@ -26,14 +26,15 @@ echo "$changed"
 
 pkgs=()
 covered=()
-for pkg in $(go list  -f '{{ if or (gt (len .TestGoFiles) 0) (gt (len .XTestGoFiles ) 0) }}{{ .ImportPath }}{{ end }}' ./... | grep -v /vendor/ | sort); do
+for pkg in $(go list -tags 'exectest buildtest pgsqltest nettest githubtest'  -f '{{ if or (gt (len .TestGoFiles) 0) (gt (len .XTestGoFiles ) 0) }}{{ .ImportPath }}{{ end }}' ./... | grep -v /vendor/ | sort); do
```
keegancsmith: why do you need to specify the tags here? I thought we got rid of them.

sqs: Yep. Whoops. That was a remnant from a previous commit to hotfix the immediate issue. A later commit removed the build tags elsewhere but did not account for those added here by that previous hotfix. The tags can be removed here.

keegancsmith: https://github.com/sourcegraph/sourcegraph/commit/1c7e0ae30a64ac690803b2e2e8353aa43479582f

PR #25, opened 2016-06-05
Make global repo search case insensitive

Unknwon:
Before:

image

After:

image

##### Reviewer tasks


rothfels:
@Unknwon lgtm

sqs:
The other approach is to name `owner` and `name` into PostgreSQL [`citext`](https://www.postgresql.org/docs/current/static/citext.html) columns, like the `uri` column. This seems to be better, since then this case-insensitive search behavior is the norm everywhere. We already use it for `uri` and it works quite well. Also, indexes on citext columns apply to these LOWER queries as well.

beyang:
Remember to run the migration if you use the `citext` type.

PR #27, opened 2016-06-06
prepare for numeric repo IDs

sqs:
These changes are OK to make now without causing any change in
behavior. They will reduce the size of the diff of the eventual
`repo-id` change.


sqs:
@rothfels FYI, I am going to merge this change, which changes the srclib coverage code in the `cli` package but should not affect behavior. It is merely paving the way for the switch to numeric repo IDs. If you notice any bad side effects, check here. I ran it and it seemed to work, but without tests, I can't know for sure.

PR #28, opened 2016-06-06
Queue for srclib imports

keegancsmith:
Fixes https://app.asana.com/0/37478073567611/139065576365312

Still to do

* [x] Remove `WithSkipInsecure` Hack!!
* [x] Permissions on store and Async service
* [x] Metrics around queue
* [ ] Alerts on queue
* [x] Bounded retries on failure
* [x] Queue logging
* [ ] Make it possible to just run the async worker

Migration SQL:
```
CREATE TABLE que_jobs
(
  priority    smallint    NOT NULL DEFAULT 100,
  run_at      timestamptz NOT NULL DEFAULT now(),
  job_id      bigserial   NOT NULL,
  job_class   text        NOT NULL,
  args        json        NOT NULL DEFAULT '[]'::json,
  error_count integer     NOT NULL DEFAULT 0,
  last_error  text,
  queue       text        NOT NULL DEFAULT '',

  CONSTRAINT que_jobs_pkey PRIMARY KEY (queue, priority, run_at, job_id)
);
```


services/backend/internal/localstore/queue.go
```diff
@@ -0,0 +1,93 @@
...
+	// TODO(keegancsmith) perm??
+	c, err := q.client(ctx)
+	if err != nil {
+		return err
+	}
+	return c.Enqueue(q.toQue(j))
+}
+
+func (q *queue) LockJob(ctx context.Context) (*store.LockedJob, error) {
+	// TODO(keegancsmith) perm??
```
beyang: Should require the same permissions level as the build queue, I think.

cli/serve_cmd.go
```diff
@@ -464,6 +468,26 @@ func (c *ServeCmd) Execute(args []string) error {
...
+			return err
+		}
+	}
+	asyncCtx = svc.WithServices(asyncCtx, server.Config(serverCtxFunc))
+	asyncCtx = accesscontrol.WithInsecureSkip(asyncCtx, true) // HACKS
+	asyncCtx, err = c.authenticateScopedContext(asyncCtx, idKey, []string{"internal:async"})
+	if err != nil {
+		return err
+	}
+	backend.StartAsyncWorkers(asyncCtx)
```
beyang: Make togglable via command line flag? That will come in handy if we're splitting user-facing and worker-facing front-end instances into 2 separate groups.

services/backend/async.go
```diff
@@ -0,0 +1,130 @@
...
+
+var Async sourcegraph.AsyncServer = &async{}
+
+type async struct{}
+type asyncWorker struct{}
+
+// StartAsyncWorkers will start async workers to consume jobs from the queue.
+func StartAsyncWorkers(ctx context.Context) {
+	w := &asyncWorker{}
+	for i := 0; i <= localcli.Flags.NumAsyncWorkers; i++ {
```
beyang: `i < localcli.Flags.NumAsyncWorkers` (off-by-one)

api/sourcegraph/sourcegraph.proto
```diff
@@ -1750,13 +1750,33 @@ service Defs {
...
 	// snapshot of the code graph. This operation is idempotent, and
 	// should be executed regularly to keep the indexes up-to-date.
 	rpc RefreshIndex(DefsRefreshIndexOp) returns (pbtypes.Void);
 }
 
+message AsyncRefreshIndexesOp {
+	// Repo will have all of its indexes refreshed
+	int32 Repo = 1;
+
+	// Source helps tie back async jobs to there source
```
beyang: s/there/their/

api/sourcegraph/sourcegraph.proto
```diff
@@ -1750,13 +1750,33 @@ service Defs {
...
-    	// RefreshIndex updates the precomputed indexes used for answering
+	// RefreshIndex updates the precomputed indexes used for answering
 	// ListRefLocations queries. The indexes are built from the current
 	// snapshot of the code graph. This operation is idempotent, and
 	// should be executed regularly to keep the indexes up-to-date.
 	rpc RefreshIndex(DefsRefreshIndexOp) returns (pbtypes.Void);
 }
 
+message AsyncRefreshIndexesOp {
+	// Repo will have all of its indexes refreshed
```
beyang: "Repo is the repository whose indexes this will trigger a refresh for."

api/sourcegraph/sourcegraph.proto
```diff
@@ -1750,13 +1750,33 @@ service Defs {
...
+	int32 Repo = 1;
+
+	// Source helps tie back async jobs to there source
+	string Source = 2;
+
+	// Force will ensure an index is refreshed, even if it has indexes the
```
beyang: "an index" or all indexes?

"even if the index already includes the latest commit"

services/backend/internal/localstore/queue.go
keegancsmith: @beyang we don't have repos associated with the queue, nor do we necessarily want the same level of access in general compared to builders. This is how we do it for builds
```
accesscontrol.VerifyUserHasReadAccess(ctx, "Builds.Create", newBuild.Repo)
```

For now we don't have much better, so I'll just check read access on a `nil` repo.

keegancsmith: we may want to extend the model such that async jobs happen on a repo?

beyang:
Ready for review yet? I'd like to get this in before the weekend, so we can prep the blog post to announce search.

keegancsmith:
> Remove WithSkipInsecure Hack

That is still not working, but I was working on it before finishing up yesterday so should be resolved soon. The other two tasks I'll followup with since they won't block this working nicely in production.

keegancsmith:
@beyang ready for review.

cli/serve_cmd.go
```diff
@@ -468,21 +467,30 @@ func (c *ServeCmd) Execute(args []string) error {
...
-	asyncCtx := clientCtx
-	// TODO(keegancsmith) setting the serverctx.Funcs seems bad, but we do
-	// need access to the store since Queue.LockJob is session
-	// based. Think of something better
-	for _, f := range serverctx.Funcs {
-		asyncCtx, err = f(asyncCtx)
+	// TODO(keegancsmith) async is the only user of this at the moment,
+	// but other background workers that need access to the store will
+	// likely pop up in the future. We need to make this less hacky
+	internalServerCtx := func(name string) (context.Context, error) {
```
keegancsmith: extra close review on this please

vendor/github.com/keegancsmith/que-go/que.go
```diff
@@ -0,0 +1,266 @@
...
+// avoid looping forever in case something is wrong.
+const maxLockJobAttempts = 10
+
+// Returned by LockJob if a job could not be retrieved from the queue after
+// several attempts because of concurrently running transactions.  This error
+// should not be returned unless the queue is under extremely heavy
+// concurrency.
+var ErrAgain = errors.New("maximum number of LockJob attempts reached")
+
+// TODO: consider an alternate Enqueue func that also returns the newly
```
beyang: Should this TODO be kept around or removed?

pkg/store/queue.go
```diff
@@ -0,0 +1,58 @@
...
+// instead we need to mark a job as finished. This allows us to pick up work
+// when processing fails on it.
+type Queue interface {
+	// Enqueue puts j onto the queue
+	Enqueue(ctx context.Context, j *Job) error
+
+	// LockJob removes a job from queue, or returns nil if there is no
+	// jobs. You must call LockedJob.MarkSuccess or LockedJob.MarkError
+	// when done.
+	LockJob(ctx context.Context) (*LockedJob, error)
```
beyang: Is the "Lock" part of `LockJob` the key descriptor here or is it more of an implementation detail? To me, `Dequeue` makes more sense.

cli/serve_cmd.go
```diff
@@ -464,6 +467,35 @@ func (c *ServeCmd) Execute(args []string) error {
 	}
 	go statsutil.ComputeUsageStats(usageStatsCtx, 10*time.Minute)
 
+	// TODO(keegancsmith) async is the only user of this at the moment,
+	// but other background workers that need access to the store will
+	// likely pop up in the future. We need to make this less hacky
+	internalServerCtx := func(name string) (context.Context, error) {
```
beyang: clearly mark this with a `// HACK`

services/backend/internal/middleware/inner/inner_middleware.go
```diff
@@ -193,6 +195,21 @@ func (s wrappedAnnotations) List(ctx context.Context, param *sourcegraph.Annotat
 	return
 }
 
+type wrappedAsync struct{}
+
+func (s wrappedAsync) RefreshIndexes(ctx context.Context, param *sourcegraph.AsyncRefreshIndexesOp) (res *pbtypes.Void, err error) {
+	start := time.Now()
+	ctx = trace.Before(ctx, "Async", "RefreshIndexes", param)
+	defer func() {
+		trace.After(ctx, "Async", "RefreshIndexes", param, err, time.Since(start))
```
beyang: Do you think it would be helpful to store the time it takes to complete a job in the DB? This would be more discoverable  than the logs and could help identify bottlenecks (I suspect that refreshing indexes on the Go stdlib takes 40-60 minutes right now).

services/backend/internal/localstore/queue.go
```diff
@@ -0,0 +1,119 @@
...
+		errFunc = func(reason string) error {
+			log15.Debug("store.Job.MarkError ignoring", "type", j.Type, "lastReason", j.LastError.String, "reason", reason)
+			return j.Delete()
+		}
+	}
+
+	return store.NewLockedJob(q.fromQue(j), j.Delete, errFunc), nil
+}
+
+func (q *queue) Stats(ctx context.Context) (map[string]store.QueueStats, error) {
```
beyang: `accesscontrol.VerifyUserHasReadAccess`?

pkg/store/queue.go
keegancsmith: LockJob is a better word than Dequeue, since it does not remove the Job from the queue. Instead it locks it, preventing any other worker from taking the job. When it is done with the job you call `MarkSuccess` or `MarkError`. The former removes from the queue, while the latter unlocks and puts back into the queue.

vendor/github.com/keegancsmith/que-go/que.go
keegancsmith: this is vendored code, and yes it seems the todo can be removed. The vendored code is a fork to support our use case, this comment is from upstream.

services/backend/internal/middleware/inner/inner_middleware.go
keegancsmith: (Note: this is generated code for all gRPC endpoints)

We have the times stored in:
* Splunk (reliable, but not properly indexed yet for long timespan queries)
* Appdash (not that reliable yet)
* Aggregates in prometheus

I think we need to educate people on how to use splunk, since it is has a very powerful query language, which feels much more suited for aggregation of statistics than a DB.

Given this should be a low-volume API call, we could look into storing it in a table (and possibly other low volume calls). However, that should be a follow-up PR.

keegancsmith:
@beyang responded to feedback / added commits. PTAL

beyang:
LGTM

keegancsmith:
Added table to all environments. Landing now. Will monitor closely.

keegancsmith:
I deployed this to prod, and we started OOMing. I have no idea why this would cause ooms, but investigating further.

pkg/store/queue.go
shurcooL:
```Go
// LockJob removes a job from queue, or returns nil if there is no
// jobs. You must call LockedJob.MarkSuccess or LockedJob.MarkError
// when done.
```

It sounds like the processing can take a while. What's the `context.Context` that is expected to be passed? Will it be valid throughout the whole time, until job processing is done and `MarkSuccess` or `MarkError` is called?

keegancsmith: * yes we have a different context than usual that is passed in for async jobs.
* We have no deadlines set in the grpc layer anyways.

The context expiring is in general an issue. Luckily though, the db handle is still retrievable in the case the context is expired.

PR #29, opened 2016-06-06
coverage: add bash

mkovacs:
Compute coverage for `srclib-bash`.

### DIFF
--- a/pkg/coverageutil/bash.go
+++ b/pkg/coverageutil/bash.go
@@ -1,0 +1,53 @@
+package coverageutil
+
+import (
+	"bytes"
+	"fmt"
+
+	"github.com/mkovacs/bash/scanner"
+)
+
+// bashIdentScanner extracts identifiers from a Bash script
+type bashIdentScanner struct {
+	is     *scanner.Scanner
+	errors []string
+}
+
+// Init initializes text scanner that extracts only idents
+func (s *bashIdentScanner) Init(src []byte) {
+	s.is = &scanner.Scanner{}
+	s.is.Init(bytes.NewReader(src))
+}
+
+func (s *bashIdentScanner) Done() {
+}
+
+func (s *bashIdentScanner) Errors() []string {
+	return []string{}
+}
+
+// Next returns the next identifier
+func (s *bashIdentScanner) Next() *Token {
+	for {
+		tok, err := s.is.Scan()
+		if err != nil {
+			s.errors = append(s.errors, fmt.Sprintf("error: %s", err))
+			return nil
+		}
+		switch {
+		case tok == scanner.EOF:
+			return nil
+		case tok == scanner.Ident:
+			text := s.is.TokenText()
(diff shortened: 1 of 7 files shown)


pkg/coverageutil/bash.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/mkovacs/bash/scanner"
+)
+
+// bashIdentScanner extracts identifiers from a Bash script
+type bashIdentScanner struct {
+	is *scanner.Scanner
+}
+
+// Initializes text scanner that extracts only idents
```
rothfels: Usually go comment style is: `// Init initializes text scanner...` (i.e. start the comment with the name of the function).

pkg/coverageutil/bash.go
```diff
@@ -0,0 +1,51 @@
...
+func (s *bashIdentScanner) Init(src []byte) {
+	s.is = &scanner.Scanner{}
+	s.is.Init(bytes.NewReader(src))
+}
+
+func (s *bashIdentScanner) Done() {
+}
+
+func (s *bashIdentScanner) Errors() []string {
+	// TODO(mate): actually record errors
```
rothfels: Want to finish this TODO? I think you should just be able to add a field to your `bashIdentScanner` struct called `errors []string` and then in your `Next` function if there's any error, add the stringified error to the struct's slice.

pkg/coverageutil/bash.go
```diff
@@ -0,0 +1,51 @@
...
+	return []string{}
+}
+
+// Next returns the next identifier
+func (s *bashIdentScanner) Next() *Token {
+	for {
+		tok, err := s.is.Scan()
+		if err != nil {
+			return nil
```
rothfels: E.g. here (for error reporting)

pkg/coverageutil/bash.go
```diff
@@ -0,0 +1,51 @@
...
+			return &Token{uint32(p.Offset - len(text)), p.Line, text}
+		}
+	}
+}
+
+func init() {
+	var factory = func() Tokenizer {
+		return &bashIdentScanner{}
+	}
+	newExtensionBasedLookup("Bash", []string{".sh"}, factory)
```
rothfels: add ".bash"

pkg/coverageutil/bash.go
```diff
@@ -0,0 +1,51 @@
...
+		case tok == scanner.Ident:
+			text := s.is.TokenText()
+			p := s.is.Pos()
+			return &Token{uint32(p.Offset - len(text)), p.Line, text}
+		}
+	}
+}
+
+func init() {
+	var factory = func() Tokenizer {
```
rothfels: more "canonical" Go: `factory := funct() ...`

rothfels:
A few comments. Do you want to add a test for this? The other toolchains all have example tests, and there's a helper function which should make this super easy.

PR #30, opened 2016-06-06
fix for amplitude tracking on first sfye pageload

uforic:
I noticed that on the first pageloads from sfye, the utm source was not being listed. Digging into it, it wasn't being set in one place.

##### Reviewer tasks

I'd love to know if there is a better way to do this URL assembly at this point, tacking on &X=A&Y=B seems error prone.

##### Test plan

Run locally, I verified that it works by stepping through with the JS debugger.


KattMingMing:
++

PR #31, opened 2016-06-06
fix external ref fetches

sqs:
Fixes the issue where pages like https://sourcegraph.com/github.com/aws/aws-sdk-go/-/info/GoPackage/github.com/aws/aws-sdk-go/private/util/-/GoFmt do not load external ref snippets.


sqs:
merged from CLI

PR #32, opened 2016-06-07
Analytics: Set isEmployee flag

KattMingMing:
##### Description

Setting isEmployee flag and resetting analytics opt out based on the user. It is not absolutely necessary, but if a user was marked as a bot it persists until explicitly set to false. This could be problematic when debugging analytics. For example, if you were marked as a bot while running CI and then wanted to check dev analytics at a later date to confirm everything was firing correctly it would not be immediately obvious why your user isn't firing events.


uforic:
LGTM - I reviewed that Boolean(undefined) is false, and that map["entry_does_not_exist"] returns "undefined", which will not be true in an if statement.

PR #33, opened 2016-06-07
Mark build as failure when key language fails

beyang:
Currently, none of the srclib indexers are required to complete successfully for a build to be marked as successful. This change makes it so that if any of `srclib-{go,java,python,csharp}` fails AND the failing language is greater than 33% of the code in the repository, the build is marked as a failure. Marking the build as failure will also prevent the graph database from updating the graph data version, so should prevent a lot of srclib coverage regressions in production.

This fixes https://github.com/sourcegraph/srclib-python/issues/44

##### Test plan

Tested locally.


rothfels:
LGTM

PR #34, opened 2016-06-07
SEO: server-side-rendered <title>, og tags, meta tags, canonical URLs, meta robots noindex/nofollow

sqs:
This PR makes it so that all of the following are rendered and returned in the initial HTML response from the server:

* HTTP status code (200/404/etc.)
* `<title>`
* Open Graph tags
* meta description tags
* Twitter card tags
* `<link rel=canonical>`
* meta robots noindex/nofollow

This should improve SEO.

It adds a new package app/internal/ui that mimics the route structure of the JavaScript app. For certain routes (blob, def, definfo, tree, repo, etc.), it attempts to fetch the relevant main data in Go prior to rendering the page. If an error occurs, a non-200 HTTP status code is returned. Otherwise, the above tags are constructed in Go on the server, so that search engine crawlers needn't render the page in JavaScript.

SEO notes:

* **Only definfo and static pages are indexed,** not repo/blob/def/etc. Also, only defs meeting certain criteria are indexed (see the `shouldIndexDef` func); e.g., name at least 3 chars, kind is `type` or `func`, exported.

#### Test plan

It's up at https://staging.sourcegraph.com. The package is well tested (94.9% test coverage), but try:

* Examining the HTML source for various routes. Make sure the meta description and (especially) the meta robots noindex/nofollow is set properly.
* Going to pages you expect to 404, and checking that the initial HTML response has a status code of 404.
* Pasting URLs to staging.sourcegraph.com into Slack, Twitter, Facebook, etc., to make sure the Open Graph data looks good.

#### Screenshots!

Open Graph:

image


uforic:
Heading here:
https://staging.sourcegraph.com/github.com/sourcegraph/appdash/-/def/GoPackage/github.com/sourcegraph/appdash/-/NewLocalCollector

I noticed two rel canonicals in the head section (could get different canonicals by navigating to different sites). I bet this has to do with Helmet / one-page apps. Screenshots attached:
screen shot 2016-06-07 at 9 36 57 am
screen shot 2016-06-07 at 9 38 36 am

Another q: Dmitri and I were thinking about the way we do canonical for a site like:
https://staging.sourcegraph.com/github.com/sourcegraph/appdash@master/-/def/GoPackage/github.com/sourcegraph/appdash/-/NewLocalCollector

And it struck us that instead of pointing it to:
https://staging.sourcegraph.com/github.com/sourcegraph/appdash/-/def/GoPackage/github.com/sourcegraph/appdash/-/NewLocalCollector , does it make more sense to point it (and all defs) at:
https://staging.sourcegraph.com/github.com/sourcegraph/appdash/-/blob/collector.go ?

Given that all the content on the various def pages will basically be the same (the blob stuff). Out of scope of this PR, just flagging it.

uforic:
screen shot 2016-06-07 at 9 46 36 am

There are two descriptions appearing here - one from the initial page load and one from future page loads.

Also, we should standardize on the description text: the Helmet one we added was:
"Code and usage examples for strings.SplitN in golang/go. SplitN slices s into substrings separated by sep and returns a slice of
the substrings between those s…"
(whitespace in the middle a bug, as well).

The one from the server is here: 
SplitN slices s into substrings separated by sep and returns a slice of
the substrings between those separators.
If sep is empty, SplitN splits after each UTF-8 sequence.
The count determines the number of substrings to return:

n > 0: at most n sub...

From MOZ:
Best under 155 characters
Each description should be unique
Well written descriptions influence
click-through rate

sqs:
* [ ] Def canonical URLs = blob URL.
* [ ] Remove JavaScript/Helmet-set meta tags (except for title). Avoid duplicating. And the initial page load's meta tags are all that matter.
* [ ] Tighten up description.

uforic:
I noticed on staging, when I navigate to this URL (via the various menus and things):
https://staging.sourcegraph.com/github.com/gorilla/mux@9fa818a44c2bf1396a17f9d5a3c0f6dd39d2ff8e/-/def/GoPackage/github.com/gorilla/mux/-/MatcherFunc

Canonical is pointed to:
https://staging.sourcegraph.com/github.com/gorilla/mux (my initial entry point)
screen shot 2016-06-07 at 4 14 08 pm

Are we expected the canonical link to change as you navigate through the page?

Direct entry works great, though!

uforic:
Ah, per your comment - we are doing the initial page loads, and follow on ones don't matter. We are sure that Google does individual page loads for single page apps (doing a fresh GET every time)?

If so, sounds good!

sqs:
> We are sure that Google does individual page loads for single page apps (doing a fresh GET every time)?

No, not sure, but the amount of effort for any site to do this perfectly is so large that I imagine Google MUST handle this case.

uforic:
https://staging.sourcegraph.com/github.com/gorilla/mux/-/info/GoPackage/github.com/gorilla/mux/-/CurrentRoute - Description has a newline. Low priority, I think the doc strings must have newlines in them. 

https://www.drupal.org/node/1978568 - mentions that OG might break if description has a newline.

I think if we replace \n, \r, and \t. it's a safe bet.

uforic:
https://staging.sourcegraph.com/github.com/golang/go@41dd1696ab13755bf7a129e0c73523ffb9fcbe66/-/info/GoPackage/strings/-/Compare

Just want to double check that the above link should refer to itself as canonical, not https://staging.sourcegraph.com/github.com/golang/go/-/info/GoPackage/strings/-/Compare

I noticed that the @41.... is in fact @master, so I guess in theory @41 and @master should point to /golang/go/ as canonical, but building this logic probably isn't necessary since we are hoping the crawler will be entering at the golang/go/ addresses.

uforic:
LGTM - double checked the logic on defInfo, and it looks like @<non-master> is no-index, and points to itself as canonical, @master is index, and points to the non-revisiioned URL as canonical, which is solid.

Blob and repo views are non-index, solid.

sqs:
> Just want to double check that the above link should refer to itself as canonical, not https://staging.sourcegraph.com/github.com/golang/go/-/info/GoPackage/strings/-/Compare

I think the behavior is correct, or at least enough in the margin of unknowability that it's fine.

PR #35, opened 2016-06-07
Dealing with C# preprocessor directives

alexsaveliev:
Ignoring preprocessor directives in C# code when extracting identifiers to reduce number of uncovered idents in C# code (spotted by @ildarisaev)

##### Test plan

Covered by unit test


rothfels:
LGTM

PR #36, opened 2016-06-07
Update list of javascript repos for coverage script run

antonina-cherednichenko:
For now commented out repos without package.json file.
Also commented out one weird repository with no javascript files, except configuration file.


rothfels:
@antonina-cherednichenko LGTM, but why don't we just remove the commented out repos? I don't think they're adding value to hang around in there.

rothfels:
@antonina-cherednichenko actually, I'll just merge this and make the update myself. Thanks.

antonina-cherednichenko:
Repos without package.json can be supported in the future - the simplest way for such case - extend scan to just include all javascript files. It's good for not to forget adding them back in this case. Repo without js file can be easily deleted.

PR #37, opened 2016-06-07
WIP: firefox build script takes place in docker

uforic:
"npm run build-docker" to build the docker image, then "npm run build-firefox"

firefox-build.zip should appear in the browser-ext directory.

##### Reviewer tasks

Naming / directory structure: I created a new folder in scripts, with the stuff specific to the firefox build. Is this reasonable?

##### Test plan

Run it on someone else's computer, does it work first try?

TODO: diff the zip file to the one created locally on the mac, and make sure they are different.

### DIFF
--- a/client/browser-ext/package.json
+++ b/client/browser-ext/package.json
@@ -7,7 +7,9 @@
     "build": "node scripts/build",
     "compress": "node scripts/compress",
     "compress-keygen": "crx keygen",
-    "clean": "rimraf build/ dev/ *.zip *.crx"
+    "clean": "rimraf build/ dev/ *.zip *.crx",
+    "build-docker": "docker build -t sourcegraph/browser-ext-build scripts/build-env",
+    "dist": "npm run build-docker && docker run -v `pwd`:/browser-ext -i -t sourcegraph/browser-ext-build /bin/bash /browser-ext/scripts/build-env/dist.sh"
   },
   "author": "John Rothfels <john@sourcegraph.com>",
   "license": "MIT",
--- a/client/browser-ext/.gitignore
+++ b/client/browser-ext/.gitignore
@@ -8,4 +8,5 @@ dev/
 *.zip
 *.crx
 *.pem
+*.xpi
 update.xml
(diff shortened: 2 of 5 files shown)


client/browser-ext/package.json
```diff
@@ -7,7 +7,9 @@
     "build": "node scripts/build",
     "compress": "node scripts/compress",
     "compress-keygen": "crx keygen",
-    "clean": "rimraf build/ dev/ *.zip *.crx"
+    "clean": "rimraf build/ dev/ *.zip *.crx",
+    "build-docker": "cd scripts/firefox-build && ./build-docker.sh",
```
rothfels: I would be annoyed if every time I ran an npm command my directory was changed. :P

How about just inline `docker build -t sourcegraph/firefox-build .`?

^^ on that note, there's nothing firefox specific about this, so maybe call the image `sourcegraph/browser-ext-build`

client/browser-ext/package.json
```diff
@@ -7,7 +7,9 @@
...
+    "build-firefox": "docker run -v /Users/mattfs/gopath/src/sourcegraph.com/sourcegraph/sourcegraph/client/browser-ext:/browser-ext -i -t sourcegraph/firefox-build /bin/bash /browser-ext/scripts/firefox-build/build-firefox.sh"
```
rothfels: This file path will only work for you. Make it a relative URL.

rothfels: Also, this doesn't need to be firefox specific right? It could do the work of zipping up a firefox AND chrome extension. #would-be-awesome.

^^ on that note, usually this step would be called `dist` -- so `npm run dist` would probably first run `npm run build-docker` and then create your artifacts.

client/browser-ext/scripts/firefox-build/build-docker.sh
```diff
@@ -0,0 +1 @@
+docker build -t sourcegraph/firefox-build .
```
rothfels: I think would be better just to inline in `package.json`

PR #38, opened 2016-06-07
srclib_support: use basic_formatter for Bash

mkovacs:
This is needed so that Bash definitions are rendered correctly.


rothfels:
LGTM

PR #41, opened 2016-06-08
services/backend/internal/localstore: fix GlobalDefs.Search with Repos / NotRepos queries

slimsag:
One thing missed in Quinn's [PR #26](https://github.com/sourcegraph/sourcegraph/pull/26) was the `GlobalDefs.Search` method which allows filtering by `Repos` and `NotRepos` fields. These fields were change to `int32` repo ID types, but because our `arg` SQL pattern takes `interface{}` it blindly worked without a compiler error. Thus, when you attempt to search by `Repos` `[1, 2, 3]` it compares against _repo URIs_ 1,2,3 NOT repo IDs 1,2,3 like it should.

Comparing against repo IDs requires a DB migration, which this PR does not perform. Instead, we just resolve the `Repos` and `NotRepos` repo URIs upon a call to `GlobalDefs.Search` and then run the query using repo URIs per usual. This is a stop-gap solution for a different PR I am working on, and we can switch to storing repo IDs in the `global_defs` table at a later date.

##### Reviewer tasks

- [x] HACKS: reviewer approves hacks introduced by this change
  - Not really a hack, just a stopgap solution. But please confirm this is OK.
  - Not sure investing in a test at this point is entirely valuable, this code could break in many more ways than can be tested easily / in a timely mannor.

##### Test plan

Locally tested. To my knowledge, nobody currently uses the `Repos` / `NotRepos` filters yet (I am with my future PR, though).

### DIFF
--- a/services/backend/internal/localstore/global_defs.go
+++ b/services/backend/internal/localstore/global_defs.go
@@ -150,15 +150,23 @@ func (g *globalDefs) Search(ctx context.Context, op *store.GlobalDefSearchOp) (*
 	{
 		var wheres []string
 		if len(op.Opt.Repos) > 0 {
+			reposURIs, err := repoURIs(ctx, op.Opt.Repos)
+			if err != nil {
+				return nil, err
+			}
 			var r []string
-			for _, repo := range op.Opt.Repos {
+			for _, repo := range reposURIs {
 				r = append(r, arg(repo))
 			}
 			wheres = append(wheres, `repo IN (`+strings.Join(r, ", ")+`)`)
 		}
 		if len(op.Opt.NotRepos) > 0 {
+			notReposURIs, err := repoURIs(ctx, op.Opt.NotRepos)
+			if err != nil {
+				return nil, err
+			}
 			var r []string
-			for _, repo := range op.Opt.NotRepos {
+			for _, repo := range notReposURIs {
 				r = append(r, arg(repo))
 			}
 			wheres = append(wheres, `repo NOT IN (`+strings.Join(r, ", ")+`)`)
@@ -422,4 +430,15 @@ func resolveRevisionDefaultBranch(ctx context.Context, repo int32) (repoPath, co
 		return
 	}
 	return repoObj.URI, string(c), nil
+}
+
+func repoURIs(ctx context.Context, repoIDs []int32) (uris []string, err error) {
+	for _, repoID := range repoIDs {
+		repo, err := store.ReposFromContext(ctx).Get(ctx, repoID)
+		if err != nil {
+			return nil, err
+		}
+		uris = append(uris, repo.URI)
+	}
+	return uris, nil
 }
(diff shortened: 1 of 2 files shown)


shurcooL:
An observation here is that this will cause an increased load on us hitting the GitHub API, since there are additional `Repo.Get` calls for GitHub repos.

It seems that it should be a small/moderate increase that we can handle, but it's worth being aware of that.

> This is a stop-gap solution for a different PR I am working on, and we can switch to storing repo IDs in the `global_defs` table at a later date.

What's the expected time frame for that change? Will it revert the extra load on hitting the github API that this adds?

Changes LGTM.

beyang:
I don't think the impact on the GitHub API rate limit should be substantial, because we cache all GitHub requests for 10 minutes.

sqs:
LGTM

services/backend/internal/localstore/global_defs.go
```diff
@@ -150,15 +150,23 @@ func (g *globalDefs) Search(ctx context.Context, op *store.GlobalDefSearchOp) (*
 	{
 		var wheres []string
 		if len(op.Opt.Repos) > 0 {
+			reposURIs, err := repoURIs(ctx, op.Opt.Repos)
```
shurcooL: Tiny nitpick, `reposURIs` -> `repoURIs`. No need to pluralize both words.

services/backend/internal/localstore/global_defs.go
```diff
@@ -150,15 +150,23 @@ func (g *globalDefs) Search(ctx context.Context, op *store.GlobalDefSearchOp) (*
...
+			}
 			var r []string
-			for _, repo := range op.Opt.Repos {
+			for _, repo := range reposURIs {
 				r = append(r, arg(repo))
 			}
 			wheres = append(wheres, `repo IN (`+strings.Join(r, ", ")+`)`)
 		}
 		if len(op.Opt.NotRepos) > 0 {
+			notReposURIs, err := repoURIs(ctx, op.Opt.NotRepos)
```
shurcooL: Tiny nitpick, `notReposURIs` -> `notRepoURIs`.

slimsag:
> What's the expected time frame for that change? Will it revert the extra load on hitting the github API that this adds?

For the first question, I don't know the timeframe / how high of a priority it is. But, that being said, `global_defs` DB table [already has some optimizations going on](https://github.com/sourcegraph/sourcegraph-issues/issues/2) and this might fall hand-in-hand with those.

services/backend/internal/localstore/global_defs.go
slimsag: Actually this was intentional, `reposURIs` is "URIs for op.Opt.Repos" is how I intend for this to be read (and makes more sense IMO). Likewise for you comment below, `notRepos` refers to `op.Opt.NotRepos` and keeping the naming similiar makes the mapping easier to understand (for me, at least).

Will leave these as-is, but if you feel strongly about it I can change in a follow-up!

shurcooL: I agree with your assessment 👍 , thanks for considering this point!

PR #42, opened 2016-06-08
temporary fix while sourcegraph.com resolving is returning github.com

uforic:
There is an Asana bug for this, but basically URLs are getting resolved as:
https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/info/GoPackage/github.com/sourcegraph/sourcegraph/pkg/csp/-/Policy?editor_type=sublime&utm_source=sourcegraph-editor

Instead of:
https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/info/GoPackage/sourcegraph.com/sourcegraph/sourcegraph/pkg/csp/-/Policy?editor_type=sublime&utm_source=sourcegraph-editor

And there are 404s using SFYE. This appears to be specific to sourecgraph.com libraries.

##### Reviewer tasks

##### Test plan

will put on a staging instance, and verifiy that SFYE works.

### DIFF
--- a/app/web_modules/sourcegraph/misc/golang.js
+++ b/app/web_modules/sourcegraph/misc/golang.js
@@ -31,9 +31,14 @@ export const route: Route = {
 			defaultFetch(`/.api/resolve-custom-import/info?def=${def}&pkg=${pkg}&repo=${repo}`)
 				.then((resp) => resp.json())
 				.then((data) => {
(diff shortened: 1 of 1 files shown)


uforic:
I just tried running against staging3, and sourecgraph build wasn't done, but the URL construction was correct.

rothfels:
<3

PR #43, opened 2016-06-08
Analytics: Chrome extension improved tracking

KattMingMing:
###### Description

Adds lightweight tracking to the chrome browser extension

##### Test plan

Manually tested each event_type and reviewed each for accuracy. Also confirmed the event_properties were correct and provided relevant contextual information.

### DIFF
--- a/client/browser-ext/app/analytics/EventLogger.js
+++ b/client/browser-ext/app/analytics/EventLogger.js
@@ -1,0 +1,32 @@
+import React from "react";
+
+export class EventLogger {
+	_amplitude: any = null;
+
+	constructor() {
+		if (global.window && !this._amplitude) {
+			this._amplitude = require("amplitude-js/amplitude");
+
+			let apiKey = "34df64abd14304b388f884b46e0abbe2";
+			if (process.env.NODE_ENV === "production") {
+				apiKey = "e3c885c30d2c0c8bf33b1497b17806ba";
+			}
+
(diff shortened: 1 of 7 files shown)


uforic:
LGTM.

I built the plugin, and installed it on Chrome. I tried files with Go, and files without Go. Behaved as expected.

I wasn't able to get the noisy logging working, so I couldn't tell if the events were tracking. window.localStorage["event-log"]=true , but the console.log in logEvent wasn't getting fired. Was there an instance where I can check if the events were logged?

I also noticed the attached screenshot, when navigating to a non-code page. 
screen shot 2016-06-07 at 7 51 51 pm

client/browser-ext/app/analytics/EventLogger.js
```diff
@@ -0,0 +1,32 @@
...
+					includeUtm: true,
+					includeReferrer: true,
+					savedMaxCount: 50,
+			});
+		}
+	}
+
+	logEvent(eventName, eventProperties) {
+		eventProperties = eventProperties ? eventProperties : {};
+		eventProperties["Platform"] = "ChromeExtension";
```
sqs: This code is also bundled into the Firefox extension, so it shouldn't refer to Chrome.

uforic:
LGTM! Tested it out, mounting is working (and on refresh), hover is working, and referrer source is working (when we click through to Sourcegraph.com via usage examples).

PR #44, opened 2016-06-08
srclib-bash: update deps and docker image hash

mkovacs:
Updating for changes from `srclib-bash` and `bash/scanner`.


rothfels:
LGTM

PR #48, opened 2016-06-08
Simplify build configuration for javascript toolchain

antonina-cherednichenko:
Remove useless node install in configure build. It will be done in toolchain itself, inside scan command.


sqs:
The benefit of doing it here (and letting users override it in a `.drone.yml` in their own repo) is that if there are additional manual steps needed for installation, or they need to run `npm install` with flags (`--ignore-scripts`, `--no-optional`, `--production`, etc.), then they can do that. Also, in general, if a srclib toolchains assumes that the user has already installed all dependencies, it lets the same srclib toolchain work both locally (perhaps using the systemwide/global deps) and on Sourcegraph (where Sourcegraph is responsible for installing the right deps). Is it possible to change srclib-javascript so that it does not run `npm install` itself but instead assumes it has already been run?

rothfels:
@antonina-cherednichenko any comments on Quinn's message?

antonina-cherednichenko:
It's question of architecture, but I am more for idea of keeping javascript toolchain independent and decoupled from sourcegraph run. Removing npm install from toolchain will complicate development process, because tern requires node_modules to be present for resolving external dependencies. So developer must always remember to run npm install locally before running srclib do-all for example, run it after repo update, otherwise it wouldn't produce any errors, but won't resolve some identifiers. From architectural point of view there is a question - why tool which is in charge of resolving defs and refs is dependent on tool for visual representation regarding this resolution? Maybe flags from user can be passed to the toolchain?

sqs:
It's not that srclib-javascript is dependent on Sourcegraoh, just that it assumes something has already done the dependency installation.

I trust your judgment here, so LGTM!

rothfels:
@antonina-cherednichenko would you like to merge this?

antonina-cherednichenko:
yes

sqs:
@Rothfels Please merge

PR #49, opened 2016-06-08
pgsql: create index if not exists

keegancsmith:
Follow the same pattern for table creation, etc and only create the index if it
does not exist. This will prevent many errors spewing in logs when running
`src pgsql create`. There are still errors for triggers, since there is no
support for conditional creation on triggers yet.


keegancsmith:
Seems to be a 9.5 feature of postgresql, which is why CI is failing (which is on 9.4 + production). Abandon.

PR #50, opened 2016-06-08
Coverage: Ability to trace uncovered files

alexsaveliev:
Added `--empty` flag to display files with 0 refs, may be useful if there is a need to see pattern - what files produce no refs and what they have in common.


rothfels:
LGTM, merging.

PR #51, opened 2016-06-08
Updated sourcegraph/drone-git Docker image

alexsaveliev:
Updated sourcegraph/drone-git Docker image to add lost possibility to tweak `.gitattributes` (see also sqs/drone-git#4). Without this feature there may be highlighting issues for repositories with forced line endings

##### Test plan

Tested locally on github.com/SonyWWS/ATF


sqs:
Looks like an ephemeral CI failure. Retriggered.

rothfels:
@alexsaveliev this has continued to fail after a number of retries. Did you publish the new image to docker hub?

sqs:
@alexsaveliev and I chatted about it. There is an inexplicable issue that he is going to need to debug further. It may be a Docker version incompatibility, but the solution was not obvious. He is aware.

alexsaveliev:
It turned out that my Docker was too new (1.11). Downgraded it to 1.9 and rebuilt Docker image, CircleCI check passes now

PR #52, opened 2016-06-08
builds: Use external auth tokens

keegancsmith:
This should allow us to build private repos that need to `go get` other
private repos off github.

Fixes https://app.asana.com/0/87040567695724/140728200807475


keegancsmith:
This actually has an issue in the migration. If a build is created with the old code `user_id` is null, which will lead to

```
sql: Scan error on column index 3: converting driver.Value type <nil> ("<nil>") to a int: invalid syntax
```

I'll update to prevent nil users.

sqs:
There's a security issue here. If Alice wanted to see Bob's private repos, Alice could grant Bob access to an innocuous private repo ("hey Bob, want to join the beta of my app?") and then send Bob a link to an unbuilt commit ID on the private repo. Bob's visiting that URL in his browser would trigger a build of that commit ID, and the build would use his credentials. Alice could make the build run `git clone https://github.com/bob/secret.git`, which is a repo Alice doesn't have access to, and then mail her the contents of the repo.

keegancsmith:
@sqs Nice catch! I don't see how we can resolve this with our current knowledge of who can see what. I researched to see how CircleCI does this and per project you can grant CircleCI permission to add a private key as you.

```
Add user key

If a deploy key can't access all of your project's private dependencies, we can configure it to use an SSH key with the same level of access to GitHub repositories that you have.

In order to do so, you'll need to grant authorization from GitHub to the "admin:public_key" scope. This will allow us to add a new authorized public key to your GitHub account.
```

We can take our current approach, but make a user explicitly opt-in to use their keys on a repo? I don't know how we will detect that a project fails due to private deps, nor where the best place to configure this is. Either way it pushes this change into user facing configuration.

Another option is to remember who adds a private repo, and always use them. This still feels like a security risk since the person adding the project may not realise they are giving away the keys to their kingdom. So we would have to communicate that clearly + find a way to get users who have already connected there private repos to enable this feature.

@rothfels I'd love to pair with someone on getting this task done properly, I can take care of the backend stuff then someone does all the React/Frontend magic.

keegancsmith:
Abandoning this for now, will open a new PR instead for the new approach.

PR #53, opened 2016-06-08
change in the list of c# repos for coverage builds

ildarisaev:
Corefx and roslyn repos are excluded as causing timeouts (https://sourcegraph.com/github.com/dotnet/corefx/-/builds/34, https://sourcegraph.com/github.com/dotnet/roslyn/-/builds/71). Mono is excluded as well.

The repos that result in  > 85% refs coverage are placed at the top of the list.


rothfels:
@ildarisaev LGTM, merging.

PR #58, opened 2016-06-08
browser-ext: auto-create repos and refresh annotation data

rothfels:
+ if the user is viewing a supported file type (Go/Java), ensure the repo has been cloned
+ refereshVCS on interval and trigger re-render cycle when new (annotation) data is available
+ include visual build indicator on file header

Test plan:

- create a new github repository with a single go file (e.g. https://github.com/rothfels/dummygo)
- click on the .go file
- see that the build indicator says "Indexing"...
- ^^ wait for 1-2 minutes, see that annotation data eventually shows up on the page & the build indicator is updated


uforic:
Sick! I did the test plan, and it was Indexed in 10 seconds (very small project). Looking through code now.

uforic:
- [ ] In terms of getting rid of uncaught errors, I'm seeing these: inject.js?6870:476 no line number element to anchor def info to. Any chance we could catch these, as well?

- [ ] When the page first loads, it goes from "Indexing" to "Indexed", even if the page was Indexed to begin with. Looks a little jittery / janky. OK if we want this to be the behavior, indicates that we are "working on something", but if we could have a message "Checking status" that transitions, could be better. Thoughts?

uforic:
Last two things:
- [ ] We say "unsupported file" when it isn't a file we handle. Could we provide a popup that says "Sourcegraph indexes Go, Java, and Python", to be a little more specific? 
- [ ] I checked in code that was purposefully broken, and the status never changes from "Indexing", even on a page refresh. Would be sick to say "errors while indexing"

rothfels:
The "no line number element to anchor def info to" is unrelated to these changes (see: https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/client/browser-ext/chrome/extension/inject.js#L414)

For the jitter, I thought about it. Does "indexing..." not serve the dual meaning of "we are building code on our server" and "we are putting annotations on the page" for you? I'm personally ok with it as is, I feel like "checking status" would feel just as jittery, arguably more (since more characters change). Lmk if you feel strongly about this.

Adding popup, yup we can do that. 

Showing "errors while indexing" is the right thing to do. It's a bit tricker than I hoped, since we don't have a simple, granular way to identify the cause of the error. Maybe nice-to-have but not necessary?

Private marketing push, yeah, I'm absolutely game for this, but I'd prefer to do this as follow-on.

PR #61, opened 2016-06-09
Analytics: Google Analytics custom event tracking improvements

KattMingMing:
##### Description

Updates GA to track custom events based on category, label, action. This also adds more context to amplitude tracking events as well. GA events follow standard GA format and amplitude events have been modified to include additional  information. Other updates to GA include: user properties, 404 count and monitoring, and internal search query improvements.

##### Test plan

Confirmed GA and Amplitude are tracking as intended. For manual testing check staging3 and enable debug / logging for GA and Amplitude.


uforic:
Looked through code, and LGTM. Some top level questions:

- (optional) Regarding category naming, how sure are we in the names chosen, i.e. "auth", "repository", "def_info" , "references"; I mainly ask in case I want to build a feature, which names should I use? Maybe a comment with the general way to think about logging in the comments would be OK (i.e. "Decide what is the top level category this will effect. Is it a global category, like auth? Alternatively, is it a specific feature, like SFYE?". 

Other than that, seems low risk, and we should merge!

uforic:
Per discussion with @KattMingMing , LGTM once constants are pulled into constants file so folks can both understand structure and not worry about typos!

KattMingMing:
@uforic Will get this updated today

uforic:
LGTM!

PR #62, opened 2016-06-09
GlobalDefs: add filters for Languages, NotLanguages, Kinds, NotKinds

slimsag:
- No user facing change. This is a preface to a user-facing change that will be made by @attfarhan 
- Add an API to filter search results based on: `Languages` / `NotLanguages` (`Go`, `Java`, etc), and `Kinds` / `NotKinds` (`func`, `var`, etc).

In order to migrate to the new SQL scheme, users will need to run the following in a `psql` prompt:

```
-- add the new global_defs.language ID column
ALTER TABLE global_defs ADD COLUMN language smallint;
```

Additionally, only newly built/indexed repositories will be sortable by language. Repositories can be manually re-indexed by running `src search --refresh=github.com/foo/bar` for each repository (adding a reindexing command for _all repositories_ is left as a TODO, which I will attend to this week).

##### Reviewer tasks

- I'm confident the migration SQL produces the same table schema that we desire, that said, adding the new `language` column to the `global_defs` table will lock the table.. I don't have a measurement for how long. Can anyone confirm how long this will take (since we will be without DB for that time period)?

  - The new languages column should not have a serious perf impact, given it is a `int16` / `smallint` type.
  - Unit tests not worth the cost at this point.

##### Test plan

- Testing: on local machine:
  - Spin up a `master` revision SG instance.
  - Run the above migration code.
  - Spin up a `sg/global-defs-filters` SG instance.
  - Confirm that global search continues to work.
- Final testing+deployment: on staging, then in prod:
  - Pin the deployed version to `master`
  - Run the above migration code.
  - Deploy the new `master` (this branch).


slimsag:
The CI failure is related to https://github.com/go-gorp/gorp/issues/148 -- I'm looking into a workaround ...

api/sourcegraph/sourcegraph.proto
```diff
@@ -2054,6 +2054,22 @@ message SearchOptions {
 	// If empty, then no repositories are excluded.
 	repeated int32 NotRepos = 2 [(gogoproto.moretags) = "url:\",omitempty\""];
 
+	// Languages, if specified, limits the returned results to just the
+	// specified languages.
+	repeated string Languages = 6;
```
sqs: What is the set of allowable values? Is it "javascript", "JavaScript", or "Javascript"?

services/backend/internal/localstore/global_defs.go
```diff
@@ -247,6 +304,32 @@ func (g *globalDefs) Update(ctx context.Context, op store.GlobalDefUpdateOp) err
...
+			if _, ok := languages[l]; ok {
+				continue
+			}
+			languages[l] = true
+			languagesList = append(languagesList, l)
+
+			// We intentionally ignore the error here, because it is regularly
+			// "violates unique constraint" because the language is already in
+			// the table.
+			_ = graphDBH(ctx).Insert(&dbGlobalDefLanguages{Language: l})
```
sqs: See elsewhere where we check the err is indeed a primary constraint violation. Grep for `repo_uri_unique` in the code.

api/sourcegraph/sourcegraph.proto
slimsag: case insensitive. I've clarified this in 6230d1b814f7289f3d7e20f115fd8f196e1bf63e

services/backend/internal/localstore/global_defs.go
slimsag: Thanks! Fixed in e0b3438f210f2f46ce837451f92aa03fd7ac86d4

beyang:
Do we need a separate DB table to map language strings to numeric IDs? Would it not be simpler to just hardcode this, since that table will only ever be in the dozens of rows?

E.g.,

```
type Lang uint16

// Mappings from language to language ID. These should not be changed once set (doing so would require a database migration).
const (
  Go Lang = 1
  Java Lang = 2
  Python Lang = 3  
  ...
)
```

beyang:
Test on staging to see how long the migration (adding a column) will take. I think it should be fine, but best to double check.

slimsag:
> Do we need a separate DB table to map language strings to numeric IDs? Would it not be simpler to just hardcode this, since that table will only ever be in the dozens of rows?

It's true the table will only ever be a dozen of rows, but keeping it hard-coded isn't a huge simplification IMO and it also means that adding a new language to Sourcegraph requires modifying some hard-coded table, which seems silly. I think the table approach is better and not too complex.

> Test on staging to see how long the migration (adding a column) will take. I think it should be fine, but best to double check.

Yeah, I will certainly test on staging before prod.

beyang:
I agree, it's not a huge simplification in terms of readability, but I think it does offer a few substantial advantages:

1. Saves you a roundtrip to the DB to fetch the lang IDs.
2. Makes it easier to write code at the application level (e.g., `if def.Lang == Python` instead of if `pythonID := sql.Fetch(...); if def.Lang == pythonID`)
3. Eliminates the need for constraint checking on every insert to `global_defs`, which can add up if the table is updated often (which it is).

It also seems easier to add a bunch of hardcoded languages now to the code, than to create the table on every staging environment and add the initial set now (and potentially have to run a migration every time we add a new unanticipated language in the future).

slimsag:
> Saves you a roundtrip to the DB to fetch the lang IDs.

We can cache these with a simple in-memory map, it's trivial to do and would only result in a roundtrip to the DB once for each language.

> Makes it easier to write code at the application level (e.g., `if def.Lang == Python instead of if pythonID := sql.Fetch(...); if def.Lang == pythonID`)

IMO this point seems moot because we could just as easily write it as `if def.Lang == pythonID(ctx)` with a silly helper function.

> Eliminates the need for constraint checking on every insert to global_defs, which can add up if the table is updated often (which it is).

We can eliminate this using the in-memory map idea.

> It also seems easier to add a bunch of hardcoded languages now to the code, than to create the table on every staging environment and add the initial set now

We're already going to have to run migration (reindexing every repository), regardless of whether or not we use a hard-coded table because `global_defs` doesn't have the language ID column populated.

> (and potentially have to run a migration every time we add a new unanticipated language in the future).

I don't understand what you mean here, exactly, could you explain?

slimsag:
An example of what I mean with respect to the in-memory cache that solves most of your pain points can be seen in d932f6fb0c6f2e5797809d61fd714ddf1e8e5534

beyang:
If the table exists just to populate the in-memory cache, why have the underlying table at all? What value does it provide?

The `pythonID()` function would have to return an error, as well, due to the potential DB roundtrip.

Currently, we index Go, Java, Python. JavaScript, etc. Let's say in the future, we add language X. To add this language, we now have a DB migration to run, instead of just updating a hardcoded list in source code.

slimsag:
(me and Beyang discussed 1:1 in a VC)

I've added the hard-coded mapping for Go, Java, and Python. As we add more languages we'll need to update this map, but it'll be clear due to the non-null constraint on the column.

Once tested on staging I will merge and deploy to prod.

services/backend/internal/localstore/global_defs.go
```diff
@@ -20,11 +20,21 @@ import (
 	"sourcegraph.com/sqs/pbtypes"
 )
 
+// Mappings from language to language ID. These should not be changed once set
+// (doing so would require a database migration). When adding a new ID, you
+// must use the exact LOWERCASE srclib toolchain language name.
+var languages = map[string]uint16{
```
beyang: Instead of a map, why not a collection of consts?
```
const (
  Go = iota
  Java
  Python
  CSS
  JavaScript
  HTML
  YAML
  JSON
  Bash
  Ruby
  PHP
  Objective-C
  ... // etc., include a much more comprehensive list of languages, because why not?
)
```

beyang:
Actually, 1 more comment ^

services/backend/internal/localstore/global_defs.go
slimsag: We could do this, but we still need a map somewhere. The public gRPC API takes an array of strings, like "Go" which then need to be translated into an ID.

beyang: That works. This seems like a good place for a custom type: `type Lang uint16`

```
const (
  Go Lang = iota
  ...
)

var langToString = map[string]Lang{ ... }
var stringToLang = map[Lang]string { ... }

func (l Lang) String() string {  return langToString[l] }
func LangFromString(s string) Lang { return stringToLang[s] }
```

slimsag: :confused: With that approach how do we populate those maps? And who is going to use these `Go` etc constants? (this change doesn't need them)


The only choice I see with that approach is to write everything three times or use some generation tool. Is this what you are suggesting?

```
type Lang int16

const (
	Go Lang = iota
	Java
	Python
	CSS
	JavaScript
	HTML
	YAML
	JSON
	Bash
	Ruby
	PHP
	ObjectiveC
)

var stringToLang = map[string]Lang{
	"Go":         Go,
	"Java":       Java,
	"Python":     Python,
	"CSS":        CSS,
	"JavaScript": JavaScript,
	"HTML":       HTML,
	"YAML":       YAML,
	"JSON":       JSON,
	"Bash":       Bash,
	"Ruby":       Ruby,
	"PHP":        PHP,
	"ObjectiveC": ObjectiveC,
}

var langToString = map[Lang]string{
	Go:         "Go",
	Java:       "Java",
	Python:     "Python",
	CSS:        "CSS",
	JavaScript: "JavaScript",
	HTML:       "HTML",
	YAML:       "YAML",
	JSON:       "JSON",
	Bash:       "Bash",
	Ruby:       "Ruby",
	PHP:        "PHP",
	ObjectiveC: "ObjectiveC",
}

func (l Lang) String() string      { return langToString[l] }
func LangFromString(s string) Lang { return stringToLang[s] }
```

beyang: That's fair, maybe it's overkill. LGTM

PR #64, opened 2016-06-09
builds: Store user who created build

keegancsmith:
This is useful for audit purposes. Salvaged from the abandoned PR #52.


api/sourcegraph/sourcegraph.proto
```diff
@@ -813,6 +819,10 @@ message BuildsCreateOp {
 	// required.
 	string CommitID = 2;
 
+	// User is who to create the build for. This is the current user by
+	// default, and can only be set by internal authed scopes.
+	UserSpec User = 6;
```
sqs: I think we should just use the integer type, not UserSpec. I don't think the abstraction has been worth it, as we saw in the repo case.

api/sourcegraph/sourcegraph.proto
```diff
@@ -686,6 +689,9 @@ message BuildJob {
 	// CommitID is the full resolved commit ID to build.
 	string CommitID = 3;
 
+	// User is the user who to build on behalf of.
+	UserSpec User = 17;
```
sqs: (same comment re: using integer type)

api/sourcegraph/sourcegraph.proto
```diff
@@ -616,6 +616,9 @@ message Build {
 	// CommitID is the full resolved commit ID to build.
 	string CommitID = 3;
 
+	// User is who created the build, or who the build was created for.
+	UserSpec User = 16;
```
sqs: (same comment re: using integer type)

keegancsmith:
@sqs agreed, will update. It made this diff more complicated, so just using an int sounds great.

Also should we be using foreign key constraints at all?

sqs:
> Also should we be using foreign key constraints at all?

Ideally yes, but gorp makes it basically impossible due to its lack of `TRUNCATE TABLE ... CASCADE`. So, no, not right now.

keegancsmith:
@sqs PTAL

services/backend/mirror_repos.go
```diff
@@ -130,18 +136,19 @@ func (s *mirrorRepos) cloneRepo(ctx context.Context, repo *sourcegraph.Repo, rem
...
+		log15.Warn("cloneRepo: failed to create build", "err", err, "repo", repo.URI, "commit", res.CommitID, "asUID", asUID, "branch", repo.DefaultBranch)
 		return nil
 	}
-	log15.Debug("cloneRepo: build created", "repo", repo.URI, "branch", repo.DefaultBranch, "commit", res.CommitID)
+	log15.Debug("cloneRepo: build created", "repo", repo.URI, "branch", repo.DefaultBranch, "commit", res.CommitID, "asUID", asUID)
 	return nil
 }
 
-func (s *mirrorRepos) updateRepo(ctx context.Context, repo *sourcegraph.Repo, vcsRepo vcs.Repository, remoteOpts vcs.RemoteOpts) error {
+func (s *mirrorRepos) updateRepo(ctx context.Context, repo *sourcegraph.Repo, vcsRepo vcs.Repository, remoteOpts vcs.RemoteOpts, asUser sourcegraph.UserSpec) error {
```
sqs: Seems dangerous...it is easy(ier) to pass in the wrong UID here and give a build an elevated privilege. If you instead just pull the actor's UID from the context, you needn't worry about checking callers' security.

sqs:
1 comment, otherwise LGTM. Feel free to merge after you address that comment (if you disagree, I trust you, so just go ahead).

services/backend/mirror_repos.go
keegancsmith: that is the problem with this whole MirrorRepos API, it has an option called `AsUser`. I'm guessing what we really need is in the `repoupdater` code, it somehow can fake itself as a user when speaking to grpc, rather than passing along this variable. Any idea if that is possible (not familiar enough with our auth code)

That would also make this code better, since I can not use the same pattern (ie use a ctx instead of passing in a `UID` in the `BuildCreateOp`)

keegancsmith:
@sqs I agree, it is super fragile. Got any ideas on how we can remove the `AsUser` pattern?

sqs:
I think it's OK for now then.

Fundamentally I think that builds and such should *probably* run as a machine user using deploy keys, to avoid this problem. For example, what happens if the person who created the build leaves the org, or revokes their GH authorization? Then our builds would fail for everyone who uses that repo on Sourcegraph. That's a bigger change that we don't need to do yet.

keegancsmith:
@sqs I'm thinking to just abandon this. The risk of someone using the UID in the wrong way far outweighs any audit trail beneifts (which could just come from logging).

PR #69, opened 2016-06-10
Global & Local search unification (phase one)

slimsag:
This is the first change needed to unificate the global search and local search
code. In short, we use an almost identical pattern as we would with global
search except that we keep data in a separate table (for scalability in terms
of repo-level sharding in the future, as well as to not increase the global
search query time) and we store ALL commits in this table in contrast to just
the latest `master`/etc branch like we do for the `global_search` table.

A follow-up change will begin using this dataset for local search, once a
proper search re-indexing migration path is written. Users with an existing
DB will need to migrate to this change by running the following in a `psql`
prompt:

```
ALTER INDEX bow_idx RENAME TO global_defs_bow_idx;
ALTER INDEX doc_idx RENAME TO global_defs_doc_idx;

CREATE TABLE local_defs (
    repo text NOT NULL,
    commit_id text NOT NULL,
    unit_type text NOT NULL,
    unit text NOT NULL,
    path text NOT NULL,
    name text,
    kind text,
    file text,
    language smallint,
    ref_ct integer DEFAULT 0,
    updated_at timestamp with time zone,
    data bytea,
    bow text,
    doc text
);
ALTER TABLE ONLY local_defs
    ADD CONSTRAINT local_defs_pkey PRIMARY KEY (repo, commit_id, unit_type, unit, path);
CREATE INDEX local_defs_bow_idx ON local_defs USING gin (to_tsvector('english'::regconfig, bow));
CREATE INDEX local_defs_doc_idx ON local_defs USING gin (to_tsvector('english'::regconfig, doc));
CREATE INDEX local_defs_name ON local_defs USING btree (lower(name));
CREATE INDEX local_defs_repo ON local_defs USING btree (repo text_pattern_ops);
CREATE INDEX local_defs_updater ON local_defs USING btree (repo, unit_type, unit, path);
```

##### Test plan

1. Perform testing locally, including migration from current `master` to this branch using the above SQL.
2. Perform testing on staging, including migration from current `master` to this branch using the above SQL.
3. Merge and deploy to master using the migration SQL above.


services/backend/internal/localstore/global_defs.go
```diff
@@ -293,117 +307,142 @@ func (g *globalDefs) Update(ctx context.Context, op store.GlobalDefUpdateOp) err
...
 	for _, repoUnit := range repoUnits {
-		repoPath, commitID, err := resolveRevisionDefaultBranch(ctx, repoUnit.Repo)
-		if err != nil {
-			return err
-		}
-		defs, err := store.GraphFromContext(ctx).Defs(
-			sstore.ByRepoCommitIDs(sstore.Version{Repo: repoPath, CommitID: commitID}),
-			sstore.ByUnits(unit.ID2{Type: repoUnit.UnitType, Name: repoUnit.Unit}),
-		)
+		repoPath, commitIDs, defaultBranchCommitID, err := resolveRevisionAllBranches(ctx, repoUnit.Repo)
```
beyang: Shouldn't the list of commit IDs to update depend on the parameters passed to `Update`? I may be mistaken, but this seems like it takes the data passed `Update` and overwrites the previous data for all named branches with it.

beyang:
Rename `global_defs` to something else? Maybe `def_search`?

slimsag:
> Rename global_defs to something else? Maybe def_search?

Completely agree, I had a TODO about this as well but removed it. I'd like to handle this as part of a separate cleanup PR if possible, because we should also rename all "global defs" types, files, etc to match this new naming scheme. Is it okay to file this as an issue and we can resolve ASAP (but after this PR)?

beyang:
Sure. Add a comment in the source, too, for future readers.

services/backend/internal/localstore/global_defs.go
slimsag: Long-term, I agree, it should be. Check out the input parameter type `*store.GlobalDefSearchOp` which in practice looks like:

```
RepoUnit{
	Repo: 1, // github.com/slimsag/test-branchly
	Unit: "github.com/slimsag/test-branchly",
	UnitType: "GoPackage",
}
```

You can then read this code as:

- `Update()` the source `Unit: "github.com/slimsag/test-branchly"` please.
- For each branch in the repository
  - Find all defs in the given source `Unit`.
    - Upsert them into the `local_defs` table.
    - Only upsert them into the `global_defs` table if it's the default branch.

This is identical to how it worked before, except we'd only do it for the `DefaultBranch` / `master` because that's all `global_defs` cares about. In the future, we can optimize this a bit by only updating branches that have changed.

slimsag: I was also slightly afraid I didn't understand this code, so I traced much of it to conclude the above^; and I've also confirmed that it works as expected with a repo with two very different branches you can see here:

- https://staging3.sourcegraph.com/github.com/slimsag/test-branchly@master
- https://staging3.sourcegraph.com/github.com/slimsag/test-branchly@file2go

(notice the very different symbols, not overridden)

beyang: Does that mean we'll only have indexed code for the tip of every named branch?
Also, in a repository with a lot of named branches, this would lead to a huge amount of wasted work.

slimsag:
@beyang has merged this under some rebased commit.

PR #70, opened 2016-06-10
show most-ref'd defs on repo main page

sqs:
Previously, the repo main page would show an arbitrary 5 defs from the
repo. Now, for the empty rev (e.g.,
https://sourcegraph.com/github.com/USER/REPO with no `@rev`), it will
query global search to get the top 7 most-referenced defs. This
unfortunately includes internal refs and unexported defs. We can
hackily filter out the latter but can't rank based on global refs
only. Still, this gets close to "show me the most commonly used
exported API of this repo."

There are some hacks (noted with a TODO), but this is a part of the codebase that will change a lot soon, and I think the functionality is worthwhile.

#### Before (irrelevant results)
image

#### After (marginally more relevant results)
image


beyang:
Note: related issue: https://github.com/sourcegraph/sourcegraph-issues/issues/8

beyang:
What's the desired behavior if viewing a revision other than the default branch? If that's the case, querying global_defs might lead to 404 links.

sqs:
> What's the desired behavior if viewing a revision other than the default branch? If that's the case, querying global_defs might lead to 404 links.

The desired behavior is probably to not show this information at all, since users who are viewing non-default branches are probably power users (thus they don't need to see this to know Sourcegraph has "code intelligence") and/or repo authors (in which case they're less likely to always want to see the top-refd API funcs/types). But this PR just falls back to the previous behavior if any rev is specified in the URL, so it doesn't regress.

beyang:
LGTM

PR #71, opened 2016-06-10
app: make nav and body fixed width (except for blob)

sqs:
Prior to this PR, the top nav would always span the full screen width,
even if the main body container was fixed. This meant that the
DefInfo, Repo, and Tree pages (and static pages) looked a bit weird,
with the nav taking up full width and the content only taking up
partial width in the center.

@chexee Your call on whether to merge this. I was going to do some other demo changes where it'd make a bigger difference.

#### Before
screenshot from 2016-06-09 23-02-12

#### After
screenshot from 2016-06-09 23-01-59


chexee:
It's a bit awkward, but I'd rather not merge as the navigation width shouldn't jump around between pages. If the nav is all the way to the left on the Blob, but in the middle on Repo pages, it inhibits muscle memory – Navigation should be as consistent across the app as possible.

PR #72, opened 2016-06-10
Global & Local search unification (phase two)

slimsag:
This effectively causes local search to begin using the new `local_search` DB table added in https://github.com/sourcegraph/sourcegraph/pull/69

This change cannot be merged/deployed until a proper reindexing of all repos+branches has been performed.

##### Reviewer tasks

##### Test plan

I will test both locally and on staging before merge/deployment.


slimsag:
Please don't review until https://github.com/sourcegraph/sourcegraph/pull/69 is merged, since this change is based on that one.

slimsag:
Please only review the last two commits.

services/backend/defs.go
```diff
@@ -166,34 +166,65 @@ func (s *defs) List(ctx context.Context, opt *sourcegraph.DefListOptions) (*sour
...
-	var numEntries int
-	if len(defs0) < opt.Offset()+opt.Limit() {
-		numEntries = len(defs0) - opt.Offset()
-	} else {
-		numEntries = opt.Limit()
-	}
-	if numEntries < 0 {
-		numEntries = 0 // for last (or non-existent) pages
-	}
+	if opt.Query == "" {
```
beyang: In this case, to ensure stable ordering, can we just sort by number of refs?

services/backend/defs.go
```diff
@@ -166,34 +166,65 @@ func (s *defs) List(ctx context.Context, opt *sourcegraph.DefListOptions) (*sour
...
+		var repos []int32
+		for _, repoRev := range opt.RepoRevs {
+			repoPath, _ := sourcegraph.ParseRepoAndCommitID(repoRev)
+			res, err := svc.Repos(ctx).Resolve(ctx, &sourcegraph.RepoResolveOp{Path: repoPath})
+			if err != nil {
+				return nil, err
+			}
+			repos = append(repos, res.Repo)
+		}
+		results, err := svc.Search(ctx).Search(ctx, &sourcegraph.SearchOp{
```
beyang: It seems weird to have `defs.List` call `search.Search`. Instead, shouldn't it be the other way around? Search results will contain more than just def results.

services/backend/defs.go
slimsag: This case is _exactly_ the same behavior as it was before this change, i.e. only the `Query != ""` case has changed. I'd prefer not to change the behavior in this PR unless we have a valid reason to? Otherwise we might cause unrelated issues to crop up.

beyang: Is it up to date with https://github.com/sourcegraph/sourcegraph/pull/70?

services/backend/defs.go
slimsag: It is weird, I agree. :) There should be a:

```
// HACK: opt.Query does not belong in the Defs.List API, remove it and tell users to use `Search.Search` instead.
```

This is just a quick way of getting these results into the RepoTree view without changing frontend code -- which I would like to do as a cleanup process.

services/backend/defs.go
slimsag: No, but that PR has been reverted.

slimsag: Context in Slack around reversion of that: https://sourcegraph.slack.com/archives/dev-ops/p1465602123000545

services/backend/defs.go
beyang: Can you file a follow-up issue describing any cleanup that should be done soon?

services/backend/defs.go
beyang: Got it, makes sense.

services/backend/defs.go
beyang: File it on me in the sourcegraph-issues repo (using that to track Search)

slimsag: Yep, filed at https://github.com/sourcegraph/sourcegraph-issues/issues/22

slimsag:
@beyang has merged this under some rebased commit.

PR #75, opened 2016-06-10
lint: Follow all go vet advice

keegancsmith:
We exclude the vendor dir

  find . -type d -maxdepth 1 -mindepth 1 -not -name vendor | xargs go tool vet


Makefile
```diff
@@ -119,6 +119,7 @@ check: ${GOBIN}/go-template-lint
 	cd app && node ./node_modules/.bin/eslint --max-warnings=0 script web_modules
 	cd app && node ./node_modules/.bin/lintspaces -t -n -d tabs ./style/*.scss ./style/**/*.scss ./templates/*.html ./templates/**/*.html
 	go-template-lint -f app/tmpl_funcs.go -t app/internal/tmpl/tmpl.go -td app/templates
+	find . -type d -maxdepth 1 -mindepth 1 -not -name vendor | xargs go tool vet
```
sqs: It warns unless I put `-type d` after the depth args:

```
$ find . -type d -maxdepth 1 -mindepth 1 -not -name vendor
find: warning: you have specified the -maxdepth option after a non-option argument -type, but options are not positional (-maxdepth affects tests specified before it as well as those specified after it).  Please specify options before other arguments.

find: warning: you have specified the -mindepth option after a non-option argument -type, but options are not positional (-mindepth affects tests specified before it as well as those specified after it).  Please specify options before other arguments.

./.srclib-cache
./api
./services
./release
./test
./docs
./deploy
./.github
./client
./dev
./.git
./cmd
./app
./cli
./pkg
```

sqs:
1 comment, otherwise LGTM after CI passes

Makefile
keegancsmith: don't you love the differences between find on OS X and Linux :)

PR #77, opened 2016-06-10
lint: Follow some advice from opennota/check tools

keegancsmith:


Makefile
```diff
@@ -120,6 +120,7 @@ check: ${GOBIN}/go-template-lint
 	cd app && node ./node_modules/.bin/lintspaces -t -n -d tabs ./style/*.scss ./style/**/*.scss ./templates/*.html ./templates/**/*.html
 	go-template-lint -f app/tmpl_funcs.go -t app/internal/tmpl/tmpl.go -td app/templates
 	go list ./... | grep -v /vendor/ | xargs go vet
+	go list ./... | grep -v /vendor/ | xargs staticcheck
```
sqs: Add something to the Makefile to install this `staticcheck` tool (and possibly vendor it, too). I see it's installed in CI but not for local dev.

PR #79, opened 2016-06-10
builds: Ensure UI triggered/new repos have high priorities.

keegancsmith:
We can run coverage job builds with a much lower priority such that any other
build will run first.

Fixes https://app.asana.com/0/87040567695724/140739668608789

### DIFF
--- a/app/web_modules/sourcegraph/build/BuildBackend.js
+++ b/app/web_modules/sourcegraph/build/BuildBackend.js
@@ -67,9 +67,6 @@ const BuildBackend = {
 						body: JSON.stringify({
 							CommitID: action.commitID,
 							Branch: action.branch,
-							// HACK: this will enable builds queued through the ui
-							// to receive a higher priority.
-							Config: {Queue: true, Priority: 40},
 						}),
 					})
 						.then(checkStatus)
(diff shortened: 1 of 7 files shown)


keegancsmith:
Will require a rebuild of the src coverage docker image once merged.

rothfels:
@keegancsmith great, LGTM. One addition I may want to make (can be in a separate PR) is to give high priority to builds that are triggered by RefreshVCS, but only when it is called for a repo on which there are no builds. (E.g. new repos cloned via the chrome extension).

keegancsmith:
@rothfels where can I read the code for how that is done?

rothfels:
[Here's](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/client/browser-ext/chrome/extension/inject.js#L244) where the browser extension calls into the "repo create" endpoint (which afaict doesn't actually queue up a build). There's an interval timer running every 30s which calls RefreshVCS, which afaict does queue up a build.

keegancsmith:
Given that the first build for a repo is a rare event, I think in the backend we can just special case that and bump the priority there (similar to how we increase priority for private repos). That way we don't need to do anything special with each client of our API.

PR #82, opened 2016-06-10
coverage: remove calls to Repos.Get

rothfels:
This prevents the coverage script from spamming the GitHub API and DOSing our service.


cli/coverage_cmd.go
```diff
@@ -196,11 +196,10 @@ func (c *coverageCache) getResolvedRev(cl *sourcegraph.Client, ctx context.Conte
...
 	if repoRev.Rev == "" {
-		repo, err := cl.Repos.Get(ctx, &sourcegraph.RepoSpec{ID: res.Repo})
-		if err != nil {
-			return sourcegraph.RepoRevSpec{}, err
-		}
-		repoRev.Rev = repo.DefaultBranch
+		// Assume default branch is master to prevent call to Repos.Get.
+		// This may break for some repos (in which case we may want to hardcode mappings
+		// for exception cases).
+		repoRev.Rev = "master"
```
shurcooL: If you do want to automate this, there are a few ways I can suggest:

- Using github API directly (and provide it with a different auth). E.g., create a [new `go-github.Client`](https://godoc.org/github.com/google/go-github/github#Client) and use it here for coverage script needs only.

- Using git protocol. It requires having the `git` binary available, but you can use [`vcsstate.RemoteVCS`'s `RemoteBranchAndRevision` method](https://godoc.org/github.com/shurcooL/vcsstate#RemoteVCS) to get the remote branch (aka default branch of a repo).

shurcooL:
LGTM, this just deletes some code. You're a better judge if this is sufficient for the needs of coverage script and the repos it runs on, so feel free to merge if it works for your needs. I've posted some alternative ideas for how to avoid DOSing our service if that is still an issue.

rothfels:
@shurcooL awesome, that may be irrelevant but thanks for the pointers!

PR #83, opened 2016-06-11
cli: add src search --refresh-list, --refresh-list-rate, --refresh-list-start

slimsag:
With this command a plaintext file of repositories `repos.txt` can be provided as `src search --refresh-list=repos.txt` in order to refresh a large number of repositories.

In particular, this is inherently useful for refreshing all repositories by first exporting a list via something like `psql -U sg -P pager=off -A -t -c "select uri from repo;" > repos.txt` from the `app` database.

### DIFF
--- a/cli/search_cmd.go
+++ b/cli/search_cmd.go
@@ -1,8 +1,13 @@
 package cli
 
 import (
+	"bufio"
 	"log"
+	"os"
 	"strings"
+	"sync"
+
+	"github.com/rogpeppe/rog-go/parallel"
 
 	"sourcegraph.com/sourcegraph/sourcegraph/api/sourcegraph"
 	"sourcegraph.com/sourcegraph/sourcegraph/cli/cli"
@@ -20,6 +25,10 @@ func init() {
 }
 
 type searchCmd struct {
+	RefreshList      string `long:"refresh-list" description:"refresh every branch for all repositories in given file list"`
+	RefreshListRate  int    `long:"refresh-list-rate" description:"how many repositories to refresh in parallel" default:"8"`
+	RefreshListStart int    `long:"refresh-list-start" description:"start at the given index instead of repo 0 in the list"`
+
 	Refresh       string `long:"refresh" description:"repository URI for which to update the search index and counts (implies --refresh-counts)"`
 	RefreshCounts string `long:"refresh-counts" description:"repository URI for which to update the search counts"`
 	Limit         int32  `long:"limit" description:"limit # of search results" default:"10"`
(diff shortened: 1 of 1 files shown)


cli/search_cmd.go
```diff
@@ -1,8 +1,13 @@
 package cli
 
 import (
+	"bufio"
 	"log"
+	"os"
 	"strings"
+	"sync"
+
+	"code.google.com/p/rog-go/parallel"
```
sqs: use the github.com/rogpeppe/rog-go/parallel import path

cli/search_cmd.go
```diff
@@ -20,6 +25,10 @@ func init() {
 }
 
 type searchCmd struct {
+	RefreshList      string `long:"refresh-list" description:"refresh every branch for all repositories in given file list"`
```
sqs: can you just make it read from cmdline args? that's how the other commands work. it would be nice to be able to do `cat repos.txt | xargs src repo search --refresh`. also, with xargs, you don't need to handle parallelism in Go; you can use `xargs -P` to parallelize the operation.

cli/search_cmd.go
slimsag: yeah, already spotted and fixed :)

cli/search_cmd.go
slimsag: I mean.. we can already do this I guess with this (not your way, because yours will quit on error, like when a repo doesn't exist):

```
xargs -n 1 sh -c 'src search --refresh $0 || exit 0' < repos.txt
```

But now you need some more bash magic in order to skip the first N lines of the file in order to 'resume' a reindexing operation (like my `--refresh-list-start` flag provides).

Maybe it's just me, but this is far from simple to remember :) We don't have to commit this, though.

mkovacs: i've been using `src search --refresh` both manually and via shell scripts a lot lately, and this feature would've made my life much easier. many things can go wrong with these "for item in list do stuff" kind of DIY scripts, so i'd prefer the `-list` flag variants for the command. similar variants would be very helpful for `src repo sync` as well.

sqs: Seems like it could add a lot of extra code and increase the learning curve, if we want to do this for each command. If you want to continue after errors, you can run `... | xargs -n 1 ...`. xargs continues onto the next invocation if an invocation fails anyway, so @slimsag's `sh -c "... || exit 0"` is not necessary.

keegancsmith: I'm also a fan of just sticking to bash, but if this makes the lives easier of people triggering refreshes its probably worth the extra complexity in the command.

cli/search_cmd.go
```diff
@@ -30,7 +39,64 @@ type searchCmd struct {
...
+			run     = parallel.NewRun(c.RefreshListRate)
+
+			indexMu sync.Mutex
+			index   = -1
+		)
+		for scanner.Scan() {
+			indexMu.Lock()
+			index++
+			index := index
+			indexMu.Unlock()
```
keegancsmith: the lock on index is unnecessary, it is only ever read and written too from a single goroutine (the main one)

cli/search_cmd.go
```diff
@@ -30,7 +39,64 @@ type searchCmd struct {
...
+
+			if index < c.RefreshListStart {
+				continue
+			}
+			repoURI := scanner.Text()
+
+			run.Do(func() error {
```
keegancsmith: does `run.Do` guarantee things will start processing in the order `Do` is called?

Also if we allow providing a lower bound, shouldn't we also be providing an upperbound?

keegancsmith:
@slimsag any updates on this?

cli/search_cmd.go
slimsag: `run.Do` doesn't guarantee that AFAIK.

slimsag:
This change seems to be controversial, so I'm going to close.

If the need for this arises again, I'll consider adding a "Helpful commands" document for people to copy+paste the bash equivalent from.

PR #84, opened 2016-06-12
github: Cache unauthed not found responses

keegancsmith:
We 404 10x more than we 200 leading to our rate limit going to zero. The likely cause of this is people browsing private repositories from the Chrome Extension without them giving access to the private repo.

See commit messages.

### DIFF
--- a/services/ext/github/repos.go
+++ b/services/ext/github/repos.go
@@ -35,16 +35,31 @@ func init() {
 
 type Repos struct{}
 
+type cachedRemoteRepo struct {
+	sourcegraph.RemoteRepo
+
+	// PublicNotFound indicates that the GitHub API returned a 404 when
+	// using an Unauthed request (repo may be exist privately).
+	PublicNotFound bool
+}
+
 func (s *Repos) Get(ctx context.Context, repo string) (*sourcegraph.RemoteRepo, error) {
+	cl := client(ctx)
+
 	// This function is called a lot, especially on popular public
 	// repos. For public repos we have the same result for everyone, so it
 	// is cacheable. (Permissions can change, but we no longer store that.) But
 	// for the purpose of avoiding rate limits, we set all public repos to
 	// read-only permissions.
-	var cachedRemoteRepo sourcegraph.RemoteRepo
-	if err := reposGithubPublicCache.Get(repo, &cachedRemoteRepo); err == nil {
-		reposGithubPublicCacheCounter.WithLabelValues("hit").Inc()
-		return &cachedRemoteRepo, nil
+	var cached cachedRemoteRepo
+	if err := reposGithubPublicCache.Get(repo, &cached); err == nil {
+		if !(cl.isAuthedUser && cached.PublicNotFound) {
+			reposGithubPublicCacheCounter.WithLabelValues("hit").Inc()
+			if cached.PublicNotFound {
+				return nil, grpc.Errorf(codes.NotFound, "github repo not found: %s", repo)
+			}
+			return &cached.RemoteRepo, nil
+		}
 	} else if err != rcache.ErrNotFound {
 		log15.Error("github cache-get error", "err", err)
 	}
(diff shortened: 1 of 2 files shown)


services/ext/github/repos.go
```diff
@@ -79,6 +87,44 @@ func (s *Repos) GetByID(ctx context.Context, id int) (*sourcegraph.RemoteRepo, e
...
+	}
+	return &cached.RemoteRepo, nil
+}
+
+// getFromAPI attempts to get a response from the GitHub API without use of
+// the redis cache
+func getFromAPI(ctx context.Context, repo string) (*sourcegraph.RemoteRepo, error) {
+	owner, repoName, err := githubutil.SplitRepoURI(repo)
+	if err != nil {
+		return nil, nil
```
shurcooL: Is it intentional that this returns `nil` error when `err != nil`?

Is it because you consider the repo to not be found "successfully" if split repo URI fails? Because for all repos that can exist, split repo URI would not return error, right? Just confirming.

services/ext/github/repos.go
```diff
@@ -35,39 +35,47 @@ func init() {
...
 func (s *Repos) Get(ctx context.Context, repo string) (*sourcegraph.RemoteRepo, error) {
 	// This function is called a lot, especially on popular public
 	// repos. For public repos we have the same result for everyone, so it
 	// is cacheable. (Permissions can change, but we no longer store that.) But
 	// for the purpose of avoiding rate limits, we set all public repos to
 	// read-only permissions.
-	var cachedRemoteRepo sourcegraph.RemoteRepo
-	if err := reposGithubPublicCache.Get(repo, &cachedRemoteRepo); err == nil {
+	if cached, cachedErr := getFromCache(ctx, repo); cached != nil || cachedErr != nil {
 		reposGithubPublicCacheCounter.WithLabelValues("hit").Inc()
```
shurcooL: Does it make sense to do `WithLabelValues("hit").Inc()` for both when error is nil and not nil?

services/ext/github/repos.go
```diff
@@ -79,6 +87,44 @@ func (s *Repos) GetByID(ctx context.Context, id int) (*sourcegraph.RemoteRepo, e
...
+		if err != rcache.ErrNotFound {
+			log15.Error("github cache-get error", "repo", repo, "err", err)
+		}
+		return nil, nil
+	}
+
+	// Do not use a cached NotFound if we are an authed user, since it may
+	// exist as a private repo for the user
+	if client(ctx).isAuthedUser && cached.PublicNotFound {
+		return nil, nil
```
shurcooL: It seems pretty confusing how the return values are structured. If I understand correctly:

- repo, nil - means repo is found in cache, cache information should be used
- nil, error - means repo is found in cache as "non-existant", cache information should be used
- nil, nil - means repo is not found in cache (but might be available from API), so cache should be ignored

I think it's worth documenting the function about that behavior. It's pretty unintuitive. Usually non-`nil` error return means "do not use results from that call, because it failed".

shurcooL: But I'd recommend changing this to be more like normal Go code, where if return `error` is non-`nil`, then cache should not be used. If return error is nil, then use cache information (which can be that repo is returned or it's said to not exist).

So the usage code becomes:

```Go
if cacheResult, err := getFromCache(ctx, repo); err != nil {
    // use cacheResult, which might be that repo is found or not found
}
```

services/ext/github/repos.go
shurcooL: Okay, I see that the error returned from that func is not used in a usual way, so this is actually fine.

shurcooL:
Also, it's worth mentioning that, while this change can be very helpful to not use up rate limit when being hit with requests for the same repo that is 404, it also inherently has a disadvantage in that a 404 repo is cached for the TTL time, and if it's created/made available during that time, it'll keep incorrectly showing up as 404 until the cache expires.

It's a worthwhile trade-off for the current situation, and something we can revisit later.

shurcooL:
Ok, I've reviewed this for behavior and it looks good, with solid tests. LGTM.

I'll merge this for now in order to see what effect it has. The comments I left are all minor style issues and parts that can be documented/cleared up better, which we can follow up after. It's better to get this in sooner.

services/ext/github/repos_test.go
```diff
@@ -122,6 +122,106 @@ func TestRepos_Get_nonexistent(t *testing.T) {
...
+	}
+
+	// If we repeat the call, we should hit the cache
+	calledGetMissing = false
+	repo, err = s.Get(ctx, privateRepo)
+	if grpc.Code(err) != codes.NotFound {
+		t.Fatal(err)
+	}
+	if calledGetMissing {
+		t.Fatal("should of hit cache")
```
shurcooL: Minor spelling typo, s/of/have.

PR #85, opened 2016-06-13
[WIP] browser-ext: annotations for pull requests and file diffs

rothfels:

PR #86, opened 2016-06-13
SFYE release

uforic:
Set of changes intended for SFYE release. Includes:
- Link on homepage for SFYE
- Improved /tools page copy
- Different tools url scheme (/tools/editor, not /tools?type=editor
- Form to gather interest from people about future languages and editors
- MailChimp integration code to collect interest

##### Reviewer tasks

- [ ] HACKS: There are major Javascript hacks to get the location to set. I'd love to know the better way to do some of these, or be pointed to some of thelper functions! (i.e. to strip "editor" off the end of "/tools/editor"
- [ ] SECURITY: New endpoint introduced, pass through to mail chimp. Want to make sure it all looks good
- [ ] COPY: How is the copy lookking

##### Test plan

Have been testing on staging2. Sample forms submitted. 

** Note ** to be able to submit a form after you've submitted it once, open a new incognito browser, or type: window.localStorage["email_subscription_submitted"] = false 

##### Known issues
- [x] Install buttons on action screen look bad, battling with CSS and want to ask for some ideas on this


app/web_modules/sourcegraph/app/routePatterns.js
```diff
@@ -44,6 +45,7 @@ export const rel: {[key: RouteName]: string} = {
 	styleguide: "styleguide",
 	home: "",
 	tools: "tools",
+	tool_selected: "tools/*",
```
rothfels: We use camelCasing elsewhere (e.g. `myRepos`, and `defInfo`). 

Also, I'm assuming `tools` is still a valid top-level route on its own? Maybe instead of `tool_selected` a better name is just `tool`, since it's a drilldown for a specific one. Just nitpicking on name, doesn't really matter, though should be consistent. FYI this is going to automatically show up as a `View*` event in amplitude, all the more reason to do camelCasing and choose a name that is meaningful in an analytics context. Naming this tool would mean we have `ViewTools` and `ViewTool` events, which may be desirable.

app/web_modules/sourcegraph/components/Button.js
```diff
@@ -16,6 +16,8 @@ class Button extends Component {
 		loading: React.PropTypes.bool,
 		color: React.PropTypes.string, // "blue", "purple", "green", "red", "orange"
 		onClick: React.PropTypes.func,
+		type: React.PropTypes.string,
```
rothfels: Looks like this is never used, kill it.

app/web_modules/sourcegraph/dashboard/AnonymousLandingPage.js
```diff
@@ -78,6 +78,31 @@ class AnonymousLandingPage extends Component {
...
+								<p>See usage examples for code instantly, as you type. It's like pair programming with the smartest developer in the world.</p>
+								<div styleName="flex">
+									<div styleName="sfye-flex">
+										<a styleName="" href="/tools/editor">
+											<Button color="blue" onClick={(v) => v && eventLogger.logEvent("ClickedInstallEditorExt")}>
+												Install for your editor
+											</Button>
+										</a>
+									</div>
+									<div styleName="sfye_beta"><em>Beta for Go in Sublime Text and Vim, <a href="/tools/editor?expanded=true">get notified when Sourcgraph is available for your editor</a></em></div>
```
rothfels: is `_beta` still correct?

rothfels: Should this be a sentence, i.e. end with a period, start with "Current in beta..."?

app/web_modules/sourcegraph/home/Selector.js
```diff
@@ -0,0 +1,25 @@
+import React from "react";
```
rothfels: Should this be a generic component? (I.e. one that would live in our style guide?)

If so, it doesn't belong in `/home/` and shouldn't extend styling from `InterestForm`. But if it must be merged as-is, this is just debt that @chexee will need to pay later.

app/web_modules/sourcegraph/home/Selector.js
```diff
@@ -0,0 +1,25 @@
...
+import CSSModules from "react-css-modules";
+import styles from "./styles/InterestForm.css";
+
+class Selector extends React.Component {
+
+	static propTypes = {
+		mapping: React.PropTypes.object.isRequired,
```
rothfels: Does it have to be an object? Why not just an array of strings? It would reduce the amount of code / boilerplate in `InterestForm`.

app/web_modules/sourcegraph/home/Selector.js
```diff
@@ -0,0 +1,25 @@
...
+		requiredTitle: React.PropTypes.string.isRequired,
+	};
+
+	render() {
+		let optionalRender = [];
+		optionalRender.push(<option key="none" value="" disabled="true">{this.props.requiredTitle}</option>);
+		for (let key in this.props.mapping) {
+			optionalRender.push(<option key={key} name={key}>{this.props.mapping[key]}</option>);
```
rothfels: If you move to array of strings, `key` property can be anything unique, e.g. index.

app/web_modules/sourcegraph/home/Selector.js
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+import CSSModules from "react-css-modules";
+import styles from "./styles/InterestForm.css";
+
+class Selector extends React.Component {
+
+	static propTypes = {
+		mapping: React.PropTypes.object.isRequired,
+		requiredTitle: React.PropTypes.string.isRequired,
```
rothfels: How about just `title`? The requirement is implied at the end of the line, doesn't add much value to force the property name to include the constraint.

app/web_modules/sourcegraph/home/Selector.js
```diff
@@ -0,0 +1,25 @@
...
+	};
+
+	render() {
+		let optionalRender = [];
```
rothfels: I don't quite understand the name `optionalRender`. How about `options`?

app/web_modules/sourcegraph/home/InterestForm.js
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	render() {
+		return (
+			<form onSubmit={this._sendForm.bind(this)}>
+				{this.state.formError && <div>{this.state.formError}</div>}
+				<div styleName="container">
+					<div styleName="table-row">
+						<span styleName="full-input">
+							<input styleName="input-field" type="text" name="firstName" placeholder="Full name*" required={true}/>
```
rothfels: Use our standard `Input` component, I think.

app/web_modules/sourcegraph/home/InterestForm.js
```diff
@@ -0,0 +1,111 @@
...
+		super(props);
+		this.state = {
+			submitted: false,
+			formError: null,
+		};
+
+		this.languages = {
+			java: "Java",
+			python: "Python",
+			php: "PHP",
```
rothfels: Awfully repetitive, I don't think these need to be objects (see later comment).

app/web_modules/sourcegraph/home/InterestForm.js
```diff
@@ -0,0 +1,111 @@
...
+import styles from "./styles/InterestForm.css";
+import {Button} from "sourcegraph/components";
+import Selector from "./Selector";
+import Dispatcher from "sourcegraph/Dispatcher";
+import * as UserActions from "sourcegraph/user/UserActions";
+
+class InterestForm extends React.Component {
+
+	static propTypes = {
+		onSubmitted: React.PropTypes.func.isRequired,
```
rothfels: For consistency, usually named `onSubmit`

app/web_modules/sourcegraph/home/InterestForm.js
```diff
@@ -0,0 +1,111 @@
...
+		};
+	}
+
+	_sendForm(ev) {
+		ev.preventDefault();
+		const name = ev.currentTarget[0]["value"];
+		let firstName = null;
+		let lastName = null;
+		if (name) {
+			const names = name.match(/\S+/g); //	gets rid of leading whitespaces and splits on whitespace
```
rothfels: get rid of leading space before `gets`

rothfels: Also, I think this wouldn't require a comment and does the same thing:
```
name.trim().split(/\s+/)
```

app/web_modules/sourcegraph/home/InterestForm.js
```diff
@@ -0,0 +1,111 @@
...
+			if (names) {
+				firstName = names[0];
+				lastName = names.slice(1).join(" ");
+			}
+		}
+
+		this.props.onSubmitted();
+		Dispatcher.Backends.dispatch(new UserActions.SubmitEmailSubscription(
+			ev.currentTarget[1]["value"].trim(),
+			firstName.trim(),
```
rothfels: The trimming wouldn't be needed here if you use the suggestion above.

app/web_modules/sourcegraph/home/InterestForm.js
```diff
@@ -0,0 +1,111 @@
...
+			lastName.trim(),
+			ev.currentTarget[3]["value"].trim(),
+			ev.currentTarget[2]["value"].trim(),
```
rothfels: The indices here are super obscure, no idea what the ordering is of `currentTarget` other than to look in the `SubmitEmailSubscription` constructor. Is there a chance it could change?

app/web_modules/sourcegraph/home/InterestForm.js
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	_sendForm(ev) {
+		ev.preventDefault();
+		const name = ev.currentTarget[0]["value"];
+		let firstName = null;
+		let lastName = null;
+		if (name) {
+			const names = name.match(/\S+/g); //	gets rid of leading whitespaces and splits on whitespace
+			if (names) {
```
rothfels: Names should always be defined, given the above.

app/web_modules/sourcegraph/home/InterestForm.js
rothfels: Also, what are the asterisks for?

app/web_modules/sourcegraph/home/InterestForm.js
```diff
@@ -0,0 +1,111 @@
...
+							<Button styleName="button" type="submit" color="purple">Keep me updated!</Button>
+						</span>
+					</div>
+				</div>
+			</form>
+		);
+	}
+}
+
+export default CSSModules(InterestForm, styles, {allowMultiple: true});
```
rothfels: Are you still using the `allowMultiple` option in here? If not, remove.

app/web_modules/sourcegraph/home/ToolComponent.js
```diff
@@ -19,42 +22,114 @@ class ToolComponent extends React.Component {
...
+			java: "Java",
+			python: "Python",
+			php: "PHP",
+			csharp: "C#",
+			javascript: "Javascript",
+			cplusplus: "C++",
+			c: "C",
+			objectivec: "Objective C",
+			ruby: "Ruby",
+			swift: "Swift",
```
rothfels: Bad to have to duplicate these objects. Extract to a common location and import them where they're needed?

rothfels: (Too easy to forget to update one and not the other).

app/web_modules/sourcegraph/home/ToolComponent.js
```diff
@@ -19,42 +22,114 @@ class ToolComponent extends React.Component {
...
+			pycharm: "PyCharm",
+			rubymine: "RubyMine",
+			IntelliJ: "IntelliJ",
+		};
+	}
+
 	_dismissModal() {
-		this.context.eventLogger.logEvent("ToolBackButtonClicked", {toolType: this.props.location.query.tool});
-		this.context.router.replace({...this.props.location, query: ""});
+		this.context.eventLogger.logEvent("ToolBackButtonClicked", {toolType: this.props.location.pathname.split("/").slice(-1)[0]});
```
rothfels: Look at you fancy pants. :)

rothfels: TIL `.slice(-1)` 👍

app/web_modules/sourcegraph/home/ToolComponent.js
```diff
@@ -19,42 +22,114 @@ class ToolComponent extends React.Component {
...
+		this.context.router.replace({...this.props.location, pathname: this.props.location.pathname.split("/").slice(0, -1).join("/")});
+	}
+
+	_toggleView() {
+		this.context.eventLogger.logEvent("EmailSubscriptionViewToggled", {visible: !this.state.formExpanded});
```
rothfels: Shouldn't you remove the `!` in front of `this.state.formExpanded`?

app/web_modules/sourcegraph/home/ToolComponent.js
```diff
@@ -19,42 +22,114 @@ class ToolComponent extends React.Component {
...
+		this.context.router.replace({...this.props.location, query: {...this.props.location.query, expanded: !this.state.formExpanded}});
```
rothfels: And here, too. Feels weird, because the names `visible` and `expanded` do not imply the need to negate `formExpanded`. It would make sense to keep the negation if the names were `hidden` and `collapsed`.

app/web_modules/sourcegraph/home/ToolComponent.js
```diff
@@ -19,42 +22,114 @@ class ToolComponent extends React.Component {
...
+		this.setState({formExpanded: !this.state.formExpanded});
+	}
+
+	_getVisibility() {
```
rothfels: Is this method any better than using `this.state.formExpanded` wherever you call the function?

app/web_modules/sourcegraph/home/ToolComponent.js
```diff
@@ -19,42 +22,114 @@ class ToolComponent extends React.Component {
...
+
+	_optionalFormView() {
+		if (!this.props.supportedTool.interestForm) {
+			return <div/>;
+		}
+
+		if (!this.state.submitted) {
+			return (<div>
+				<div styleName="dont-see-div">
+					<a styleName="dont-see-link" onClick={() => this._toggleView()}>{this.state.formExpanded ? <TriangleDownIcon /> : <TriangleRightIcon />}{this.props.supportedTool.interestForm.title}</a>
```
rothfels: trick: you can replace `() => this._toggleView()` with `this._toggleView`, because you don't pass it any special args.

app/web_modules/sourcegraph/home/ToolComponent.js
```diff
@@ -19,42 +22,114 @@ class ToolComponent extends React.Component {
...
+							<CloseIcon className={base.pt2} />
+						</Button>
+					</div>
+					<div styleName="flex-container">
+						<span>{this.props.supportedTool.hero.img ? <img styleName="tool-img" src={`${this.context.siteConfig.assetsRoot}${this.props.supportedTool.hero.img}`}></img> : <div styleName="tool-img"></div>}</span>
+						<div>
+							<Heading align="left" level="2" className={base.pt5}>{this.props.supportedTool.hero.title}</Heading>
+							<div styleName="tool-item-paragraph">
+								<b>{this.props.supportedTool.hero.subtitle}</b>
+								<br/><br/>
```
rothfels: Generally avoid line breaks like this for presentation, and put presentation in CSS. Not a biggie, fine to leave here, just general advice for the future.

app/web_modules/sourcegraph/home/ToolComponent.js
```diff
@@ -5,12 +5,15 @@ import base from "sourcegraph/components/styles/_base.css";
...
 import {CloseIcon} from "sourcegraph/components/Icons";
 import Modal from "sourcegraph/components/Modal";
+import {TriangleRightIcon, TriangleDownIcon, CheckIcon} from "sourcegraph/components/Icons";
+import InterestForm from "./InterestForm";
 
 class ToolComponent extends React.Component {
 
 	static propTypes = {
 		location: React.PropTypes.object.isRequired,
 		supportedTool: React.PropTypes.object.isRequired,
```
rothfels: General advice for the future (especially important/helpful in JS):

"what properties are on `supportedTool`" will be a common question teammates / others looking at this will ask. For things like `location`, which is used everywhere and has canonical meaning, it's less relevant, but instead of using `React.PropTypes.object` you can do the following:

```
React.PropTypes.shape({
    mySpecialProperty: React.PropTypes.string.isRequired,
    ...
}).isRequired
```

Here's an example of when that becomes very obviously useful: https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/app/web_modules/sourcegraph/admin/CoverageDrilldown.js

(imagine if the toplevel property was just "object")

app/web_modules/sourcegraph/home/ToolsHomeComponent.js
```diff
@@ -26,11 +26,22 @@ class ToolsHomeComponent extends Component {
 		this.state = {
 			showChromeExtensionCTA: !window.localStorage["installed_chrome_extension"],
 		};
+	}
+
+	componentDidMount() {
+		this.timeout = setTimeout(() => this.setState({
```
rothfels: Do you have to keep a reference to the timeout? Or can you just fire and forget?

app/web_modules/sourcegraph/home/ToolsHomeComponent.js
```diff
@@ -26,11 +26,22 @@ class ToolsHomeComponent extends Component {
...
+			showChromeExtensionCTA: !document.getElementById("sourcegraph-app-bootstrap"),
```
rothfels: Note to self, I've changed the name of this in: https://github.com/sourcegraph/sourcegraph/pull/85

app/web_modules/sourcegraph/home/ToolsHomeComponent.js
```diff
@@ -26,11 +26,22 @@ class ToolsHomeComponent extends Component {
...
+		}), 1);
+	}
 
+	reconcileState(state, props, context) {
+		// TODO(matt): Context wasn't avaiable in the constructor, and we need context to get the assetUrl.
```
rothfels: I don't see you using Context in here. Am I missing something?

app/web_modules/sourcegraph/home/ToolsHomeComponent.js
```diff
@@ -39,33 +50,27 @@ class ToolsHomeComponent extends Component {
...
-
-		this.optionalSelectedToolComponent = this.supportedTools[state.location.query.tool] ? <ToolComponent supportedTool={this.supportedTools[state.location.query.tool]} location={state.location}/> : null;
+		const toolName = state.location.pathname.split("/").slice(-1)[0];
+		this.optionalSelectedToolComponent = this.supportedTools[toolName] ? <ToolComponent supportedTool={this.supportedTools[toolName]} location={state.location}/> : null;
 	}
 
 	_toolClicked(toolType) {
 		this.context.eventLogger.logEventForPage("ToolCTAClicked", "DashboardTools", {toolType: toolType});
-		this.context.router.replace({...this.props.location, query: {tool: toolType}});
+		this.context.router.replace({pathname: "/tools/".concat(toolType)});
```
rothfels: Use string interpolation (the backtic syntax) instead of `.concat`

app/web_modules/sourcegraph/home/index.js
```diff
@@ -11,6 +11,16 @@ export const tools = {
 	},
 };
 
+export const tool_selected = {
```
rothfels: Note to future Matt: this could be structured better as a child component/route of the tools route. `react-router` allows you to nest routes/components, which is essentially what's happening here. Not worth changing in this diff, just remember in the future.

app/web_modules/sourcegraph/home/index.js
```diff
@@ -37,6 +47,10 @@ export const routes: Array<Route> = [
 		path: rel.tools,
 	},
 	{
+		...tool_selected,
```
rothfels: Reminder: if you change the name in `routePatterns`, change the name here.

app/web_modules/sourcegraph/user/UserActions.js
```diff
@@ -167,3 +167,31 @@ export class ResetPasswordCompleted {
 		this.eventName = "ResetPasswordCompleted";
 	}
 }
+
+export class SubmitEmailSubscription {
```
rothfels: If you don't give this an event name, it won't automatically be logged.

services/httpapi/router/router.go
```diff
@@ -77,6 +78,8 @@ func New(base *mux.Router) *mux.Router {
 	base.Path("/forgot").Methods("POST").Name(ForgotPassword)
 	base.Path("/reset").Methods("POST").Name(ResetPassword)
 
+	base.Path("/email-subscription").Methods("POST", "GET").Name(EmailSubscription)
```
rothfels: POST only, not GET

services/httpapi/user.go
```diff
@@ -38,3 +38,24 @@ func serveUserEmails(w http.ResponseWriter, r *http.Request) error {
...
+func serveEmailSubscription(w http.ResponseWriter, r *http.Request) error {
+	newBody, newErr := ioutil.ReadAll(r.Body)
+	if newErr != nil {
+		return newErr
+	}
+
+	client := &http.Client{}
+	req, err := http.NewRequest("POST", "https://us8.api.mailchimp.com/3.0/lists/dd6c4706a1/members", bytes.NewReader(newBody))
+	req.Header.Set("Content-Type", "application/json")
+	req.SetBasicAuth("user", "132948e46a1e7ce8bad6fe375567383e-us8")
```
rothfels: This is a secret, right? We shouldn't check it into the codebase; it should be read from the environment. (I can show you how to do this).

Do you want this to work on test/dev/staging tiers? Is it ok if the endpoint occasionally gets spammed by developers?

services/httpapi/user.go
```diff
@@ -38,3 +38,24 @@ func serveUserEmails(w http.ResponseWriter, r *http.Request) error {
...
+	resp, err := client.Do(req)
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+
+	return writeJSON(w, string(body))
```
rothfels: Have you tested to see if this works? I thought the second arg needed to be a pointer to a struct, but I could be wrong.

services/worker/plan/srclib_images.go
```diff
@@ -9,7 +9,7 @@ import (
 // developing to make it easier to test out changes to a given toolchain. E.g.,
 // `droneSrclibGoImage = "sourcegraph/srclib-go"`.
 var (
-	droneSrclibBashImage       = "sourcegraph/srclib-bash@sha256:08ed88e0d833ea3f987e685b797867400325de566bc63d4f018e0a7a0ed15c2f"
+	droneSrclibBashImage       = "sourcegraph/srclib-bash@sha256:1b03e7f7b30122238674bbc660e07fdaa09c8840428b8b64595f002f691595a1"
```
rothfels: How did this get in here?

app/web_modules/sourcegraph/home/Selector.js
uforic: @uforic come back later, and do this

app/web_modules/sourcegraph/home/InterestForm.js
uforic: Hey! Silly question, the reason I didn't use the standard "<Input>" element is that I couldn't figure out how to attach styles to it, i.e. I want it to be a certain width. styleName= didn't seem to carry through. Thoughts?

app/web_modules/sourcegraph/home/ToolComponent.js
uforic: Cool - any suggestions on where to put them? Make a class called "StaticContstants" or something, and which folder should it live in? Just wondering what we've done in the past.

app/web_modules/sourcegraph/home/ToolComponent.js
uforic: Haha if you asked me to explain it to you now... I should just link to the stack overflow post :-)

app/web_modules/sourcegraph/home/ToolComponent.js
uforic: Was just thinking about this:

case this.state.formExpanded = false -> We are about to open it, so flipping visilbe = true makes sense. Maybe I'm missing what you were going for - we basically haven't flipped the bit yet. I reordered it, and didn't flip the bit, that  looks better. Toughts?

app/web_modules/sourcegraph/home/ToolComponent.js
uforic: Weird - I tried this, and it didn't work. Thoughts?

app/web_modules/sourcegraph/home/ToolComponent.js
uforic: Got it  so shape is kind of a catch all, you can overload it to have whatever properties you'd like?

Question - the object incoming was also defined, does this result in a double de?

app/web_modules/sourcegraph/home/ToolsHomeComponent.js
uforic: @KattMingMing  - mind jumping in here? Would love to understand this, not sure it makes sense to me,.

app/web_modules/sourcegraph/home/ToolsHomeComponent.js
uforic: Ah, this._sfyeCTA() does (off this this object).

app/web_modules/sourcegraph/home/ToolsHomeComponent.js
uforic: TIL. Nice!

app/web_modules/sourcegraph/home/index.js
uforic: Cool - is there a file I could look at that does this? Also, figure I can ask here - what is the "..." syntax do?

app/web_modules/sourcegraph/user/UserActions.js
uforic: @KattMingMing mind taking a look at this? I checked to see if we do it elsewhere, I didn't see it - so probably worth filling in?

services/httpapi/router/router.go
uforic: Security vuln elsewise? Or just stick to the known quantity?

services/httpapi/user.go
uforic: @KattMingMing - let's figure out what we want to do here. This is kinda sorta secret, I think. I've seen us stick some variables in the environmental variables for the server, maybe we'll do this. I think we are OK with developers doing that, but I've seen we can have a developer and prod key, and just have two different lists. Probably better to do it that way

services/worker/plan/srclib_images.go
uforic: Note to self - fix this tomorrow.

uforic:
Phewph. A couple outstanding items, thanks again @rothfels . @KattMingMing , let's knock the last couple things out tomorrow.

I haven't pushed my changes yet cause i hate when GH minimizes all of them, but anything with a thumbs up is already done.

services/httpapi/user.go
KattMingMing: We should not have this getting spammed by developers since we can only have a certain number of subscriptions. We could swap out the list_id based on dev and hit the sandbox env

services/httpapi/user.go
KattMingMing: I tested this and it worked, I checked the response tab and the JSON was getting back correctly and there weren't any errors, but maybe I missed something?

app/web_modules/sourcegraph/home/index.js
rothfels: repo/routes.js

app/web_modules/sourcegraph/home/Selector.js
uforic: Two things: a) having trouble doing this, after I move the file, the Javascript breaks cryptically, even after changing the various constants and import paths.

b) there is already a dropdown, but it doesn't look very purty. Thinking we can convert all this to a better UI at a separate point. Good point though!

app/web_modules/sourcegraph/home/InterestForm.js
uforic: Roger! Will add some style ideas to the Design Asana.

app/web_modules/sourcegraph/home/ToolComponent.js
uforic: Req'd this.togleView.bind(this)

app/web_modules/sourcegraph/home/ToolComponent.js
uforic: Not doing this for now - the tools page is going to change a lot, and it turns out each supported tool has a bunch of optional arguments, will be worth it once we standardize a bit more!

app/web_modules/sourcegraph/home/ToolsHomeComponent.js
uforic: I took out the variable storage!

PR #87, opened 2016-06-13
services/ext/github: Style changes to improve readability.

shurcooL:
Followup to #84 comments left by me. This should be a refactor that does not change behavior. These are suggestions, so @keegancsmith feel free to accept/reject them depending on if you agree.

Make `getFromCache` and `getFromAPI` signatures more traditional. They now both return (*Resource, error) where Resource can/should only be used if and only if error is nil. This behavior is also now documented.

---

Also, perform `githubutil.SplitRepoURI` first thing in github.Repos.Get, before even hitting the cache.

The cache is redis and may be a network call, etc., so the assumption is that `githubutil.SplitRepoURI` is a much cheaper/quicker operation. If the local parsing of `githubutil.SplitRepoURI` fails, there's no way it's a valid repo that would be cached.

This is a minor optimization that can help in a situation where an invalid repo URL is requested at a high rate.

##### Reviewer tasks

- [ ] STYLE: reviewer accepts style changes introduced by this change (and considers them an improvement)

##### Test plan

Tests pass locally.

### DIFF
--- a/services/ext/github/repos.go
+++ b/services/ext/github/repos.go
@@ -1,6 +1,7 @@
 package github
 
 import (
+	"errors"
 	"fmt"
 
 	"google.golang.org/grpc"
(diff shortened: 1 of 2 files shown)


keegancsmith:
LGTM

keegancsmith:
Thanks, this is much better.

PR #88, opened 2016-06-13
services/ext/github: Test that 404 responses get cached even if user is authed.

shurcooL:
This optimization was introduced in 7f3478a8342c2d2064f631cb0e7171be15d49715.
Caching 404 responses from authed clients is safe to do, since we're
only caching not found responses and not found repositories. This does
not reveal sensitive information.

Add test to enforce it so it doesn't regress.

Try to clear up some existing tests, explain a potentially puzzling
part of a test.

### DIFF
--- a/services/ext/github/repos_test.go
+++ b/services/ext/github/repos_test.go
@@ -206,12 +206,12 @@ func TestRepos_Get_publicnotfound(t *testing.T) {
 		t.Fatal("should have hit cache")
 	}
 
-	// authed user + 404 == never cache. Do twice to ensure we do not
-	// cache
+	// Authed user should never use public cache. Do twice to ensure we do not
+	// use the cached 404 response.
 	for i := 0; i < 2; i++ {
 		calledGetMissing = false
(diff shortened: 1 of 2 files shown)


services/ext/github/repos_test.go
```diff
@@ -222,6 +222,55 @@ func TestRepos_Get_publicnotfound(t *testing.T) {
...
+	}
+
+	mock := &minimalClient{}
+	ctx := testContext(mock)
+
+	s := &Repos{}
+	missingRepo := "github.com/owner/doesnotexist"
+	mock.repos = mockGetMissing
+
+	// If we get as an authed user, we will not store the 404 repo in cache.
```
keegancsmith: this comment is a little unclear. We want to store the 404, but we don't want to retrieve from the cache. I'll update this before merging.

shurcooL: Good catch. I forgot to update the comment from the time when this WIP test was testing the opposite behavior.

PR #90, opened 2016-06-13
builds: Include Go as supported language

keegancsmith:
Example output from `Detect projects` now

```
Detected 1 programming languages in use by this repository:
 - Go

Found 2 non-programming languages in use by this repository that will not be analysed:
 - Markdown
 - YAML
```

Before it would of been

```
Detected 3 programming languages in use by this repository:
 - Go
 - Markdown
 - YAML
```

but then it would of warned about not being able to build all 3 languages (but
would still analyse Go).

Fixes https://app.asana.com/0/87040567695724/142418359113253

### DIFF
--- a/services/worker/builder/builder.go
+++ b/services/worker/builder/builder.go
@@ -197,6 +197,16 @@ func (b *Builder) plan(ctx context.Context) (finalConfig string, axes []matrix.A
 
 				if err == nil {
 					w := state.Log()
+					langs := make([]*inventory.Lang, 0, len(inv.Languages))
+					skipped := make([]*inventory.Lang, 0, len(inv.Languages))
+					for _, l := range inv.Languages {
+						if _, ok := skipLangs[l.Name]; ok {
+							skipped = append(skipped, l)
+						} else {
+							langs = append(langs, l)
+						}
+					}
+					inv.Languages = langs
 					if len(inv.Languages) == 0 {
 						if !b.DroneYMLFileExists {
 							fmt.Fprintln(w, "No recognized programming languages were detected in this repository.")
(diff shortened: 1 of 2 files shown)


services/worker/builder/builder.go
```diff
@@ -210,6 +221,12 @@ func (b *Builder) plan(ctx context.Context) (finalConfig string, axes []matrix.A
 							fmt.Fprintf(w, " - %s\n", lang.Name)
 						}
 					}
+					if len(other) > 0 {
+						fmt.Fprintf(w, "\nFound %d non-programming languages in use by this repository that will not be analysed:\n", len(other))
```
alexsaveliev: Doesn't `CSS` fall into `non-programming` category?

keegancsmith: good point. I see this would exclude us analysing it. We also have other support like HTML and JSON. We have a whitelist of languages to exclude that is no longer in use. I'll use that instead.

keegancsmith:
@alexsaveliev updated so that we only exclude non-programming languages in the build steps. We will still analyse non programming languages if they are supported.

services/worker/builder/builder.go
```diff
@@ -410,12 +426,20 @@ func (noopMonitor) End(ok, allowFailure bool) {}
 
 func (noopMonitor) Logger() (stdout, stderr io.Writer) { return ioutil.Discard, ioutil.Discard }
 
+// skipLangs are languages we shouldn't build or analyse. They are usually
+// langauges that are common in repos, but including them in warnings would be
```
alexsaveliev: typo :)

services/worker/plan/infer.go
```diff
@@ -22,7 +22,10 @@ func inferConfig(inv *inventory.Inventory) (*droneyaml.Config, []matrix.Axis, er
 	for _, lang := range inv.Languages {
 		c, ok := langConfigs[lang.Name]
 		if !ok {
-			unsupported = append(unsupported, lang.Name)
+			// Only warn about langauges which are possible to build.
```
alexsaveliev: typo :)

alexsaveliev:
LGTM

mkovacs:
We're indexing Shell scripts now. Please make sure that's not in `skipLangs` if you need to modify it in the future.

PR #91, opened 2016-06-13
Fix text-ellipses and text alignment at small sizes.

chexee:
The ellipses truncation broke at some point (possibly after adding the RevSwitcher).

This adds back ellipses truncation and fixes text and RevSwitcher menu misalignment on mobile.


uforic:
Sick. I'm going to run this locally, and double check it looks good, then let's get this merged in. Thanks!

uforic:
LGTM, just verified it is fixed on my machine.

Also double checked the blob and repos pages

Ready to merge!

PR #92, opened 2016-06-13
make firefox build process start from scratch, not use files on the mac

uforic:
I realized the old process used the node_modules folder lying around on the mac, and potentially the build and dev folder.

The new one will wipe these, and start from scratch.

Takes longer, but less likely to run into a problem. A git diff across node_modules from Linux and macs showed a difference, so this is the safe approach.

Test - starting with a build/ and dev/ folder and node_modules, make sure it works and all folders are wiped. Verified that it worked on my box, would appreciate a plus 1!


rothfels:
LGTM

sqs:
A safer way to do this would be to check in the node_modules dir, as we do for `app/`. That way, we always use the same versions and aren't reliant on the external servers being up or fast.

keegancsmith:
@uforic any updates on this?

uforic:
hey @keegancsmith - forgot about this puppy. Merging in.

@sqs - agreed - would be better to check in node_modules. I'll look into this, what I was seeing is that node_modules were different across linux and mac, causing builds to fail in certain envs. Will make sure this isn't the case first.

PR #97, opened 2016-06-14
client/browser-ext: don't try to render or refresh repo if on sourcegraph.com

sqs:
Previously, the Chrome extension would try to render the page, even on 
sourcegraph.com. It would parse the URL path, which would lead to incorrect
behavior (such as trying to refresh the repo
"github.com/github.com/USER"), because the URL paths begin with
"github.com" on sourcegraph.com.

The Chrome extension is intended for use on GitHub.com only (except for
reading the access token from Sourcegraph.com), anyway.


sqs:
@attfarhan Was there any reason to call refreshState if the user is on a sourcegraph.com page? Just checking. If not, then this PR can be safely merged.

attfarhan:
@sqs No, there wasn't any reason for that, this should be good to go!

PR #99, opened 2016-06-14
pkg/vcs: handle 'unusual' chars in filenames in git repos

sqs:
Previously, on some systems (with git config core.quotePath set to on, or
unset), 'unusual' filenames (such as those containing bytes higher than 0x80)
would be returned quoted. This could cause filenames in the UI to display with
quotes surrounding them and literal backslashes with character escape codes.

Adding -z suppresses this quoting.


shurcooL:
LGTM.

Using `-z` is a good idea in context of parsing output with tools that can handle invisible bytes (better than humans can on a screen).

We did use this in the past, I wonder why it was ever removed. E.g., see https://github.com/sourcegraph/go-vcs/blob/5fa037434186ecbbbceff6778db7a1f8353e7007/vcs/gitcmd/repo.go#L1163.

PR #102, opened 2016-06-14
Global search analytics

attfarhan:
This change improves analytics for global search.  It adds several properties to the existing events `GlobalSearchInitiated` and `GlobalSearchItemSelected` to better understand whether search is providing our users with the results they want.   

For `GlobalSearchInitiated` we add the following properties: `kinds`, `languages`, `pageName`, `totalResults`.  

For `GlobalSearchItem` we now have the following properties: `kindSelected`, `languageSelected`, `repoSelected`, `pageName`, `totalResults`, `indexSelected`.

##### Reviewer tasks


app/web_modules/sourcegraph/search/GlobalSearch.js
```diff
@@ -99,13 +101,29 @@ class GlobalSearch extends Container {
 		}
 	}
 
+	__onDispatch(action) {
+		if (action.constructor === SearchActions.ResultsFetched) {
+			let globalSearchEventDict = {};
+			globalSearchEventDict["globalSearchQuery"] = this.state.query;
+			globalSearchEventDict["page name"] = this.props.location.pathname === "/search" ? "Global search homepage" : "Global search repo page";
```
beyang: This seems brittle (prone to break if the routes ever change). Can we use the router to check if it matches up with the global search URL?

attfarhan: thanks, just changed this to check if it matches `rel.search` from routePatterns.js.

app/web_modules/sourcegraph/search/GlobalSearch.js
```diff
@@ -99,13 +105,28 @@ class GlobalSearch extends Container {
 		}
 	}
 
+	__onDispatch(action) {
```
KattMingMing: One option might be to not actually subscribe for the registration inside GlobalSearch and just do all of this inside of the EventLogger. I'm in the process of making some changes to the way we are tracking events, but right now the event logger is notified when all dispatch events that emit a change take place. We could add this to the switch statement in there to reduce the event logging noise inside of `GlobalSearch` and just pass into the constructor all of the analytics data that needs to go through. This isn't absolutely necessary to change, but just a thought.

KattMingMing: All events that dispatch a change are already getting forwarded to the EventLogger so it would just be making the EventLogger aware of SearchActions.

app/web_modules/sourcegraph/search/GlobalSearch.js
```diff
@@ -215,7 +236,7 @@ class GlobalSearch extends Container {
 
 		const def = this.state.matchingResults.Defs[i - offset];
 		const url = urlToDefInfo(def) ? urlToDefInfo(def) : urlToDef(def);
-		this.context.eventLogger.logEvent("GlobalSearchItemSelected", {globalSearchQuery: this.state.query, selectedItem: url});
+		this.context.eventLogger.logEvent("GlobalSearchItemSelected", {globalSearchQuery: this.state.query, selectedItem: url, languageSelected: this.state.matchingResults.Defs[i].FmtStrings.Language, kindSelected: this.state.matchingResults.Defs[i].FmtStrings.Kind, repoSelected: this.state.matchingResults.Defs[i].Repo, indexSelected: i, totalResults: this.state.matchingResults.Defs.length});
```
KattMingMing: Will there be any problems if any of the objects you're trying to access inside of `this.state.matchingResults.Defs[i].FmtStrings.Kind` are null?

beyang:
+1 from me, assuming King has no further comments.

app/web_modules/sourcegraph/search/GlobalSearch.js
attfarhan: Added a check to make sure any of the Defs[i].FmtString properties are not undefined, which will solve this.

KattMingMing:
+1

attfarhan:
Merged

PR #104, opened 2016-06-14
unify search UI: remove local def search, add 's' hotkey for global search, repo homepage w/frequently used defs

sqs:
The directory tree listing for repos no longer searches or lists defs.
Instead, you can open the global search modal anywhere by hitting the
`s` hotkey. This removes some functionality (local def searching, and
seeing defs on the repo homepage), but it will be added back shortly. 

Greatly improves the performance of the file finder and
branch/tag menu filtering (instead of using Fuze.js, now uses a
simpler fuzzysearch lib).

Adds a "Frequently used" list of defs to the repo homepage. After removing the defs from the TreeSearch, it's important to show some defs on the repo homepage so that users know there is more than just files.

Test plan:

https://staging4.sourcegraph.com

1. Go to a never-seen-before repo and ensure it clones properly and defs show up without refreshing the browser.
2. Use global search (hotkey: `s`) from many different pages.
3. Use TreeSearch for jump-to-file from many different pages.
4. Use the RevSwitcher.

NOTE: The repo homepage and the global search modal are not polished. These are not announced features/screens, so this is OK.

#### bare repo TreeSearch (w/no defs)
screenshot from 2016-06-14 02-34-59

#### global search modal (accessible using `s` hotkey from anywhere)
screenshot from 2016-06-14 02-34-46

#### new repo homepage (not polished, but no worse than current homepage)
screenshot from 2016-06-14 02-34-26


beyang:
The CSS appears buggy to me on staging4:
image

And locally, too:
image

beyang:
The search modal for global search still appears to use a different component?

beyang:
The new repo page looks very off-center and it wasn't clear to me where I should start on the page.

sqs:
@beyang is going to take this over

PR #105, opened 2016-06-14
gitserver: Avoid unnecessary bytes.Buffer for Run and Output

keegancsmith:
This is a hotspot when doing heap profiles, and a large amount of the time we
read it into a buffer then throw it away. This change slightly complicates exec
to avoid always reading into a buffer.

Future work: Callers which parse large outputs use a streaming interface.


keegancsmith:
@shurcooL PTAL

pkg/gitserver/exec_test.go
```diff
@@ -108,6 +108,72 @@ func TestExec(t *testing.T) {
 	}
 }
 
+func TestExec_outputs(t *testing.T) {
+	server := make(chan *request)
+	servers = [](chan<- *request){server}
```
shurcooL: Fun observation, this has got to be one of the most unusual lines of Go I've seen in a while. Perfectly valid, logical, but just unusual looking, hehe.

`[](chan<- *request){server}`

keegancsmith:
@neelance PTAL

keegancsmith:
After chatting with @neelance we are not sure this will make that much of a difference, so lets just avoid complicating the code. Abandon

PR #107, opened 2016-06-14
Require RepoRev or Query when hitting /defs endpoint

attfarhan:
Requires RepoRev or Query when hitting /defs http api endpoint.

##### Reviewer tasks


beyang:
@attfarhan can you add a unit test for this case?

services/backend/defs_test.go
```diff
@@ -34,6 +34,11 @@ func TestDefsService_List_Repos(t *testing.T) {
 	if !*calledReposResolve {
 		t.Error("!calledReposResolve")
 	}
+
+	_, err = s.List(ctx, &sourcegraph.DefListOptions{})
+	if err == nil {
+		t.Fatal(err)
```
beyang: Should use `t.Error`, since you're reporting an error the test is designed to catch. `t.Fatal` should be used if there's some precondition that should always be satisfied that isn't.

beyang: Also, print an understandable error message. Like "expected error for def list request with unspecified repository revision and empty query, but got none".

services/backend/defs.go
```diff
@@ -123,6 +123,11 @@ func (s *defs) List(ctx context.Context, opt *sourcegraph.DefListOptions) (*sour
 		return nil, err
 	}
 
+	if len(opt.RepoRevs) == 0 && opt.Query == "" {
+		err := fmt.Errorf("RepoRev and Query required")
```
beyang: This is not accurate. Should be, "Either RepoRev or Query should be non-empty".

beyang:
Looks like this is causing some other tests in CI to fail. Once you fix those and address the above comments, this should look good.

PR #109, opened 2016-06-15
Quality local search

beyang:
##### Reviewer tasks

##### Test plan

Test on staging:
- Go to gorilla/mux.
- Search for something locally with the syntax you're accustomed to be supported by global search. Observe that it works.
- Now go to a specific commit in gorilla/mux that has been built. Observe that search works here, too.
- Try to break the search box or make it run super slow with queries.


app/web_modules/sourcegraph/tree/TreeSearch.js
```diff
@@ -63,6 +64,48 @@ function pathDir(path: string): string {
 	return pathJoin(parts.splice(0, parts.length - 1));
 }
 
+type TreeSearchProps = {
```
renfredxh: The [method recommended by flow](https://flowtype.org/docs/react.html#defining-components-as-reactcomponent-subclasses) to define state types is to put them in the state/props attributes inside the component class. Any reason to change this? Saves defining a couple new variables.

app/web_modules/sourcegraph/search/SearchActions.js
```diff
@@ -2,42 +2,40 @@
 
 import type {Def} from "sourcegraph/def";
 
-export class WantResults {
+export type WantResultsProps = {
```
renfredxh: This is a nice improvement, but "Props" might be too much of an overloaded term in react-land, would be better just to call them `WantResultsPayload` since that's familiar [flux terminology](https://facebook.github.io/flux/docs/overview.html#actions)

app/web_modules/sourcegraph/tree/TreeSearch.js
```diff
@@ -188,19 +222,42 @@ class TreeSearch extends Container {
...
-		// Global search results only show up for admin users
-		if (this.context.user && this.context.user.Admin && (prevState.query !== nextState.query || prevState.repo !== nextState.repo)) {
-			Dispatcher.Backends.dispatch(new SearchActions.WantResults(nextState.query, null, [nextState.repo], GLOBAL_DEFS_LIMIT));
+		} else {
+			if (prevState.query !== nextState.query || prevState.repo !== nextState.repo) {
+				if (nextState.rev && nextState.rev.length === 40) {
+					Dispatcher.Backends.dispatch(new SearchActions.WantResults({
+						query: nextState.query,
+						repos: [nextState.repo],
+						notRepos: null,
```
renfredxh: This is an optional field so the null shouldn't be necessary.

app/web_modules/sourcegraph/tree/TreeSearch.js
```diff
@@ -188,19 +222,42 @@ class TreeSearch extends Container {
...
+						commitID: nextState.rev,
+						limit: LOCAL_DEFS_LIMIT,
+					}));
+				} else {
+					Dispatcher.Backends.dispatch(new SearchActions.WantResults({
+						query: nextState.query,
+						repos: [nextState.repo],
+						notRepos: null,
```
renfredxh: Same as above

app/web_modules/sourcegraph/tree/TreeSearch.js
```diff
@@ -188,19 +222,42 @@ class TreeSearch extends Container {
...
+						limit: LOCAL_DEFS_LIMIT,
+					}));
+				}
+			}
+			Dispatcher.Backends.dispatch(new SearchActions.WantResults({
+				query: nextState.query,
+				repos: null,
```
renfredxh: 🔥 null

services/backend/internal/localstore/defs.go
```diff
@@ -0,0 +1,533 @@
...
+	Name string `db:"name"`
+	Kind string `db:"kind"`
+	File string `db:"file"`
+	Data []byte `db:"data"`
+
+	Language  dbLang     `db:"language"`
+	RefCount  int        `db:"ref_ct"`
+	UpdatedAt *time.Time `db:"updated_at"`
+
+	BoW string `db:"bow"`
```
renfredxh: Possibly stupid question, but what is `BoW`?

services/backend/internal/localstore/defs.go
```diff
@@ -0,0 +1,533 @@
...
+}
+
+type dbDef struct {
+	Rev    int64 `db:"rid"`   // foreign key to repo_rev table
+	DefKey int64 `db:"defid"` // foreign key to def_key table
+
+	Name string `db:"name"`
+	Kind string `db:"kind"`
+	File string `db:"file"`
+	Data []byte `db:"data"`
```
renfredxh: Comment on what kind of data (`sourcegraph.Def`?) goes here would be helpful

services/backend/internal/localstore/global_defs.go
```diff
@@ -9,38 +9,53 @@ import (
...
-// (doing so would require a database migration). When adding a new ID, you
-// must use the exact LOWERCASE srclib toolchain language name.
-var languages = map[string]uint16{
-	"go":     1,
-	"java":   2,
-	"python": 3,
+// dbLang is a numerical identifier that identifies the language of a definition
+// in the database. NOTE: this values of existing dbLang constants should NOT be
+// changed. Doing so would require a database migration.
+type dbLang uint16
```
renfredxh: Might be too late to change at this point, but this seems like something that would be better to do within Postgres itself. They have [ENUM types](https://www.postgresql.org/docs/9.1/static/datatype-enum.html) that can do this sort of thing. This way it's type safe and less fragile than relying on the app to maintain this list. It's 4 bytes instead of 2 but that shouldn't be a big deal.

PR #110, opened 2016-06-15
builds: Add test to ensure images are pullable

keegancsmith:


beyang:
This will test if the images are pullable on CI, but the version of Docker on CI is not necessarily the same as the version in the worker image, right?

keegancsmith:
yes, this PR is still incomplete, I pushed it out just before heading out for the long weekend and forgot to mark it as WIP. It is a base to add that check.

keegancsmith:
Also this test seems unreliable + takes 172s. I am going to abandon it.

If I find some extra time I'll look into how easy it is to query the remote registry the metadata. Otherwise, there is more important stuff to focus on for now.

PR #111, opened 2016-06-15
services/worker/plan: Update srclib-go to incorporate sourcegraph/srclib-go#102.

shurcooL:
Undo revert commit 78d51f514d504f9c1b491638cf28c73ee9760c5c. This time, it's built with the right Docker version and should work.

##### Reviewer tasks

- [ ] DOCKER VERSION: reviewer approves image was built using right docker version

##### Test plan

Tested on staging3.

https://staging3.sourcegraph.com/github.com/shurcooL/play/-/builds/1
https://staging3.sourcegraph.com/github.com/shurcooL/play/-/def/GoPackage/github.com/shurcooL/play/187/tictactoe/-/Player

(Compare with https://sourcegraph.com/github.com/shurcooL/play/-/builds/8.)

Worked successfully, and indexed good Go packages despite there being bad Go packages that refer to missing repos/packages. So it also shows that fix in https://github.com/sourcegraph/srclib-go/pull/102 works.

### DIFF
--- a/services/worker/plan/srclib_images.go
+++ b/services/worker/plan/srclib_images.go
@@ -10,7 +10,7 @@ import (
 // `droneSrclibGoImage = "sourcegraph/srclib-go"`.
 var (
 	droneSrclibBashImage       = "sourcegraph/srclib-bash@sha256:c26891ddab9f7e73138927bba8dc323c0ba4c5abdb82b4a56ac4b23063e204e9"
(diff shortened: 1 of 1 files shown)


shurcooL:
Also see https://staging3.sourcegraph.com/github.com/camlistore/camlistore/-/builds/1.

shurcooL:
Hmm, looks like camlistore build failed. Looking into it, but it looks like a different reason.

shurcooL:
It looks like it'll take more work to make camlistore build successfully. But this PR still introduces an improvement compared to before, so this isn't a blocker for getting it merged.

PR #117, opened 2016-06-15
changed the scroll blob so that the directory path isn't hidden on sc…

CarolineFrost:
before:

screen shot 2016-06-15 at 3 24 16 pm

after:

screen shot 2016-06-15 at 3 23 49 pm

### DIFF
--- a/app/web_modules/sourcegraph/components/styles/_colors.css
+++ b/app/web_modules/sourcegraph/components/styles/_colors.css
@@ -49,6 +49,7 @@
 @value c_cool-pale-gray rgba(213, 229, 242, 1);
 @value c_cool-pale-gray-8 rgba(213, 229, 242, 0.8);
 @value c_cool-pale-gray-5 rgba(213, 229, 242, 0.5);
+@value c_cool-pale-gray-3 hsla(207, 53%, 98%, 1);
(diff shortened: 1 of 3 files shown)


app/web_modules/sourcegraph/components/styles/_colors.css
```diff
@@ -49,6 +49,7 @@
 @value c_cool-pale-gray rgba(213, 229, 242, 1);
 @value c_cool-pale-gray-8 rgba(213, 229, 242, 0.8);
 @value c_cool-pale-gray-5 rgba(213, 229, 242, 0.5);
+@value c_cool-pale-gray-3 hsla(207, 53%, 98%, 1);
```
chexee: The numbers here indicate the alpha opacity – notice the values above and below all have the same RGB values, but different alpha opacity values. 

What's the thought behind changing the background color? Is it so there's no overlap when you scroll behind the breadcrumb?

If that's the case, I'd recommend using `c_pale-blue` here for now, but this is a note for me: we should include opaque levels of all these colors and not solely depend on opacity for gradiated color values :)

rothfels:
Closing in favor of https://github.com/sourcegraph/sourcegraph/pull/127

PR #118, opened 2016-06-15
UX: Redirect to tools page for users not using chrome when clicking browser CTA

KattMingMing:
##### Description

A temporary fix to redirect logic and checklist state for users not using Chrome. Also fixes a layout bug on the tool component page.

##### Reviewer tasks


uforic:
LGTM - low risk, and better logic around buttons. If you've tested locally, good to merge!

PR #120, opened 2016-06-16
browser-ext: refactor / add tests

rothfels:
The last part is to refactor the annotations code and make it support diff blobs. I'll do this next but on a separate PR.

(Some of the newer experimental PR annotations stuff that John/Matt did last Saturday is will WIP/commented out, but going to be touched by the next diff so I'm just leaving it in there for now).


client/browser-ext/chrome/extension/inject.js
```diff
@@ -1,516 +1,182 @@
...
-	e.href = url;
-	if (sameRepo) e.dataset.pjax = "#js-repo-pjax-container";
-	if (sameRepo) e.classList.add("js-navigation-open");
-	document.body.appendChild(e);
-	e.click();
-	setTimeout(() => document.body.removeChild(e), 1000);
+	// Add invisible div to the page to indicate injection has completed.
+	if (!document.getElementById("sourcegraph-app-bootstrap")) {
+		let el = document.createElement("div");
+		el.id = "sourcegraph-app-bootstrap";
```
uforic: pop this bad boy out into its own variable, it's referred to 5 times in the code-base, so I figure better to avoid typos and have it be a constant.

client/browser-ext/chrome/extension/inject.js
```diff
@@ -1,516 +1,182 @@
...
+		el.style.display = "none";
+		document.body.appendChild(el);
+	}
 }
 
-window.addEventListener("load", bootstrapApp);
+window.addEventListener("load", injectModules);
+document.addEventListener("pjax:success", () => {
+	removeSearchFrame();
```
uforic: This action is triggered on every GitHub page transition - what if there is no search frame? Could be catch all, another possible solution is each component as a deconstructor, and we just iterate across components and call deconstruct. Could be excess work considering its pretty simple though, so maybe not worth it,

client/browser-ext/chrome/extension/inject.js
```diff
@@ -1,516 +1,182 @@
...
+		const info = fileInfos[i];
+		const infoFilePath = isDelta ? info.querySelector(".user-select-contain").title : path;
+		const blobAnnotatorId = `sourcegraph-blob-annotator-${infoFilePath}`;
+		let blobAnnotatorContainer = document.getElementById(blobAnnotatorId);
+		if (!blobAnnotatorContainer) { // prevent injecting build indicator twice
+			blobAnnotatorContainer = document.createElement("span");
+			blobAnnotatorContainer.id = blobAnnotatorId;
+			blobAnnotatorContainer.style.display = "none";
+			info.appendChild(blobAnnotatorContainer);
+			injectComponent(<BlobAnnotator path={infoFilePath} blobElement={blobs[i]} />, blobAnnotatorContainer);
```
uforic: {blobs[i]} could be null I think, should we double check that it isn't? really edge case, but I guess with these overloaded for loops worth making sure. Or make sure that render won't break on null (can't tell if it will).

client/browser-ext/chrome/extension/inject.js
```diff
@@ -1,516 +1,182 @@
...
+		injectComponent(<Background />, backgroundContainer);
 	}
+}
 
-	_renderDefInfo(props) {
-		const def = props.def.content[keyFor(props.repo, props.rev, props.defPath)];
-
-		const id = "sourcegraph-def-info";
-		let e = document.getElementById(id);
+function injectBlobAnnotator() {
```
uforic: injectBlobAnnotator -> injectBlobAnnotators?

client/browser-ext/chrome/extension/inject.js
```diff
@@ -1,516 +1,182 @@
...
-			indicatorText = "Unsupported file"
-		} else {
-			indicatorText = "Indexing...";
-		}
-
-		const fileInfo = document.querySelector(".file-info");
-		const buildIndicator = document.getElementById("sourcegraph-build-indicator");
-		if (fileInfo && !buildIndicator && window.location.href.split("/")[5] !== "pull") { // don't add build indicator on PRs
+function injectBuildIndicators() {
+	if (!isGitHubURL()) return;
```
uforic: Thing about popping these out to the main inject section - so its clear that "oh, these 4 are on GitHub, and these 2 are on Sourcegraph", as opposed to having the checks in function.

client/browser-ext/app/components/BlobAnnotator.js
```diff
@@ -0,0 +1,139 @@
...
+		}
+	}
+
+	_refresh() {
+		this.props.actions.getAnnotations(this.props.repo, this.props.rev, this.props.path);
+	}
+
+
+	componentWillReceiveProps(nextProps) {
+		this._addAnnotations(nextProps);
```
uforic: This is going to be sick. Nice.

client/browser-ext/chrome/extension/inject.js
rothfels: Actually, I don't think so.

We actually don't want the main app and the browser extension to share any code.

client/browser-ext/chrome/extension/inject.js
rothfels: I'm assuming it's invariant that there is a file header per blob; if that breaks, the exception that will start to be thrown is as good as any to indicate to us that something is broken. We'll want to look into it, too, so I'm glad it would complain.

Speaking of which, we should hook up Sentry to this bad boy so we can get error reporting (and btw if you aren't on Sentry, get on it -- real time view of exceptions going on for users in your apps)

client/browser-ext/chrome/extension/inject.js
rothfels: It's smart enough to noop if there is no search frame.

We actually don't want to remove it from the page. I'm just manually toggling show/hide of components that are both always rendered on the page. I've renamed it "hide" to make it more obvious.

client/browser-ext/chrome/extension/inject.js
rothfels: Ah, thing is, we need them to be function invocations in here.

This function is going to get called multiple times during its lifecycle, and without the function being invoked like this we would't be able to update/stay in sync w/ the global state.

attfarhan:
Noticed some odd behavior when running this locally (along with the annotations refactor) related to srclib-data-versions and annotations returning 404s: http://cl.ly/0k2U2d3G372E. I start at https://github.com/gorilla/mux/blob/master/mux.go, and then click on string which leads me to Go std lib. When I get there, annotations aren't added and the fetch to annotations endpoint 404s. If I refresh the page, though, the annotations do get injected successfully. But now if I press the back button to go back to gorilla/mux, srclib-data-version 404s there, even though it was successful when I landed there initially.

PR #122, opened 2016-06-16
Trying to resolve SHA of custom default branch

alexsaveliev:
This is my proposal how to fix https://app.asana.com/0/87040567695724/144531999537159. It includes
- trying to resolve SHA when repository contains custom default branch (for example, github.com/apache/log4j). If we were failed first time, trying to query default branch from GitHub and retry
- using updated `sourcegraph/drone-git` Docker image based on github.com/sqs/drone-git#5. Without it on my local instance I cannot start build of github.com/apache/log4j because clone fails

Please ensure that this PR does not adds too many extra calls to GitHub

### DIFF
--- a/services/backend/internal/localstore/global_defs.go
+++ b/services/backend/internal/localstore/global_defs.go
@@ -16,6 +16,7 @@ import (
 	"sourcegraph.com/sourcegraph/sourcegraph/pkg/search"
 	"sourcegraph.com/sourcegraph/sourcegraph/pkg/store"
 	"sourcegraph.com/sourcegraph/sourcegraph/services/backend/accesscontrol"
+	"sourcegraph.com/sourcegraph/sourcegraph/services/ext/github"
 	"sourcegraph.com/sourcegraph/srclib/graph"
 	sstore "sourcegraph.com/sourcegraph/srclib/store"
 	"sourcegraph.com/sqs/pbtypes"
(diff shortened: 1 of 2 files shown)


services/backend/internal/localstore/global_defs.go
```diff
@@ -489,7 +490,16 @@ func resolveRevisionDefaultBranch(ctx context.Context, repo int32) (repoPath, co
 	}
 	c, err := vcsrepo.ResolveRevision(repoObj.DefaultBranch)
```
keegancsmith: this should always be set, but may have an incorrect value of `master`. https://github.com/sourcegraph/sourcegraph/pull/9 potentially solves that in the correct place.

I'm surprised if this change fixes the linked issue, since I'd assume only a small percentage of repos would suffer from this bug, while the described bug seems to have a wider impact.

keegancsmith: nevermind, we only set those fields in the repo service, not the repo store :/
Another frustrating part of our store vs services infra. cc @beyang @neelance.

rothfels:
@keegancsmith does this look god to merge?

services/backend/internal/localstore/global_defs.go
```diff
@@ -489,7 +490,16 @@ func resolveRevisionDefaultBranch(ctx context.Context, repo int32) (repoPath, co
...
 	if err != nil {
-		return
+		if !strings.HasPrefix(strings.ToLower(repoObj.URI), "github.com/") {
```
keegancsmith: this block of code shouldn't exist anymore, since we now sync metadata. However, I assume it is still an issue. I'm going to add a reference to the issue tracking the underlying problem and merge this in.
https://app.asana.com/0/87040567695724/147734985562458

beyang: FYI global_defs isn't isn't used in prod anymore and will soon go away. New impl is in defs.go

keegancsmith:
Had to update to a new API we use for accessing GitHub. If tests are green will merge.

services/backend/internal/localstore/global_defs.go
keegancsmith: this helper is still used by other implementations.

PR #123, opened 2016-06-16
Add examples, documentation, and a few more components to the styleguide

chexee:


app/web_modules/sourcegraph/components/Affix.js
```diff
@@ -0,0 +1,36 @@
...
+		className: React.PropTypes.string,
+		style: React.PropTypes.object,
+		children: React.PropTypes.any,
+		offset: React.PropTypes.number,
+	};
+
+	componentDidMount() {
+		const {_affix} = this.refs;
+		const initialOffset = _affix.offsetTop;
+		window.addEventListener("scroll", (e) => {
```
KattMingMing: Since we are adding a listener. I think we will also need to add ` window.removeEventListener` inside of `componentWillUnmount`

chexee:
Alright, there's a working, reviewable version on staging: https://staging2.sourcegraph.com/styleguide

KattMingMing:
This is really awesome and incredibly helpful. Thanks for adding all the documentation to explain when and were to use each of the components.

KattMingMing:
For the checklist item would it make sense to have the text stop before the button on all rows if a button exists in any row? Screenshot below:

Without a button all of the text wraps at the same spot:
screenshot 2016-06-18 10 47 24

With a button the text wraps at different spot if a button is visible

screenshot 2016-06-18 10 45 23

chexee:
@KattMingMing I think that'll be a future enhancement for the feature – from a design perspective, I'd actually prefer to have a consistent right alignment width, rather than whitespace if there is no CTA

PR #127, opened 2016-06-17
changed the scroll blob so that the directory path isn't hidden on sc…

CarolineFrost:
updated colors from directorypaths

### DIFF
--- a/app/web_modules/sourcegraph/app/styles/App.css
+++ b/app/web_modules/sourcegraph/app/styles/App.css
@@ -26,6 +26,8 @@ the entire viewport height.
 .breadcrumb {
 	composes: bb ph3 pv3 from base;
 	composes: b--cool-pale-gray-8 from colors;
+	position: fixed;
+	width: 100%;
 }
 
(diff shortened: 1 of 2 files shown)


sqs:
@chexee can you please take a look at this?

app/web_modules/sourcegraph/components/styles/_colors.css
```diff
@@ -49,6 +49,7 @@
 @value c_cool-pale-gray rgba(213, 229, 242, 1);
 @value c_cool-pale-gray-8 rgba(213, 229, 242, 0.8);
 @value c_cool-pale-gray-5 rgba(213, 229, 242, 0.5);
+@value c_cool-pale-gray-3 hsla(207, 53%, 98%, 1);
```
chexee: Same comment as the last branch – This value should be a percentage opacity. If this needs to be solid, please add a new color or use an existing color.

Also, prefer rgba to hsla

chexee:
LGTM!

PR #128, opened 2016-06-17
Improve global refs update

beyang:
This minimizes the transaction zone of the global refs update step, which should hopefully improve perf.

It also forces a global_refs reindex on every successful srclib import. This will make `src repo sync -f $REPO` work as expected (previously, this wouldn't actually force a rebuild of the global_refs index, which was confusing when debugging).

### DIFF
--- a/services/backend/internal/localstore/global_refs.go
+++ b/services/backend/internal/localstore/global_refs.go
@@ -291,6 +291,10 @@ func (g *globalRefs) getRefStats(ctx context.Context, defKeyID int64) (int64, er
 }
 
 func (g *globalRefs) Update(ctx context.Context, op *sourcegraph.DefsRefreshIndexOp) error {
+	if err := accesscontrol.VerifyUserHasWriteAccess(ctx, "GlobalRefs.Update", op.Repo); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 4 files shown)


beyang:
cc @keegancsmith

services/backend/internal/localstore/global_refs.go
```diff
@@ -488,6 +479,7 @@ func (g *globalRefs) versionUpdate(tx gorp.SqlExecutor, repoPath string, commitI
 	return err
 }
 
+// StatRefresh refreshes the global_refs_stats tables. This should ONLY be called from test code.
```
shurcooL: > This should ONLY be called from test code.

If that's the case, can it be moved to `_test.go` files?

keegancsmith:
This gets rid of us having one process updating the table for a repo, to multiple processing doing it concurrently. The way it was done previously was intentional to prevent concurrent updates failing all the time .

If I understand, the issue you want to actually fix is multiple concurrent processes all waiting to update for a repo, which before the async worker was not an issue (we just had an idle goroutine), but now that the async workers are limited is an issue. Lets chat offline about what to do about that / if this is an alright fix.

PR #129, opened 2016-06-17
Annotations overhaul (no diff support yet, just a rewrite of the old logic -- with removal of tricky bugs)

rothfels:
@uforic @attfarhan here's just the new annotation code. Pretty much all of the old stuff is removed. And a lot of bugs with it.


client/browser-ext/app/utils/annotations.js
```diff
@@ -4,177 +4,259 @@ import styles from "../components/App.css";
...
 
-		// off-by-one case
-		if (annsByStartByte[byteCount].EndByte - annsByStartByte[byteCount].StartByte === 1) {
-			return `${insert}</a>`;
-		}
-		cacheDefaultBranch(matchDetails.URL)
-		annotating = true;
-		return insert;
+	// The logic here is to simply recurse on each of the child nodes; everything is eventually
+	// just a text node or the special-cased "quoted string node" (see below).
```
attfarhan: Is this true? It seems like in most cases we reach a token which has no children, which is an element node. For example: https://github.com/gorilla/mux/blob/master/mux.go#L38, the `Router` token has the form `<span class="pl-v">Router</span>` and has no children, and has nodeType elementNode, so it seems to me that this would go into infinite recursion.  Can you explain?

client/browser-ext/app/utils/annotations.js
```diff
@@ -4,177 +4,259 @@ import styles from "../components/App.css";
...
+}
+
+// isQuotedStringNode is a predicate function to identify the special-cased
+// string identifier DOM element, which takes a like of go code like this:
+//
+//    import (
+//        "fmt"
+//    )
+//
+// and creates for the "fmt" and creates this:
```
attfarhan: Typo?

attfarhan:
Noticed some odd behavior when running this locally related to srclib-data-versions and annotations returning 404s: http://cl.ly/0k2U2d3G372E.  I start at https://github.com/gorilla/mux/blob/master/mux.go, and then click on `string` which leads me to Go std lib.  When I get there, annotations aren't added and the fetch to annotations endpoint 404s.  If I refresh the page, though, the annotations do get injected successfully.  But now if I press the back button to go back to `gorilla/mux`, srclib-data-version 404s there, even though it was successful when I landed there initially.  It seems it may be related to the other PR/first part of the refactor, so I'll repost this there.   

Made a couple of comments above to clarify some of the logic, but otherwise +1.

client/browser-ext/app/utils/annotations.js
rothfels: .children is different from .childNodes. In your example, `Router` is a child node of type Node.TEXT_NODE

PR #132, opened 2016-06-18
UX: Workaround for react-router hash scrolling

KattMingMing:
##### Description

Currently react router hash scrolling is not working correctly. This work around allows us to use `<Link to="/foo hash="bar"..` to scroll to `id=bar` on `/foo`. The resulting url takes the form of `...com/foo#bar`. This works on both reload, navigation to a new page and also scrolling on the same page. Since this is defined inside  of `browser.js` this means it will work on every route. Referenced from an ongoing issue filed in 2014. https://github.com/reactjs/react-router/issues/394

##### Reviewer tasks


KattMingMing:
If we are cool with this workaround then PR #123 can have it's eventListener workaround for scrolling removed

chexee:
oh, yes please!

chexee:
It would be good to have someone more qualified actually code review though :)

app/web_modules/sourcegraph/init/browser.js
```diff
@@ -26,6 +26,21 @@ context.reset(window.__sourcegraphJSContext);
 resetOnAuthChange();
 __webpack_public_path__ = document.head.dataset.webpackPublicPath; // eslint-disable-line no-undef
 
+// Work around for react-router hash scroll behavior - https://github.com/reactjs/react-router/issues/394
+function hashLinkScroll() {
+	const {hash} = window.location;
```
shurcooL: Something's odd here. The constant is named `hash`, but it's being assigned `window.location`.

Did you mean for this to be `const {hash} = window.location.hash;` here?

KattMingMing: @shurcooL I see what you're saying and the syntax here might be confusing so happy to change it, but this works correctly. What is happening here is as follows:

For the URL `http://localhost:3080/search#about`
`const hash = window.location` would set hash to: `Location {hash: "#about", search: "", pathname: "/search", port: "3080", hostname: "localhost"…}` 
However in the case of setting `const {hash} = window.location` we actually end up with `"#about"` as a string.

shurcooL: Ah, gotcha. I didn't know about that constant syntax, I thought it was the same as `const hash = ...`. Thanks for explaining!

KattMingMing: You bet! :)

app/web_modules/sourcegraph/init/browser.js
```diff
@@ -48,7 +63,7 @@ function matchWithRedirectHandling(recursed) {
 			return;
 		}
 
-		setTimeout(() => ReactDOM.render(<Router {...renderProps} render={applyRouterMiddleware(useScroll(shouldUpdateScroll))} />, document.getElementById("main")));
+		setTimeout(() => ReactDOM.render(<Router {...renderProps} onUpdate={hashLinkScroll} render={applyRouterMiddleware(useScroll(shouldUpdateScroll))} />, document.getElementById("main")));
```
rothfels: Why are you doing this here instead of e.g. in `App` (where you can define an onEnter handler)? That might make it so you don't need to do the janky `setTimeout` bit too.

Also, any objection moving the `hashLinkScroll` function int the module where `shouldUpdateScroll` is defined?

app/web_modules/sourcegraph/init/browser.js
chexee: yay ES6 :)

chexee:
What's the eta on this? :) looks like there are a few open questions/comments.

KattMingMing:
@chexee will get to this tonight / tomorrow morning

app/web_modules/sourcegraph/init/browser.js
KattMingMing: Good call, I'll move that stuff over

KattMingMing: @rothfels Actually I'm pretty confused what you mean. I didn't add this `setTimeout` and I'm not quite sure where to add this in `App`. Could you be more specific?

PR #133, opened 2016-06-19
Tokenize camelCase and snake_case global search results

attfarhan:
This change tokenizes camelCased and snake_cased paths in global search results by splitting out the individual words and adding them to the bag of words representations for the results.  This improves search results, since we can now match based on parts of words in the path instead of only full words.  Right now, all separated words from camelCased and snake_cased words are given a score of 1, while an additional 10 points are given for words that are separated from the last part of the path (typically the "name" of the result).

##### Reviewer tasks

### DIFF
--- a/pkg/search/search_tok.go
+++ b/pkg/search/search_tok.go
@@ -1,6 +1,7 @@
 package search
 
 import (
+	"github.com/fatih/camelcase"
 	"path/filepath"
 	"regexp"
 	"strings"
@@ -36,7 +37,26 @@ func BagOfWords(def *graph.Def) map[string]int {
 		words[w] += 2
 		if len(defParts)-1 == i {
 			words[w] += 30 // mega extra points for matching the last component of the def path (typically the "name" of the def)
+
+			if snakeSplit := snakeCaseSplit(w); len(snakeSplit) > 1 {
+				snakeSplit := snakeCaseSplit(w)
+				for _, tokPart := range snakeSplit {
+					words[tokPart] += 10 // more points for matching last component of def path
+				}
+			} else {
+				camelSplit := camelcase.Split(w)
+				for _, tokPart := range camelSplit {
+					words[tokPart] += 10 // more points for matching last component of def path
+				}
+			}
 		}
+	}
+	camelCaseWords, snakeCaseWords := splitCaseWords(defParts)
+	for _, w := range camelCaseWords {
+		words[w]++ // add each individual word of camel cased tokens
+	}
+	for _, w := range snakeCaseWords {
+		words[w]++ // add each individual word of snake cased tokens
 	}
 	for i, w := range fileParts {
 		words[w]++
@@ -49,6 +69,27 @@ func BagOfWords(def *graph.Def) map[string]int {
 	words[def.Kind] += 1
 
 	return words
+}
+
+func splitCaseWords(defParts []string) ([]string, []string) {
+	var camelCaseWords []string
+	var snakeCaseWords []string
+
+	for _, w := range defParts {
+		if len(snakeCaseSplit(w)) > 1 {
+			snakeCaseWords = append(snakeCaseWords, snakeCaseSplit(w)...)
+		} else {
+			camelCaseWords = append(camelCaseWords, camelcase.Split(w)...)
+		}
+	}
+	return camelCaseWords, snakeCaseWords
+}
+
+func snakeCaseSplit(src string) (entries []string) {
+	if strings.Split(src, "_")[0] != src {
+		entries = strings.Split(src, "_")
+	}
+	return entries
 }
 
 func UserQueryToksToTSQuery(toks []string) string {
--- a/pkg/search/search_tok_test.go
+++ b/pkg/search/search_tok_test.go
@@ -1,0 +1,88 @@
+package search
+
+import (
+	"reflect"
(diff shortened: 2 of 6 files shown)


pkg/search/search_tok.go
```diff
@@ -32,12 +33,44 @@ func BagOfWords(def *graph.Def) map[string]int {
...
+
 		if len(defParts)-1 == i {
 			words[w] += 30 // mega extra points for matching the last component of the def path (typically the "name" of the def)
+
+			if len(snakeCaseSplit(w)) > 1 {
+				snakeSplit := snakeCaseSplit(w)
+				for _, tokPart := range snakeSplit {
+					words[tokPart] += 10 // more points for matching last component of def path delimited by a space
+				}
+				continue // camelcase.Split will split snake_cased tokens but include _ in the BoW, so skip next case
```
beyang: can we first detect if a path is camel-case or snake case? 99% of the time, it's one or the other.

pkg/search/search_tok.go
```diff
@@ -32,12 +33,44 @@ func BagOfWords(def *graph.Def) map[string]int {
...
 		}
 	}
+	var camelCaseWords []string
+	var snakeCaseWords []string
 	for i, w := range defParts {
 		words[w] += 2
+
+		if len(camelcase.Split(w)) > 1 {
+			camelCaseWords = append(camelCaseWords, camelcase.Split(w)...)
+			words[strings.Join(camelcase.Split(w), " ")]++ // add the camel case word delimited by a space into bag of words
```
beyang: why add the space separated words as a single word?

pkg/search/search_tok.go
```diff
@@ -32,12 +33,44 @@ func BagOfWords(def *graph.Def) map[string]int {
...
+			if len(camelcase.Split(w)) > 1 {
+				camelSplit := camelcase.Split(w)
+				for _, tokPart := range camelSplit {
+					words[tokPart] += 10 // more points for matching last component of def path delimited by a space
+				}
+			}
 		}
 	}
+	for _, w := range camelCaseWords {
+		words[w]++ // add each individual word of camel cased tokens
```
beyang: why increment the count again here? Isn't this double counting from above?

pkg/search/search_tok.go
```diff
@@ -32,12 +33,44 @@ func BagOfWords(def *graph.Def) map[string]int {
 			words[w] += 2
 		}
 	}
+	var camelCaseWords []string
```
beyang: I would factor out the computation of `camelCaseWords` and `snakeCaseWords` into a separate function. Call out to those functions. Then you can just iterate over the returned arrays and `words[word] += wordWeight` in a single, readable for-loop.

pkg/search/search_tok.go
```diff
@@ -32,12 +33,44 @@ func BagOfWords(def *graph.Def) map[string]int {
...
+	var snakeCaseWords []string
```
beyang: This function is now complicated enough that you should add a unit test to test both expected behavior and that edge cases don't break.

pkg/search/search_tok.go
attfarhan: Above we've only incremented +10 for words that are in the last component of the def path, anything else hasn't yet been incremented.  Although the ones that are +10 would get an additional 1 point, it seems to work nicely for results.

attfarhan:
@beyang I've updated this, PTAL

pkg/search/search_tok.go
```diff
@@ -36,8 +37,27 @@ func BagOfWords(def *graph.Def) map[string]int {
 		words[w] += 2
 		if len(defParts)-1 == i {
 			words[w] += 30 // mega extra points for matching the last component of the def path (typically the "name" of the def)
+
+			if len(snakeCaseSplit(w)) > 1 {
```
beyang: `if snakeSplit := snakeCaseSplit(w); len(snakeSplit) > 1 {`

pkg/search/search_tok.go
```diff
@@ -51,6 +71,27 @@ func BagOfWords(def *graph.Def) map[string]int {
...
+	var snakeCaseWords []string
+
+	for _, w := range defParts {
+		if len(snakeCaseSplit(w)) > 1 {
+			snakeCaseWords = append(snakeCaseWords, snakeCaseSplit(w)...)
+		} else {
+			camelCaseWords = append(camelCaseWords, camelcase.Split(w)...)
+		}
+	}
+	return &camelCaseWords, &snakeCaseWords
```
beyang: No need to return references here. The slice struct itself holds a pointer to the head of its contents, so effectively slices are already passed by reference.

pkg/search/search_tok_test.go
```diff
@@ -0,0 +1,77 @@
+package search
+
+import (
+	"reflect"
+	"sourcegraph.com/sourcegraph/srclib/graph"
+	"testing"
+)
+
+func TestBagOfWords(t *testing.T) {
```
beyang: To make this more readable, it's good to keep the expected output close to input for each test case. A well established pattern to do this is to define a test case struct inside this function. E.g.,

```
type testcase struct {
  def *graph.Def
  expected map[string]int
}

cases := []testcase{{
  def: graph.DefKey{ ... },
  expected: map[string]int{ ... },
}, {
...
}}
```

PR #134, opened 2016-06-20
fix defs table update

beyang:
Fix the update of defs table to make the update repo-wide, as opposed to one source unit at a time (which was overwriting previous source unit data).


services/httpapi/srclib_import.go
```diff
@@ -100,6 +100,14 @@ func serveSrclibImport(w http.ResponseWriter, r *http.Request) (err error) {
 		return fmt.Errorf("srclib import of %s failed: %s", repo.URI, err)
 	}
 
+	// Update defs table in DB
+	if _, err := cl.Defs.RefreshIndex(ctx, &sourcegraph.DefsRefreshIndexOp{
+		Repo:     repoRev.Repo,
+		CommitID: repoRev.CommitID,
+	}); err != nil {
+		return err
+	}
```
shurcooL: Do we want this to happen for forks too?

shurcooL:
Not seeing any issues in the code.

services/httpapi/srclib_import.go
beyang: Yes, it should happen for any repository that we build.

beyang:
Thanks for the quick review.

PR #137, opened 2016-06-21
Blob ui updates

renfredxh:
Update blob UI styles to match some improvements from latest mocks

image

### DIFF
--- a/app/web_modules/sourcegraph/blob/styles/Blob.css
+++ b/app/web_modules/sourcegraph/blob/styles/Blob.css
@@ -9,7 +9,7 @@
 @value c_blue from colors;
 @value contentWidth 690px;
 @value marginWidth 290px;
-@value codeLineHeight 16px;
+@value codeLineHeight 20px;
 @value firstCodeLineTopPadding 8px;
 
 .pathComponent {}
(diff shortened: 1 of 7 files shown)


app/web_modules/sourcegraph/blob/styles/Blob.css
```diff
@@ -118,6 +118,9 @@
 	overflow: visible;
 	white-space: pre;
 }
+.lineContent a {
+	font-weight: bold;
```
renfredxh: I actually like the way this looks in may cases, but sometimes it's overwhelming in areas with a ton of linked code. Maybe because I'm just not used to it. Example: 

screen shot 2016-06-20 at 6 03 12 pm

@chexee what do you think?

rothfels:
@renfredxh is there a font weight between normal and bold?

app/web_modules/sourcegraph/components/styles/code.css
```diff
@@ -40,57 +40,6 @@
 		color: #15b8ae;
 	}
 	:global .fun {
-		color: #0287bd;
+		color: #00a4de;
 	}
 }
-
-/*
-PREVIOUS THEME
```
rothfels: Why are we keeping this around?

rothfels:
Two small comments otherwise LGTM

renfredxh:
@rothfels depends, for Menlo (the monospace font) as far as I could tell there were only two weights. Something in-between would be ideal but it doesn't exist for the font we're using right now.

app/web_modules/sourcegraph/components/styles/code.css
renfredxh: Not really sure, that's why I got rid of it

PR #138, opened 2016-06-21
Set Secure Flag on Cookies

ggilmore:


sqs:
Can you please rebase instead of leaving in these merge commits to clean up the git history?

app/auth/cookie.go
```diff
@@ -28,6 +29,10 @@ const sessionCookieName = "session-oauth2-token"
 // HTTP request.
 var ErrNoSession = errors.New("no session cookie")
 
+// OnlySecureCookies is an environment variable that should be set to true iff
+// all issued cookies should have the "secure" flag enabled
```
sqs: period to end sentence

app/auth/cookie.go
```diff
@@ -28,6 +29,10 @@ const sessionCookieName = "session-oauth2-token"
...
+var OnlySecureCookies = conf.GetenvBool("SG_SECURE_COOKIES")
```
sqs: name the env var with the same words as the var name for consistency (`SG_ONLY_SECURE_COOKIES`)

sqs:
Is there any reason to have a separate env var to determine whether to use secure-only cookies, vs. just checking whether `conf.AppURL(ctx).Scheme == "https"`? Would you ever want to use an HTTPS URL for the app and NOT use secure cookies?

samlavery:
If you wanted to terminate SSL somewhere that isn't the primary application or rpc server is one reason you might want to keep them split.

sqs:
Even if you terminate SSL at a frontend proxy server, the app's canonical URL would still be HTTPS, though, right? (As for our current setup for Sourcegraph.com.)

samlavery:
Yeah, but I assumed that setting was used for more than prefixing outbound urls, maybe it's not.

sqs:
Nope, not anything substantially more, afaik. You can double check with Sourcegraph, of course. 😄

app/app.go
```diff
@@ -38,7 +38,9 @@ func NewHandlerWithCSRFProtection(handler http.Handler) http.Handler {
 	// Prevent setting a different cookie for subpaths if someone
 	// directly visits a subpath.
 	h.SetBaseCookie(http.Cookie{
-		Path: "/",
+		Path:     "/",
+		HttpOnly: true,
```
shurcooL: What's the rationale for setting `HttpOnly` to true here?

samlavery: to prevent xss attacks

shurcooL: Do we need our own scripts to be able to access these cookies, or not?

samlavery: I would wonder why, since the browser will automatically handle sending them back the sever, what does our client side JS need with them?

app/app.go
```diff
@@ -34,18 +34,23 @@ import (
 // 	<input type="hidden" name="csrf_token" value="{{$.CSRFToken}}">
 //
 func NewHandlerWithCSRFProtection(handler http.Handler) http.Handler {
-	h := nosurf.New(handler)
-	// Prevent setting a different cookie for subpaths if someone
-	// directly visits a subpath.
-	h.SetBaseCookie(http.Cookie{
-		Path: "/",
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
```
sqs: Remove unnecessary newline

app/auth/cookie.go
```diff
@@ -28,6 +30,12 @@ const sessionCookieName = "session-oauth2-token"
 // HTTP request.
 var ErrNoSession = errors.New("no session cookie")
 
+// OnlySecureCookies takes a context object, and returns whether or
+// not the secure flag should be set of for all issued cookies.
```
sqs: remove "of"

cli/auth_cmds.go
```diff
@@ -81,7 +81,10 @@ func (c *authCookieCmd) Execute(args []string) error {
 	fmt.Fprintln(os.Stderr, "// Auth cookie for the given OAuth2 access token:")
 	fmt.Fprintln(os.Stderr)
 
-	sess, err := appauth.NewSessionCookie(appauth.Session{AccessToken: c.AccessToken})
+	// Since there is no context object to derive the isSecure flag from, isSecure
+	// is arbitrarilly set to false. This is of no consquence since this field isn't used
```
sqs: typo: cons*e*quence

services/httpapi/auth.go
```diff
@@ -127,11 +127,13 @@ func finishLoginOrSignup(ctx context.Context, cl *sourcegraph.Client, w http.Res
...
 	}
 
+	secureCookie := appauth.OnlySecureCookies(ctx)
+
 	// Authenticate future requests.
 	ctx = sourcegraph.WithCredentials(ctx, oauth2.StaticTokenSource(&oauth2.Token{TokenType: "Bearer", AccessToken: tok.AccessToken}))
 
 	// Authenticate as newly created user.
-	if err := appauth.WriteSessionCookie(w, appauth.Session{AccessToken: tok.AccessToken}); err != nil {
+	if err := appauth.WriteSessionCookie(w, appauth.Session{AccessToken: tok.AccessToken}, secureCookie); err != nil {
```
sqs: unnecessary local var `secureCookie`, just inline the expr

app/internal/oauth2client/github.go
```diff
@@ -236,11 +236,12 @@ func linkAccountWithGitHub(w http.ResponseWriter, r *http.Request, ctx context.C
 
 	// Write cookie.
 	cred := sourcegraph.CredentialsFromContext(ctx)
+	secureCookie := appauth.OnlySecureCookies(ctx)
```
sqs: unnecessary local var `secureCookie`, just inline the expr

also, this expr is in between the decl of `cred` and its only use, so even if it were necessary, it should be placed as close to the usage site as possible and not in between the decl/use of something else

app/auth/cookie.go
```diff
@@ -28,6 +30,12 @@ const sessionCookieName = "session-oauth2-token"
 // HTTP request.
 var ErrNoSession = errors.New("no session cookie")
 
+// OnlySecureCookies takes a context object, and returns whether or
```
sqs: "takes a context object" is implied by the func signature, so you can just say `// OnlySecureCookies indicates whether ...`

sqs:
LGTM after fairly minor comments

app/auth/cookie.go
shurcooL: The most idiomatic verb for funcs that return bool is "reports". Instead of "returns true if ...".

Source: Rob Pike.

PR #140, opened 2016-06-21
Add repo origin fields for storing mirror repo origin information

sqs:
See individual commit messages.


shurcooL:
Overall, this looks solid. I came up with some medium comments while reading it, I'll post them inline.

api/sourcegraph/sourcegraph.proto
```diff
@@ -16,6 +16,52 @@ option (gogoproto.unmarshaler_all) = true;
...
+	//
+	// If there are multiple API versions for a service, separate
+	// entries may be added per API version. In that case, the
+	// APIBaseURL may need to differ as well. The API client code is
+	// responsible for handling these cases.
+	enum ServiceType {
+		// None indicates that there is no origin (the resource lives
+		// canonically on the Sourcegraph server itself). This is used for
+		// non-mirror repos, for example.
+		None = 0;
```
shurcooL: Couple thoughts here:

1. Based on the comment above, is there any reason not to call this `Sourcegraph` instead of `None` ?
2. Based on https://github.com/sourcegraph/sourcegraph/pull/140/files#diff-02250b78eaa5a1b5e8e2f6f3a72991b3R182, is there a reason to even have `None`? If `Origin` is not set for non-mirror repos, when would `None` be used?

api/sourcegraph/sourcegraph.proto
```diff
@@ -418,7 +470,16 @@ message ReposCreateOp {
...
-		int32 FromGitHubID = 2;
+
+		// FromGitHubID creates a new repo that mirrors a repo hosted
+		// on GitHub.com.
+		//
+		// DEPRECATED: Use Origin instead.
+		int32 FromGitHubID = 2 [deprecated=true];
+
+		// Origin creates a new repo mirrored from an external
+		// service.
```
shurcooL: Having `None` as one of service types makes this confusing. What happens if you use Origin and set None? Is it the same as `NewRepo`? This question would go away if `None` service type didn't exist.

api/sourcegraph/sourcegraph.proto
```diff
@@ -16,6 +16,52 @@ option (gogoproto.unmarshaler_all) = true;
...
+
+	// Service is the type service that the resource canonically lives
+	// on. It is used to determine which API client should be used to
+	// access it on the origin service (e.g., GitHub vs. Bitbucket).
+	ServiceType Service = 2;
+
+	// APIBaseURL is the base URL to the API of the origin service for
+	// the resource. (E.g., "https://api.github.com" for
+	// GitHub.com-hosted repos.)
+	string APIBaseURL = 3;
```
shurcooL: What's the motivation to allow leaving this blank for "default"? Is it to save space in database? To be able to dynamically change the default endpoint value without having to update all the rows in database?

It's something you would still have to do if you ever wanted to update enterprise API urls, but I don't think we'd ever want to do that (instead, opt to create new repos at the new url, optionally remove old ones).

---

The reason I ask is because the logic and documentation would be simpler if you could just always use this value directly, without having to pass it through `checkValidOriginAndSetDefaultURL` to set the default URL. So I'd like to hear the benefits mentioned explicitly to justify that.

However, I suspect it may be worth it to use blank string to refer to "default API" endpoint. In that case, I suggest making it more explicit and not allow ambiguity - to set the default API endpoint, this *must* be blank.

Basically, right now, both the documentation and the expected rows seemingly could use either "https://api.github.com" OR blank string to indicate default URL. I think we should establish/force one canonical way. Then documentation is clearer, and the database model is simpler.

app/web_modules/sourcegraph/repo/RepoBackend.js
```diff
@@ -90,8 +90,13 @@ const RepoBackend = {
 
 		case RepoActions.WantCreateRepo:
 			{
-				let body = {Op: {FromGitHubID: action.remoteRepo.GitHubID}};
-				if (!action.remoteRepo.GitHubID) { // non-GitHub repositories
+				let body;
+				if (action.remoteRepo.GitHubID) {
+					body = {
+						Op: {Origin: {ID: action.remoteRepo.GitHubID.toString(), Service: 1}},
```
shurcooL: Use a named constant here instead of magic value 1.

api/sourcegraph/sourcegraph.proto
sqs: Good point. We do not need a zero None value in this type, as it is only used in the Origin message, which should be null/nil to indicate that there is no origin.

api/sourcegraph/sourcegraph.proto
```diff
@@ -16,6 +16,52 @@ option (gogoproto.unmarshaler_all) = true;
...
+		// the URL to the GitHub Enterprise server's API.
+		GitHub = 1;
+	}
+
+	// ID is an identifier for the resource on its origin
+	// service. Although numeric IDs are used on many services (GitHub
+	// and Bitbucket, for example), this field is a string so that it
+	// supports non-numeric IDs (which are used on Google Cloud
+	// Platform and probably other services that Sourcegraph might
+	// support in the future).
```
shurcooL: How confident are you that GCP uses non-numeric IDs as actual IDs and not names that can be changed? That's surprising. Is it UUIDs or something?

It's unfortunate not being able to use an integer for IDs when that's what they are in most cases, but if there's indeed some chance of non-numeric IDs to be supproted, then we don't have much choice. I just wanted to confirm before we make this compromise.

api/sourcegraph/sourcegraph.proto
sqs: I didn't want to require every API client (Chrome ext, our own app JS, our CLI) to pass the constant string "https://api.github.com". It is filled in at the gRPC method level, though, so it's always the full value in the DB.

If you are using Sourcegraph and it's connected to both GitHub.com and your GitHub Enterprise, I can see it being useful for API clients to be explicit about which one they want. People might assume that the default is their GitHub Enterprise if 90% of the mirrored repos on their Sourcegraph are mirrored from their GitHub Enterprise. Is that justification enough to allow specifying "https://api.github.com" as the string itself and also using that as the default value if the API input is empty?

api/sourcegraph/sourcegraph.proto
sqs:
```
$ gcloud alpha source repos list --format json
[
  {
    "createTime": "2015-11-06T15:19:03.904Z",
    "id": "Vte3env5lhT",
    "name": "default",
    "projectId": "sourcegraph-dev",
    "state": "LIVE",
    "vcs": "GIT"
  }
]
```

(changed the ID but retained the format, in case it's a secret value)

sqs:
Addressed everything but https://github.com/sourcegraph/sourcegraph/pull/140#discussion_r67829240, waiting on your final reply.

api/sourcegraph/sourcegraph.proto
shurcooL: Agreed.

Let's just change the documentation only then. In `ReposCreateOp` (not here), make it more clear that `Origin.APIBaseURL` can be one of:

- blank string for default API endpoint for the service,
- an actual API endpoint URL.

sqs: 👍

shurcooL:
LGTM. I think the data model is sound. Nice tests! You should feel free to merge when CI is fixed and you're ready.

sqs:
on https://staging4.sourcegraph.com

services/backend/internal/localstore/repos.go
```diff
@@ -174,6 +192,24 @@ func finishGetRepo(ctx context.Context, repo *sourcegraph.Repo) (*sourcegraph.Re
 		if err := accesscontrol.VerifyActorHasGitHubRepoAccess(ctx, auth.ActorFromContext(ctx), "Repos.Get", repo.ID, repo.URI); err != nil {
 			return nil, err
 		}
+
+		// Automatically migrate repos that don't have their origin fields set.
+		//
+		// NOTE: This can be removed when all repos in the DB have their
+		// origin fields set.
```
keegancsmith: task to follow up on this?

PR #141, opened 2016-06-21
builds: Prevent adding docker images prod does not support

keegancsmith:
In production we only support docker images which are v1. If you build an image
with docker v0.10 or higher, it breaks srclib in production. This commit adds a
test which speaks to the docker registry to ensure the image exists and is an
old enough version.

This was tested by using the image in ff1c35f6a9796e7117623848234ea96a10c4d4f3
which introduced a bad image. Also we manually modified a sha to be non-existant
to test 404.

### DIFF
--- a/services/worker/plan/srclib_test.go
+++ b/services/worker/plan/srclib_test.go
@@ -1,7 +1,13 @@
 package plan
 
 import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
 	"reflect"
+	"strings"
+	"sync"
 	"testing"
 
 	"gopkg.in/yaml.v2"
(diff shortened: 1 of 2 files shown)


services/worker/plan/srclib_test.go
```diff
@@ -242,3 +248,85 @@ func config2yaml(c droneyaml.Config) string {
 	}
 	return string(b)
 }
+
+func TestImages_Hash(t *testing.T) {
```
beyang: Add a short description of what the test does

beyang:
Minor comment. Otherwise LGTM

keegancsmith:
I'll hold off merging this until https://github.com/sourcegraph/sourcegraph/pull/147 is merged in, since it solves the core problem this test tries to prevent. Once merged I'll update the test case to just check existence of images + only allow tags.

sqs:
Is this good to merge? #147 is merged.

keegancsmith:
Updated tests. Safe to merge once CI is happy.

PR #143, opened 2016-06-21
added blog to global nav bar

CarolineFrost:
Blog option is now available under the user dropdown in global nav


sqs:
LGTM after CI passes

PR #145, opened 2016-06-22
Documentation search

attfarhan:
Index documentation to produce relevant results for natural language queries.

##### Reviewer tasks

### DIFF
--- a/services/backend/internal/localstore/global_defs.go
+++ b/services/backend/internal/localstore/global_defs.go
@@ -180,7 +180,15 @@ func (g *globalDefs) Search(ctx context.Context, op *store.GlobalDefSearchOp) (*
 		// The ranking critieron is the weighted sum of xref count,
 		// text similarity score, and whether the last term matches
 		// the name.
-		scoreSQL = `5.0*log(10 + ref_ct) + 100.0*ts_rank(to_tsvector('english', bow), to_tsquery('english', ` + arg(bowQuery) + `)) + 100.0*((LOWER(name)=LOWER(` + arg(lastTok) + `))::int) score`
+
+		if bowQuerySplit := strings.Split(bowQuery, " & "); len(bowQuerySplit) > 2 {
+			fmt.Println("More than 3")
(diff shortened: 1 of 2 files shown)


attfarhan:
Results from local instance with some queries:
http://cl.ly/1F3S1I3w0Q1y
http://cl.ly/40423C1J1k3B
http://cl.ly/1e1d1g1m1v2q

beyang:
^ these look promising!

services/backend/internal/localstore/global_defs.go
```diff
@@ -180,7 +180,12 @@ func (g *globalDefs) Search(ctx context.Context, op *store.GlobalDefSearchOp) (*
 		// The ranking critieron is the weighted sum of xref count,
 		// text similarity score, and whether the last term matches
 		// the name.
-		scoreSQL = `5.0*log(10 + ref_ct) + 100.0*ts_rank(to_tsvector('english', bow), to_tsquery('english', ` + arg(bowQuery) + `)) + 100.0*((LOWER(name)=LOWER(` + arg(lastTok) + `))::int) score`
+		if bowQuerySplit := strings.Split(bowQuery, " & "); len(bowQuerySplit) > 2 {
```
beyang: Operating on the tsquery is hacky. Better to apply the transformation to the original (`op.TokQuery`). That way, the behavior is tied to what the user actually types into the search bar, not to something we've already transformed a bit for another purpose.

How many cases have you tested this out on? Is 2 the ideal split? What if I have a 3-component symbol that I'm looking for (e.g., gorilla/mux.Router)? Does search still work well in that situation? Can you feature flag this so we can play around with it? E.g., you can enable it only for admin users (like here: https://github.com/sourcegraph/sourcegraph/blob/cb259045902836e6de77508f1f420f933ff4d80f/app/web_modules/sourcegraph/tree/TreeSearch.js#L197)

beyang:
@attfarhan I'll take this over this week so you can focus other stuff

beyang:
fixes https://github.com/sourcegraph/sourcegraph-issues/issues/23

PR #147, opened 2016-06-22
use tags for versioning of srclib toolchain images

neelance:
This is to avoid a breaking change in docker that makes pulling images by digest fail in production, see https://docs.docker.com/engine/breaking_changes/.


services/worker/dockerfiles/README.md
```diff
@@ -12,13 +12,13 @@ Instructions for deploying srclib updates to Sourcegraph.com
 
 ```
 make clean && make srclib
```
keegancsmith: unrelated nit: this can just be `make clean srclib`

beyang:
LGTM

PR #148, opened 2016-06-22
Global: More instrumentation for RefreshIndexes

keegancsmith:
We currently don't have a good idea of what parts are slow. This adds fine-grained instrumentation all over global_refs and global_def updates.


neelance:
LGTM

PR #149, opened 2016-06-22
Expire Password Reset Tokens

ggilmore:
Adds functionality to expire password reset tokens both after usage, and after a configurable period of time.


ggilmore:
I will squash commit once this change is accepted.

ggilmore:
The SQL required would look something like this: 

```postgres
ALTER TABLE password_reset_requests ADD COLUMN expires_at timestamp;
UPDATE password_reset_requests SET expires_at now() + '4 hours';
ALTER TABLE password_reset_requests ALTER COLUMN expires_at SET NOT NULL;
```

pkg/amortize/amortize_helper.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"crypto/rand"
+	"fmt"
+	"math/big"
+)
+
+// ShouldAmortize is a non-deterministic function that returns true 'a' out of
+// every 'b' times on average. ShouldAmortize panics if a > b.
+func ShouldAmortize(a, b int) (bool, error) {
```
shurcooL: Nothing is said about what happens if a or b are negative, does that mean they're allowed?

shurcooL: From looking at the code, I can see `big.NewInt(int64(b))` will panic if b is 0 or less.

pkg/amortize/amortize_helper.go
```diff
@@ -0,0 +1,25 @@
...
+	if a <= 0 || b <= 0 {
+		panic("a and b must be positive")
+	}
+	if a > b {
+		panic("a must be <= b")
+	}
+
+	bigN, err := rand.Int(rand.Reader, big.NewInt(int64(b)))
+	if err != nil {
+		return false, fmt.Errorf("Error when trying to generate a random number: %s", err)
```
sqs: do not capitalize, do not begin with "error" (redundant) https://github.com/golang/go/wiki/CodeReviewComments#error-strings

services/backend/internal/localstore/accounts.go
```diff
@@ -120,11 +122,21 @@ func init() {
...
 
 type passwordResetRequest struct {
-	Token string
-	UID   int32
+	Token     string
+	UID       int32
+	ExpiresAt time.Time `db:"expires_at"`
 }
 
+const passwordReset time.Duration = 4 * time.Hour
```
sqs: `time.Duration` explicit type is not necessary

services/backend/internal/localstore/accounts.go
```diff
@@ -120,11 +122,21 @@ func init() {
 }
 
 type passwordResetRequest struct {
-	Token string
-	UID   int32
+	Token     string
+	UID       int32
+	ExpiresAt time.Time `db:"expires_at"`
```
sqs: Include the migration SQL in the relevant commit message

services/backend/internal/localstore/accounts.go
```diff
@@ -136,33 +148,64 @@ func (s *accounts) RequestPasswordReset(ctx context.Context, user *sourcegraph.U
...
 		return nil, fmt.Errorf("Error saving password reset token: %s", err)
 	}
 	return &sourcegraph.PasswordResetToken{Token: token}, nil
 }
 
+var epoch = time.Unix(0, 0)
+
+func unmarshalResetRequest(ctx context.Context, token string) (passwordResetRequest, error) {
+	var req passwordResetRequest
+	return req, appDBH(ctx).SelectOne(&req, `SELECT * FROM password_reset_requests WHERE Token=$1`, token)
```
sqs: this is tricky - it is not immediately clear to me that the `req` value is modified by the second return expr before its value is captured for the first return expr

services/backend/internal/localstore/accounts.go
```diff
@@ -136,33 +148,64 @@ func (s *accounts) RequestPasswordReset(ctx context.Context, user *sourcegraph.U
...
 	} else if err != nil {
 		return genericErr
 	}
-	log15.Info("Resetting password", "store", "Accounts", "UID", req.UID)
-	if err := (password{}).SetPassword(ctx, req.UID, newPass.Password); err != nil {
-		return fmt.Errorf("Error changing password: %s", err)
+	if time.Now().Before(req.ExpiresAt) {
+		log15.Info("Resetting password", "store", "Accounts", "UID", req.UID)
+		if err := (password{}).SetPassword(ctx, req.UID, newPass.Password); err != nil {
+			return fmt.Errorf("Error changing password: %s", err)
```
sqs: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

services/backend/internal/localstore/accounts.go
```diff
@@ -136,33 +148,64 @@ func (s *accounts) RequestPasswordReset(ctx context.Context, user *sourcegraph.U
...
+		}
+		// Updating the expiration date of the token to the epoch will permanently expire it.
+		req.ExpiresAt = epoch
+		count, err := appDBH(ctx).Update(&req)
+		if err != nil {
+			log15.Warn("Error updating token", "store", "Accounts", "error", err)
+		}
+		if count > 1 {
```
sqs: Is it ever possible for count to be greater than 1? The table has a primary key, so there shouldn't ever be more than 1 row with the same token value.

services/backend/internal/localstore/accounts.go
```diff
@@ -136,33 +148,64 @@ func (s *accounts) RequestPasswordReset(ctx context.Context, user *sourcegraph.U
...
+			log15.Warn("Updated more than one token", "store", "Accounts", "updated", count)
+		}
+		return nil
 	}
+	return grpc.Errorf(codes.InvalidArgument, "this token has expired")
```
sqs: clarify error msg, like "password reset token has expired"

services/backend/internal/localstore/accounts.go
```diff
@@ -136,33 +148,64 @@ func (s *accounts) RequestPasswordReset(ctx context.Context, user *sourcegraph.U
...
+}
 
-	if _, err := appDBH(ctx).Exec(`DELETE FROM password_reset_requests WHERE Token=$1`, newPass.Token.Token); err != nil {
-		log15.Warn("Error deleting token", "store", "Accounts", "error", err)
-		return nil
+// cleanExpiredResets deletes password reset requests from the DB whose expiration date has passed.
+func (s *accounts) cleanExpiredResets(ctx context.Context) error {
+	res, err := appDBH(ctx).Exec(`DELETE FROM password_reset_requests WHERE expires_at < now()`)
+	if err != nil {
+		return fmt.Errorf("Error when cleaning up expired password resets: %s", err)
```
sqs: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

services/backend/internal/localstore/accounts.go
```diff
@@ -136,33 +148,64 @@ func (s *accounts) RequestPasswordReset(ctx context.Context, user *sourcegraph.U
...
+	}
+	rows, err := res.RowsAffected()
+	if err != nil {
+		log15.Warn("Error when counting the expired password reset requests that we cleaned", "store", "Accounts", "error", err)
+	} else {
+		log15.Info("Cleaned up expired password reset requests", "store", "Accounts", "removed", rows)
```
sqs: This seems like a debug log to me...it will so rarely be important

pkg/amortize/amortize_helper.go
sqs: In what cases can rand.Int fail? Does it have a maximum block timeout waiting for entropy or something?

services/backend/internal/localstore/accounts.go
```diff
@@ -136,33 +148,64 @@ func (s *accounts) RequestPasswordReset(ctx context.Context, user *sourcegraph.U
...
-	log15.Info("Resetting password", "store", "Accounts", "UID", req.UID)
-	if err := (password{}).SetPassword(ctx, req.UID, newPass.Password); err != nil {
-		return fmt.Errorf("Error changing password: %s", err)
+	if time.Now().Before(req.ExpiresAt) {
+		log15.Info("Resetting password", "store", "Accounts", "UID", req.UID)
+		if err := (password{}).SetPassword(ctx, req.UID, newPass.Password); err != nil {
+			return fmt.Errorf("Error changing password: %s", err)
+		}
+		// Updating the expiration date of the token to the epoch will permanently expire it.
+		req.ExpiresAt = epoch
```
sqs: I don't think epoch needs to be its own pkg-scope var. You can just say `req.ExpiresAt = time.Time{}`, which is more Go-idiomatic. Check that this works with PostgreSQL (it should work and set a time before `now()`, which is all that's necessary to have it cleaned up in the code below).

services/backend/internal/localstore/accounts.go
sqs: lowercase column name `token` for consistency (postgresql actually doesn't care)

services/backend/internal/localstore/accounts.go
sqs: give a more explicit variable name like `passwordResetTokenExpiration`

services/backend/internal/localstore/accounts.go
```diff
@@ -120,11 +122,21 @@ func init() {
...
-	UID   int32
+	Token     string
+	UID       int32
+	ExpiresAt time.Time `db:"expires_at"`
 }
 
+const passwordReset time.Duration = 4 * time.Hour
+
 func (s *accounts) RequestPasswordReset(ctx context.Context, user *sourcegraph.User) (*sourcegraph.PasswordResetToken, error) {
+	doClean, err := amortize.ShouldAmortize(1, 1000)
```
sqs: Explain why 1000 was chosen, or just say it's a random guess (which is totally fine). Will help anyone who needs to tweak it in the future (if ever).

services/backend/internal/localstore/accounts_db_test.go
```diff
@@ -205,11 +217,22 @@ func TestAccounts_ResetPassword_ok(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-
+	oldReq, err := unmarshalResetRequest(ctx, token.Token)
+	if err != nil {
+		log.Fatal(err)
```
sqs: `t.Fatal` not `log.Fatal`

services/backend/internal/localstore/accounts_db_test.go
```diff
@@ -229,3 +252,30 @@ func TestAccounts_ResetPassword_badtoken(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	newPass := &sourcegraph.NewPassword{Password: "a", Token: &sourcegraph.PasswordResetToken{Token: token.Token}}
+	if err := s.ResetPassword(ctx, newPass); err != nil {
+		t.Fatal(err)
+	}
+	s.cleanExpiredResets(ctx)
+	_, err = unmarshalResetRequest(ctx, newPass.Token.Token)
+	if err != sql.ErrNoRows {
+		t.Fatalf("Should have gotten a NoRow error when trying to retrieve a cleaned reset request, got this error instead: %s", err)
```
sqs: don't capitalize `Should`, for consistency

sqs: use (actual, expected) order, not (expected, actual) - https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures. so put "got this ..." before "should have ..."

services/backend/internal/localstore/accounts.go
ggilmore: Discussed offline.

PR #150, opened 2016-06-23
Analytics: Update browser extension tracking

KattMingMing:
##### Description

Updates our extensions to include more accurate tracking for usage across multiple integrations. This will allow us to accurately measure usage of users across multiple integrations.


KattMingMing:
Ready for review, just need to fix this flow issue

client/browser-ext/app/components/Background.js
```diff
@@ -76,12 +78,17 @@ export default class Background extends React.Component {
...
 		document.removeEventListener("click", this._clickRef);
+		document.addEventListener("updateIdentityForExtension", this._updateIdentity);
 		if (this._updateIntervalID !== null) {
 			clearInterval(this._updateIntervalID);
 			this._updateIntervalID = null;
 		}
 	}
 
+	_updateIdentity(e) {
+		this.props.actions.updateIdentity(e.detail);
```
rothfels: Can we just do this directly (without the redux flow)? Redux doesn't seem necessary.

client/browser-ext/app/components/Background.js
```diff
@@ -157,6 +164,12 @@ export default class Background extends React.Component {
 
 			this._renderDefInfo(this.props, urlProps);
 		});
+
+		chrome.runtime.sendMessage(null, {type: "getIdentity"}, {}, (identity) => {
+			if (identity) {
+				EventLogger.updateAmplitudePropsForUser(identity);
```
rothfels: Why are we only identifying the user on clicks? What about if the user is just highlighting tokens? What if they are using search?

This doesn't feel like the right place.

app/web_modules/sourcegraph/util/EventLogger.js
```diff
@@ -336,6 +344,7 @@ export function withViewEventsLogged(Component: ReactClass): ReactClass {
 			// values are updated.
 			if (this.props.location.pathname !== nextProps.location.pathname) {
 				this._logView(nextProps.routes, nextProps.location);
+				document.dispatchEvent(new CustomEvent("updateIdentityForExtension", this.context.eventLogger.getAmplitudeIdentificationProps()));
```
rothfels: Since these events are on the global namespace, probably a good idea to namespace the event string under "sourcegraph" or something like that.

rothfels: I.e. `sourcegraphUpdateIdentity`

client/browser-ext/app/components/Background.js
KattMingMing: This is called when the page is loaded, I'm not sure I follow

client/browser-ext/app/components/Background.js
```diff
@@ -76,12 +78,17 @@ export default class Background extends React.Component {
 		document.removeEventListener("pjax:success", this._cleanupAndRefresh);
 		document.removeEventListener("popstate", this._popstateUpdate);
 		document.removeEventListener("click", this._clickRef);
+		document.addEventListener("updateIdentityForExtension", this._updateIdentity);
```
rothfels: Let's add this listener in `inject.js`, not this Background component.

client/browser-ext/app/components/Background.js
KattMingMing: Added this to the redux flow to make it easier to pass additional event information that could alter the state since this is an event listener from the main app into the extension.

client/browser-ext/app/components/Background.js
KattMingMing: Sounds good!

rothfels:
@KattMingMing cleaned up and merged: https://github.com/sourcegraph/sourcegraph/commit/40307bd25c3484205ff5e7e593e10640740b44b4

rothfels:
(Feel free to take a look and make sure everything looks good.)

PR #153, opened 2016-06-23
global_refs: Use a bulk upsert / query for defKeyIDs

keegancsmith:
We re-use the helpers that were created for defs. They use a much better pattern
for accessing the data from postgres. I haven't been able to reliably benchmark
on my machine if this will improve performance, but the real world metrics we
are measuring heavily indicate that this will be orders of magnitude faster.
Also note that surprisingly this is by far the slowest part of global_refs
update (roughly 80% of the time taken in the p99 case).

Fixes https://app.asana.com/0/146253180133821/147371528518098


neelance:
Improve performance AND reduce complexity. Like! LGTM

keegancsmith:
FYI The results that are coming in show a great improvement :)

neelance:
Cool!

PR #154, opened 2016-06-23
add special case for the import path of the Kubernetes repo

neelance:
There is no information in the K8s repository to automatically derive the correct import path. This is why a manual special case is required.


keegancsmith:
It's a shame we have to do this. But thanks for fixing LGTM

PR #155, opened 2016-06-23
return users to original page after logging in

sqs:
For users who have already created an account and are logging in again, this change makes it so they are returned to the original page they were viewing. New account creation still returns the user to the tour page.

TODOs
* [ ] @KattMingMing: Change Google Analytics to not attribute users coming from GitHub OAuth to GitHub

TEST PLAN
* [x] Try logging in using all possible GitHub auth links (login modal, signup modal, etc.) on both homepage and non-root page
* [x] Try logging in from the embedded product CTAs (homepage buttons, tools page, DefInfo, Blob, "my repositories", tour)
* [x] Try the /login and /signup pages. Log in or sign up from those pages directly.


app/web_modules/sourcegraph/util/EventLogger.js
```diff
@@ -370,6 +370,13 @@ export function withViewEventsLogged(Component: ReactClass): ReactClass {
 				}
 
 				this.context.eventLogger.logEventForCategory(AnalyticsConstants.CATEGORY_EXTERNAL, AnalyticsConstants.ACTION_REDIRECT, this.props.location.query._event, eventProperties);
+
+				// After recording the event properties, remove them from the querystring.
```
sqs: @KattMingMing Can you review this added section? No need to review the rest of the diff.

sqs: Nevermind, I removed that change from this PR.

PR #157, opened 2016-06-23
Add 'from' keyword to javascript list of keywords

antonina-cherednichenko:
In construction 'import' ... 'from' ... from plays role of keyword, was not previously included in the list of javascript keywords


alexsaveliev:
- Please run `gofmt -w pkg/coverageutil/javascript.go` to fix build
- I'd rather would add special processing of `import .. from` statements, but adding `from` to the list of keywords LGTM

PR #158, opened 2016-06-24
app/TreeSearch: make search case-insensitive

mkovacs:
Fix to make fuzzy search case insensitive.
See https://app.asana.com/0/135425975204116/147877634399138

Tested locally.


sqs:
LGTM

PR #159, opened 2016-06-24
{api,services/backend}: remove Users.Count method

sqs:
This was only used to set the first account created as admin. Admin does not
mean much in the current system, so this is not necessary.


sqs:
@beyang ok'd removal of smoke tests

shurcooL:
LGTM.

PR #161, opened 2016-06-24
store: Simplify batch def_keys API

keegancsmith:


keegancsmith:
gonna not use pg9.5, since CI is not co-operating yet. Pushing a version of this without the move to pg9.5 upsert

neelance:
LGTM

PR #162, opened 2016-06-24
OneFlow: Update ID on global nav and footer

KattMingMing:
##### Description

Update ID on global nav and footer


renfredxh:
LGTM

PR #164, opened 2016-06-24
app/TreeSearch: add fuzzy scoring for search results

mkovacs:
Improve tree search: sort results by their fuzzy match score.

It's live on staging3.sourcegraph.com

### DIFF
--- a/app/web_modules/sourcegraph/tree/TreeSearch.js
+++ b/app/web_modules/sourcegraph/tree/TreeSearch.js
@@ -4,6 +4,7 @@ import React from "react";
 import ReactDOM from "react-dom";
 import {Link} from "react-router";
 import fuzzysearch from "fuzzysearch";
+import fuzzy_score from "sourcegraph/tree/fuzzy_score";
 import Container from "sourcegraph/Container";
 import Dispatcher from "sourcegraph/Dispatcher";
 import debounce from "lodash/function/debounce";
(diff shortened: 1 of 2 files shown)


app/web_modules/sourcegraph/tree/TreeSearch.js
```diff
@@ -331,14 +324,21 @@ class TreeSearch extends Container {
...
-			} else if (nextState.fileList && nextState.fileList.Files) {
-				nextState.fileResults = nextState.fileList.Files
-					.filter((f) => fuzzyMatchPath(nextState.query, f))
-					.map((f) => ({
-						name: f,
-						isDirectory: false,
-						url: urlToBlob(nextState.repo, nextState.rev, f),
-					}));
+			} else {
+				// NOTE(mate): moving this constant declaration inside the if-block seems to trigger
```
rothfels: Is the note necessary to leave in here? Sounds like you're just qualifying that this shouldn't be moved into the `if` block, but if what you're saying is true, then `flow` is going to complain at the next dev who decides to try, so there's no value in keeping the comment here.

rothfels:
@mkovacs LGTM, let's get this merged.

rothfels:
(Left one minor comment which you can consider prior to merging.)

app/web_modules/sourcegraph/tree/TreeSearch.js
mkovacs: It took me about 10 minutes of head-scratching to figure out why Flow was complaining about `nextState.fileList.Files` being possibly undefined right after it's checked. I'm hoping that note will save somebody 10 minutes.

PR #166, opened 2016-06-25
browser extension zips have hash and version number in filename

uforic:

##### Test plan

cd into client/browser-ext. Run `npm run dist`. Verify that the resulting chrome and firefox zip files work on their respective browsers!


client/browser-ext/scripts/build-env/dist.sh
```diff
@@ -1,11 +1,14 @@
...
 rm -rf node_modules
 rm -rf build/
 rm -rf dev/
 npm install
 npm run build
 cd /browser-ext/build
-zip -r /browser-ext/firefox-sourcegraph-dist.xpi *
-zip -r /browser-ext/chrome-sourcegraph-dist.zip *
+export version_string=`grep \"version\" manifest.json`
+version_string=${version_string:14:6}
```
rothfels: This isn't the best idea, since it'll break e.g. as soon as we go from `1.1.xx` to `1.2.0`

rothfels: I might just recommend sed'ing out the `\s?"version":\s?` and suffix'd `",`

rothfels: Or use regex groups, though I'm not sure the unix-fu to do that.

rothfels:
Also, before we submit this to Firefox, there's a TODO to resolve in the `EventLogger` for disambiguating Chrome vs. Firefox events before we submit this to Firefox, just flagging here now so we don't forget.

sqs:
`s/GITHUB_HASH/GIT_HASH/g` then LGTM

PR #167, opened 2016-06-25
app/web_modules: Fix ineffective signed in check on about page.

shurcooL:
This change enables `signedIn` to be set when rendering the about page. That page uses it to not display "Sign up with GitHub" buttons when the user is already signed in.

Currently, the about page always displays those because those `signedIn` is always `undefined`, no matter if the user is logged in or not:

image

image

Setting `AboutPage.contextTypes` seems to fix that. I got the idea from [the code for pricing page](https://github.com/sourcegraph/sourcegraph/blob/47f606ac15d644cc93623ffef9eb2e930b8952cd/app/web_modules/sourcegraph/page/PricingPage.js#L109-L112). However, I am not familiar with React's details enough to know if this is the best way.

##### Reviewer tasks

I am not familiar with React's details enough to know if this is the best and most correct fix, or if the issue should be resolved in another way.

- [x] REACT: reviewer approves this is the proper React way to fix this issue

##### Test plan

I tested this change locally while working on a feature branch. It resolved the bug. I've recreated this change from memory, however.

### DIFF
--- a/app/web_modules/sourcegraph/page/AboutPage.js
+++ b/app/web_modules/sourcegraph/page/AboutPage.js
@@ -67,5 +67,8 @@ function AboutPage(props, {signedIn}): React$Element {
 		</div>
 	);
 }
+AboutPage.contextTypes = {
+	signedIn: React.PropTypes.bool,
+};
 
 export default CSSModules(AboutPage, styles);


rothfels:
@shurcooL thanks, this looks like the correct solution to me. LGTM, I'm going to merge since this change can't break anything.

shurcooL:
I've confirmed that the issue is fixed now.

PR #170, opened 2016-06-27
components/Dropdown: only bind toggle event to angle icon

Unknwon:
See details: https://app.asana.com/0/87040567695724/147442522463990

cc/ @rothfels


rothfels:
LGTM

PR #171, opened 2016-06-27
Dealing with ports forwarding

alexsaveliev:
Introduced two optional environment variables:
-  `PUBLIC_WEBPACK_DEV_SERVER_URL` points to public address of webpack dev server (IP/HOST:PORT). If set, webpack server will still listen on address extracted from `WEBPACK_DEV_SERVER_URL` but will use `PUBLIC_WEBPACK_DEV_SERVER_URL` as a public one. If not set, webpack dev server will listen on `localhost:8080`
- `WEBPACK_DEV_SERVER_ADDR` webpack dev server bind IP address. If set, webpack dev server will bind to given IP, if not, default `127.0.0.1` will be used.

Default behavior is to leave them unset, however in the case when virtual box container maps public address `PUBLIC-NAME:PUBLIC-PORT` to virtual box's `LOCAL-IP:LOCAL-PORT` we may set:
- `PUBLIC_WEBPACK_DEV_SERVER_URL` to `PUBLIC-NAME:PUBLIC-PORT`
- `WEBPACK_DEV_SERVER_ADDR` to `0.0.0.0`

##### Reviewer tasks

@sqs Could you please run the app with my changes but without extra environment variables set  (default developer's config where server is running on `localhost:3080` with webpack dev server on `localhost:8080` ? I'd like to make sure I didn't break things.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,8 @@ app-dep:
 	cd app && $(NPM_RUN_DEP)
 
 WEBPACK_DEV_SERVER_URL ?= http://localhost:8080
+PUBLIC_WEBPACK_DEV_SERVER_URL ?= $(WEBPACK_DEV_SERVER_URL)
+WEBPACK_DEV_SERVER_ADDR ?= 127.0.0.1
 SERVEFLAGS ?=
 serve-dev: serve-dep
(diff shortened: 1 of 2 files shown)


sqs:
LGTM. I cherrypicked into the base branch.

PR #172, opened 2016-06-27
read .sg-drone.yml instead of .drone.yml

neelance:
Sourcegraph is currently using drone to run its builds. While this enables some reuse of drone functionality during the Sourcegraph build, our environment is still different to the normal drone environment. Using the same config file leads to conflicts that can not be easily resolved. This commit makes Sourcegraph use the .sg-drone.yml file instead, so the project maintainer can decide with configuration to share between the two environments and which parts to not run on Sourcegraph.
Resolves https://app.asana.com/0/91952577398025/133290561357955


sqs:
Looking at the .drone.yml file (https://sourcegraph.com/github.com/go-swagger/go-swagger/-/blob/.drone.yml), there are some things that actually are relevant to Sourcegraph that we wouldn't want to make them specify in 2 places, such as the Go versions and the `go get` commands (in case they didn't vendor everything). In general, srclib needs to run in an environment similar to that in which the project is built.

But I think we don't need to address this yet. No projects are actually overriding .drone.yml with custom stuff for Sourcegraph, so we can just eliminate the ability to use a custom .drone.yml file (i.e., always start from scratch instead of merging with the existing .drone.yml data). When we see the need to allow repo authors to create a .sg-drone.yml file, we can add this functionality then.

We should check that there are no repos that work BECAUSE we run their existing .drone.yml steps, though. There might be some such repos. :)

neelance:
@sqs Our main repository IS overriding the .drone.yml to make its import path work, see https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/.drone.yml#L2

In normal builds, we're automatically using `go get` to fetch dependencies, but we do not do a normal Go build of the source code. We just analyze it. I don't see how to use more than one Go version there, since we can't handle more than one srclib output. I also think using different Go versions is unnecessary because of Go's backwards compatibility guarantee.

sqs:
👍

neelance:
What does :+1: mean?

sqs:
SGTM. I will review it soon.

sqs:
LGTM.

PR #174, opened 2016-06-28
pkg/dockerutil: make ContainerHost support Docker For Mac; do not hard-code VirtualBox address

slimsag:
- Add support for `Docker for Mac` (newest Docker installer for OS X).
- Continued support for Docker Toolbar (previous Docker installer for OS X). i.e., do not hard-code the VirtualBox address.

##### Reviewer tasks

##### Test plan

- [ ] (Stephen) Manually tested and working on OS X 10.11.5 w/ Docker for Mac.
- [ ] (Reviewer) Test on OS X w/ Docker Toolbox.

I am convinced this will work with Docker Toolbox, and would be surprised if it doesn't, but we should confirm before merging. I don't have it installed, but can test if not easy for the reviewer. To test, simply `make serve-dev` and then build a repository. If the cloning step succeeded, then it works!


slimsag:
@shurcooL adding you as reviewer, if you're not the right person just lmk 😄

pkg/dockerutil/container_host.go
```diff
@@ -3,30 +3,95 @@ package dockerutil
...
+	addrs, err := iface.Addrs()
+	if err != nil {
+		return "", err
+	}
+	for _, addr := range addrs {
+		if ipn, ok := addr.(*net.IPNet); ok {
+			return ipn.IP.String(), nil
+		}
+	}
+	return "", nil
```
shurcooL: What does the pair `"", nil` mean? It differs from both `"not empty string", nil` and `"", not nil error`.

pkg/dockerutil/container_host.go
```diff
@@ -3,30 +3,95 @@ package dockerutil
...
+		//
+		//  Host does not exist: "default"
+		//
+		// (because Docker For Mac uses a VM managed entirely by Docker itself,
+		// and hence VirtualBox doesn't play a role).
+		//
+		// We use this difference to either choose the vboxnet0 interface
+		// (typically 192.168.99.1) for Docker Toolbox, or otherwise use
+		// localhost for Docker For Mac (which "just works").
+		_, err := exec.Command("docker-machine", "ip", "default").Output()
```
shurcooL: It's really strange to do `_, err := cmd.Output()`. If you want to discard output, just do `err := cmd.Run()`.

pkg/dockerutil/container_host.go
slimsag: That no IP address was found for the named interface.

slimsag: I'll add a comment describing this.

pkg/dockerutil/container_host.go
shurcooL: Ok, I see you're probably using it because `If c.Stderr was nil, Output populates ExitError.Stderr.`, but `Run` doesn't.

It'd still probably be cleaner/simpler to just do:

```Go
cmd := exec.Command
var stderr bytes.Buffer
cmd.Stderr = &stderr
...
```

But leaving it as is okay too.

pkg/dockerutil/container_host.go
```diff
@@ -3,30 +3,95 @@ package dockerutil
...
 		}
-		for _, addr := range addrs {
-			if ipn, ok := addr.(*net.IPNet); ok {
-				return ipn.IP.String(), nil
-			}
+		if ip != "" {
+			return ip, nil
 		}
+		log15.Crit("Unable to determine Docker container host address (as seen by containers). Please report this error to support@sourcegraph.com and include your OS and Docker version.", "iface", iface)
+		return "", fmt.Errorf("unable to determine Docker container host address (as seen by containers), using interface %s", iface)
```
shurcooL: I would suggest swapping these two.

[Indent the error flow](https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow), keep the normal path less indented. Meaning:

```Go
if ip == "" {
    return "", fmt.Errorf(...)
}
return ip, nil
```

pkg/dockerutil/container_host.go
shurcooL: I guess I meant to ask why you don't want to return an error when no named IP was found.

pkg/dockerutil/container_host.go
slimsag: Good point, will do here and below on line 93.

pkg/dockerutil/container_host.go
slimsag: Yeah I was doing it for the population of `exit.Stderr`. I agree with you, though, that way is better. I'll do that :)

pkg/dockerutil/container_host.go
slimsag: Hmm.. I guess we could return an error, but unless we return the more verbose error:

```Go
fmt.Errorf("unable to determine Docker container host address (as seen by containers), using interface %s", iface)
```

then I don't think it would help clarify the calling code, would it? Maybe that verbose error (+ `log.Crit`) would be ok here? what do you think?

pkg/dockerutil/container_host.go
```diff
@@ -3,30 +3,95 @@ package dockerutil
...
+		//  Host does not exist: "default"
+		//
+		// (because Docker For Mac uses a VM managed entirely by Docker itself,
+		// and hence VirtualBox doesn't play a role).
+		//
+		// We use this difference to either choose the vboxnet0 interface
+		// (typically 192.168.99.1) for Docker Toolbox, or otherwise use
+		// localhost for Docker For Mac (which "just works").
+		_, err := exec.Command("docker-machine", "ip", "default").Output()
+		if exit, ok := err.(*exec.ExitError); ok {
```
shurcooL: Do you know in what situations `err` would not successfully assert to `*exec.ExitError`? Do you care about handling them or not?

pkg/dockerutil/container_host.go
```diff
@@ -3,30 +3,95 @@ package dockerutil
...
+			lower := strings.ToLower(string(exit.Stderr))
+			if strings.Contains(lower, "host does not exist") {
+				// Likely Docker For Mac
+				return firstLocalIP4()
+			}
```
shurcooL: If the user does not have `docker-machine` binary installed on their system, is that an indicator they would probably have Docker for Mac, if anything? Or does Docker for Mac also ship with that binary?

If `docker-machine` binary doesn't exist, you don't catch that error condition.

```
$ goexec 'exec.Command("does-not-exist").Output()'
([]uint8)(nil)
(*exec.Error)(&exec.Error{
	Name: (string)("does-not-exist"),
	Err: (*errors.errorString)(&errors.errorString{
		s: (string)("executable file not found in $PATH"),
	}),
})
```

Would it be good idea to check for that?

shurcooL:
A high level question, have you considered trying to ping the addresses in order to find one that works, rather than using docker-machine command's output to guess which address will likely work?

That said, see my comments, and LGTM. I expect this will work, and don't see issues. You should merge when ready.

pkg/dockerutil/container_host.go
shurcooL: I don't think the more verbose error is appropriate.

I only asked because returning 3 type of tuples is less common than 2 types, and the function was not documented, so I couldn't tell if it was an unintended typo (i.e., you forgetting to `return "", errNotFound`) or intended.

pkg/dockerutil/container_host.go
slimsag: I don't know, but the docs for `Output` method read as:

> Output runs the command and returns its standard output. Any returned error will usually be of type *ExitError. If c.Stderr was nil, Output populates ExitError.Stderr.

This led me to believe it is unlikely to be anything else, so I doubt it's worth handling. Maybe I should reconsider though.

pkg/dockerutil/container_host.go
slimsag: I think that `docker-machine` is still around for Docker for Mac, surprisingly. After I uninstalled Docker Toolbox and installed Docker for Mac, I still have it. Also, in their official docs https://docs.docker.com/docker-for-mac/ the section `Step 2. Check versions of Docker Engine, Compose, and Machine` suggests running `docker-machine --version` to confirm it is upgraded.

I think it's used for other VM-related management, even in Docker for Mac, so a non-existent binary would imply Docker is not installed from what I gather.

shurcooL: Gotcha, then there's no need to handle/differentiate based on that.

slimsag:
Richard has beaten me to the punch (and has a much nicer implementation /cc @shurcooL ): https://github.com/sourcegraph/sourcegraph/commit/40e8c86e7d1480a3b4912502c000472b4367bee6

PR #175, opened 2016-06-28
Definfo "Loading..." Fix

ggilmore:
Fixed issue where a 404 response from the endpoint would result in the DefInfo page always displaying Loading... instead of saying "No examples found".


ggilmore:
I will squash merge once this change is accepted.

app/web_modules/sourcegraph/def/DefBackend.js
```diff
@@ -127,6 +130,13 @@ const DefBackend = {
 	},
 };
 
+function filterNotFound(data) {
+	if (get(data, "Error.response.status") === 404) {
```
rothfels: What is the benefit of doing this over `if (data && data.Error && data.Error.response && data.Error.response.status === 404)`?

I had to look up what that lodash function was doing.

rothfels: Also, I think instead of filtering the error like this, you're better off just having a helper function like `is404`, and if there's a 404, to dispatch the empty object to `ExamplesFetched`.

ggilmore: Discussed offline.

PR #178, opened 2016-06-28
WIP - User settings page with repos list

sqs:
Adds a basic structure for user settings, and adds the first user settings page: the repo list. This repo list replaces the previous `/repositories` page with a bare-bones, unstyled table. The table UI will support enabling/disabling repos (and checking the hook/build status of repos) for the improved search experience.

@shurcooL @rothfels: This adds the basic structure for a user settings area. You might find it useful.

image


rothfels:
@sqs thanks!

sqs:
@chexee I am going to merge this so there is no overhang. We will need something like this for the search stuff, but this is not the highest priority page, so I don't think it's worth your time to look into this page right now. :smile:

PR #179, opened 2016-06-28
fix JS error when clicking on search results

neelance:
See https://github.com/sourcegraph/sourcegraph-issues/issues/39


beyang:
Lgtm

PR #180, opened 2016-06-28
srclib: Update docker image of all srclib toolchains

keegancsmith:
This updates removes the need for us to support sha256 style tags, which are no longer supported by docker. So we also update the tests to ensure we don't re-introduce them.


beyang:
LGTM thanks!

PR #181, opened 2016-06-28
repos: Use elevated perms to sync github metadata

keegancsmith:
We are missing metadata for lots of repositories in our DB thanks to only
unauthed users browsing them. Unauthed users will do a `Repos.Get`, which
internally will do a `Repos.Update` to sync the missing metadata. However, that
would fail since `Repos.Update` requires write permissions. Instead, it is safe
to use an elevated permission here.

Fixes https://app.asana.com/0/87040567695724/147734985562463

### DIFF
--- a/services/backend/repos.go
+++ b/services/backend/repos.go
@@ -100,7 +100,7 @@ func (s *repos) List(ctx context.Context, opt *sourcegraph.RepoListOptions) (*so
 }
 
 func (s *repos) setRepoFieldsFromRemote(ctx context.Context, repo *sourcegraph.Repo) error {
-	var updateOp *store.RepoUpdate
+	var updateWithDataFromRemote *store.RepoUpdate
 	repo.HTMLURL = conf.AppURL(ctx).ResolveReference(app_router.Rel.URLToRepo(repo.URI)).String()
 	// Fetch latest metadata from GitHub (we don't even try to keep
 	// our cache up to date).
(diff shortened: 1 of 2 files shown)


services/backend/repos.go
```diff
@@ -114,7 +114,12 @@ func (s *repos) setRepoFieldsFromRemote(ctx context.Context, repo *sourcegraph.R
 
 	if updateOp != nil {
 		log15.Debug("Updating repo metadata from remote", "repo", repo.URI)
-		if err := store.ReposFromContext(ctx).Update(ctx, *updateOp); err != nil {
+		// setRepoFieldsFromRemote is used in read requests, including
+		// unauthed ones. However, this write isn't as the user, but
+		// rather an optimization for us to save the data from
+		// github. As such we use an elevated context to allow the
+		// write.
+		if err := store.ReposFromContext(ctx).Update(elevatedActor(ctx), *updateOp); err != nil {
```
shurcooL: This is safe to do because, as you say, this isn't a write by the user, it's using data from a public GitHub client response.

Let's make that more clear by renaming `updateOp` to `updateWithDataFromGitHub` or so. It's an uglier name, but it communicates the security implications better IMO - that its contents are not controlled by the user. Because it's related to security, having a longer name is okay (even good) I think.

shurcooL:
One suggestion. Otherwise LGTM.

keegancsmith:
@shurcooL updated variable name to `updateWithDataFromRemote` since we may have other services in future including data here.

neelance:
Hmm, this is okay, but from a security perspective it would fit better if the store itself would have a `updateWithDataFromGitHub` operation which fetches the data from GitHub and updates the DB. This operation would fully encapsulate the safety of the operation. It can not do any harm, thus the user of that operation does not need write permissions. That way you would not need an elevated actor.

keegancsmith:
@neelance I agree, this logic belongs in the store. I'd likely follow-up with that so we can unblock all the missing metadata. Also, we should probably move to only querying github in the case of an update / refresh and for privacy checks. But for now this will help us "catch-up" all the missing metadata.

There is more than one place in our codebase where we deal with missing metadata due to the store interface not filling in missing data, so we need this area of our code is ripe for improvement.

PR #182, opened 2016-06-28
added listener for sourcegraph desktop url events

uforic:
When a user clicks on a token in their editor, we want to change the URL that the webview points to. However, by re-rendering the web view, the whole page reloads. We lose the react router goodness.

This will pass the new URL through to the application, and call router.replace.

We had to refactor App.js to be a React component, which made this tougher that it would have been.

##### Reviewer tasks

- [x] METRICS: reviewer verified correct usage metric tracking was added

##### Test plan

Will upload to a staging box, and verify that normal behavior isn't broken. Will then connect Sourcegraph Desktop to that staging service, and verify that it behaves as expected.

@KattMingMing - Let's think about logging in all of this. I suspect it might not be done correctly until we make some changes.

@anyone - would love to know if this is a security vulnerability in some way, now that we are calling router.replace for a URL received.


app/web_modules/sourcegraph/app/App.js
```diff
@@ -23,22 +24,49 @@ const reactElement = React.PropTypes.oneOfType([
...
+			document.addEventListener("sourcegraph:desktop", this._handleSourcegraphDesktop);
+		}
+	}
+
+	componentWillUnmount() {
+		document.removeEventListener("sourcegraph:desktop", this._handleSourcegraphDesktop);
+	}
+
+	_handleSourcegraphDesktop(e) {
+		console.log(e);
```
rothfels: Kill

app/web_modules/sourcegraph/app/App.js
```diff
@@ -1,6 +1,7 @@
 // @flow
 
 import React from "react";
+import Component from "sourcegraph/Component";
```
rothfels: Don't use this.

app/web_modules/sourcegraph/app/App.js
```diff
@@ -23,22 +24,49 @@ const reactElement = React.PropTypes.oneOfType([
 	React.PropTypes.element,
 ]);
 
-function App(props, {signedIn}) {
-	let styleName = "main-container";
-	if (props.location.state && props.location.state.modal) styleName = "main-container-with-modal";
-	if (!signedIn && location.pathname === "/") styleName = "main-container-homepage";
+class App extends Component {
```
rothfels: React.Component

KattMingMing:
LGTM ++ 
I'll send over some info about metrics for this

uforic:
Talked to Kingy, metrics conversation to be had later, a lot of considerations to log the stuff we want.

uforic:
Tested on staging3, looking good.

PR #185, opened 2016-06-29
Remove global defs

beyang:
This removes the `GlobalDefs` store and adds instrumentation to its replacement, the `Defs` store.

After this is merged, we should be able to drop the `global_defs` table.


services/backend/internal/localstore/defs.go
```diff
@@ -402,11 +402,11 @@ func (s *defs) UpdateFromSrclibStore(ctx context.Context, op store.DefUpdateOp)
 	if err := accesscontrol.VerifyUserHasWriteAccess(ctx, "Defs.UpdateFromSrclibStore", op.Repo); err != nil {
 		return err
 	}
-
 	repo, err := store.ReposFromContext(ctx).Get(ctx, op.Repo)
 	if err != nil {
 		return err
 	}
+	__observer := newDefsUpdateObserver("defs2", repo.URI)
```
keegancsmith: why the funky name? (`__` prefix)

services/backend/internal/localstore/defs.go
```diff
@@ -439,6 +441,10 @@ func (s *defs) Update(ctx context.Context, op store.DefUpdateOp) error {
 		return err
 	}
 
+	__observer := newDefsUpdateObserver("defs2", repo.URI)
+	__observer.start("update_total")
```
keegancsmith: this api is pretty tricky. I much smoother to use API would just be getting start to return a `func()` which records the end time. That way you don't have to remember to use the same string / prevents typos causing runtime errors.

services/backend/internal/localstore/defs.go
keegancsmith: also there is nothing measuring the total time for `UpdateFromSrclibStore`

services/backend/internal/localstore/defs.go
```diff
@@ -467,17 +473,22 @@ func (s *defs) Update(ctx context.Context, op store.DefUpdateOp) error {
 	dbh := graphDBH(ctx)
 
 	// Update def_keys
+	__observer.start("update def_keys")
 	if err := updateDefKeys(ctx, dbh, repo.URI, chosenDefs); err != nil {
+		__observer.end("update def_keys")
 		return err
 	}
+	__observer.end("update def_keys")
```
keegancsmith: nit: I don't think it is enforced, but generally prometheus labels I have seen don't contains whitespace

services/backend/internal/localstore/defs_helpers.go
```diff
@@ -1,44 +1,74 @@
...
+		Name:      "search_duration_seconds",
+		Help:      "Duration for def search",
+		MaxAge:    time.Hour,
+	}, []string{"table", "repo", "part"})
+	prometheus.MustRegister(defsSearchDuration)
 }
 
-type defsUpdateObserver struct {
+type observer struct {
+	mu          sync.Mutex
```
keegancsmith: if you use the API I mentioned earlier of just returning a closure for end, you can avoid keeping track of the starts map and this mutex.

keegancsmith:
Some inline comments, but otherwise LGTM

keegancsmith:
@beyang I want to merge this in ASAP since it will help the async queue move faster. So I did my suggestions in https://github.com/sourcegraph/sourcegraph/pull/185/commits/b2e982e877acd3083229bb2f7cb4a3aec1f2cbf2
If that looks good to you I'll merge.

beyang:
LGTM thanks!

PR #187, opened 2016-06-29
Analytics: Set user property for permissions and remove file fetched

KattMingMing:
##### Description

- File fetched isn't providing any value now that backend tracking is implemented.
- We should be explicitly setting a user property for github permissions to reduce confusion when querying


app/web_modules/sourcegraph/user/UserActions.js
```diff
@@ -61,10 +61,12 @@ export class FetchedEmails {
 export class FetchedGitHubToken {
 	uid: number;
 	token: ExternalToken | {Error: any};
+	eventName: string;
 
 	constructor(uid: number, token: ExternalToken | {Error: any}) {
 		this.uid = uid;
 		this.token = token;
+		this.eventName = "FetchedGitHubToken";
```
uforic: Do we have a string constants file for this type of thing nowadays?

app/web_modules/sourcegraph/util/EventLogger.js
```diff
@@ -258,6 +258,12 @@ export class EventLogger {
 				this.logEventForCategory(AnalyticsConstants.CATEGORY_ENGAGEMENT, AnalyticsConstants.ACTION_SUCCESS, action.eventName);
 			}
 			break;
+		case UserActions.FetchedGitHubToken:
+			if (action.eventName && action.token) {
+				let allowedPrivateAuth = action.token.scope && action.token.scope.includes("repo") && action.token.scope.includes("read:org");
```
uforic: is eventName used here / why do we care about event name?

uforic: Also, just confirming we are not logging this - this is because it is done server side?

uforic:
LGTM - Just confirm that FetchedGitHubToken doesn't log, and that having an action eventName is necessary.

app/web_modules/sourcegraph/util/EventLogger.js
KattMingMing: We need an explicit property set on the user in amplitude land - this is not currently done. We can remove event name from this

KattMingMing:
FetchedGitHubToken won't be recorded since there is no more eventNames and it is also inside of it's own case

PR #191, opened 2016-06-29
Sourcegraph-desktop autoupdate endpoint

attfarhan:
Adds an HTTP endpoint that will be called by Sourcegraph-desktop for autoupdating.  The endpoint returns the download URL for the latest released version of the app.   

##### Reviewer tasks


services/httpapi/sourcegraph_desktop.go
```diff
@@ -0,0 +1,31 @@
...
+		return err
+	}
+
+	if latestVersion.Version == clientVersion {
+		w.WriteHeader(http.StatusNoContent)
+		return nil
+	}
+
+	url := map[string]string{
+		"url": "https://github.com/attfarhan/desktop-test/releases/download/" + latestVersion.Version + "/Sourcegraph.zip",
```
attfarhan: Note: I've hardcoded the URL format here, is this OK or should it be called from somewhere else?

services/backend/internal/middleware/outer/outer_middleware.go
```diff
@@ -29,25 +29,26 @@ import (
...
+		Meta:               wrappedMeta{ctxFunc, services},
+		MirrorRepos:        wrappedMirrorRepos{ctxFunc, services},
+		Notify:             wrappedNotify{ctxFunc, services},
+		Orgs:               wrappedOrgs{ctxFunc, services},
+		People:             wrappedPeople{ctxFunc, services},
+		RepoStatuses:       wrappedRepoStatuses{ctxFunc, services},
+		RepoTree:           wrappedRepoTree{ctxFunc, services},
+		Repos:              wrappedRepos{ctxFunc, services},
+		Search:             wrappedSearch{ctxFunc, services},
+		SourcegraphDesktop: wrappedSourcegraphDesktop{ctxFunc, services},
```
shurcooL: Have you considered dropping `Sourcegraph` from `SourcegraphDesktop`? I think it might be clear from context, so adding `Sourcegraph` as a prefix to a service name doesn't seem needed.

services/backend/sourcegraph_desktop.go
```diff
@@ -0,0 +1,29 @@
...
+	"sourcegraph.com/sourcegraph/sourcegraph/pkg/githubutil"
+)
+
+var SourcegraphDesktop sourcegraph.SourcegraphDesktopServer = &sourcegraphdesktop{}
+
+type sourcegraphdesktop struct{}
+
+var _ sourcegraph.SourcegraphDesktopServer = (*sourcegraphdesktop)(nil)
+
+func (s *sourcegraphdesktop) GetLatest(ctx context.Context, clientVersion *sourcegraph.ClientDesktopVersion) (*sourcegraph.LatestDesktopVersion, error) {
```
shurcooL: `clientVersion` parameter is unused, is that intentional? Are there future plans to use it, otherwise why is it passed here?

services/httpapi/sourcegraph_desktop.go
shurcooL: It might be fine for now, but we may want to factor it out into a `const` later.

services/httpapi/router/router.go
```diff
@@ -105,6 +106,7 @@ func New(base *mux.Router) *mux.Router {
 	repoPath := `/repos/` + routevar.Repo
 	base.Path(repoPath).Methods("GET").Name(Repo)
 
+	base.Path("/sourcegraph-desktop/").Methods("GET").Name(SourcegraphDesktop)
```
shurcooL: Just checking, but are you keeping "sourcegraph" in the name here? It's fine if you are.

api/sourcegraph/sourcegraph.proto
```diff
@@ -2255,6 +2255,21 @@ service Search {
...
+	rpc GetLatest(pbtypes.Void) returns (LatestDesktopVersion);
+}
+
+// LatestDesktopVersion is the latest version of the Sourcegraph desktop app.
+message LatestDesktopVersion {
+	string Version = 1;
+}
+
+// clientVersion is a user's currently installed version of the Sourcegraph desktop app.
+message ClientDesktopVersion {
```
shurcooL: `ClientDesktopVersion` is unused and can be removed, right?

services/backend/sourcegraph_desktop.go
```diff
@@ -0,0 +1,30 @@
...
+var Desktop sourcegraph.DesktopServer = &desktop{}
+
+type desktop struct{}
+
+var _ sourcegraph.DesktopServer = (*desktop)(nil)
+
+func (s *desktop) GetLatest(ctx context.Context, _ *pbtypes.Void) (*sourcegraph.LatestDesktopVersion, error) {
+	gh := githubutil.Default.UnauthedClient()
+
+	latestRelease, _, err := gh.Repositories.GetLatestRelease("sourcegraph", "sourcegraph-desktop")
```
shurcooL: There's no caching of results going on here as far as I can tell, which can cause high use of the GitHub API rate limit quota.

It might be within acceptable limits at first, so merging this as is would be acceptable. But you'd need to watch the usage, and if it gets very high, then spending time on adding caching here will be very justified.

Until then, I think you should ad least add a prometheus counter here, so we can see how often it's being hit.

shurcooL:
Left more comments, but after that I think it's ready to merge for me.

services/httpapi/sourcegraph_desktop.go
```diff
@@ -0,0 +1,33 @@
...
+	"sourcegraph.com/sqs/pbtypes"
+)
+
+const downloadLink = "https://github.com/sourcegraph/sourcegraph-desktop/releases/download/"
+const zips = "/Sourcegraph.zip"
+
+func serveSourcegraphDesktopUpdateURL(w http.ResponseWriter, r *http.Request) error {
+	ctx, cl := handlerutil.Client(r)
+
+	clientVersion := r.Header.Get("Sourcegraph-Version")
```
sqs: Why pass as a header and not in the querystring? Seems strictly more custom/complicated.

api/sourcegraph/sourcegraph.proto
```diff
@@ -2255,6 +2255,21 @@ service Search {
 	rpc Search(SearchOp) returns (SearchResultsList);
 }
 
+// Desktop provides the latest version of the Sourcegraph desktop app to enable auto-updating.
+service Desktop {
+	rpc GetLatest(pbtypes.Void) returns (LatestDesktopVersion);
```
sqs: Do we really need to add this to our site's API? Can't we just use an existing auto-updating service for it? What about using the Mac App Store, Windows App Store, and apt/etc. repos on Linux?

I don't think GitHub has a custom API for updating GitHub desktop...I bet they use an existing library that handles it, or they use the platform app stores to do so.

sqs:
Overall comment is: this custom logic should not have to be in our API and app. This is adding too much complexity, it seems. Can we just use an existing auto-updating solution, like the Mac/Windows App Stores and apt/rpm/etc. repos on Linux (and we don't need to prioritize Linux and Windows for smooth auto-updating initially)? Or if not that, can we use a cross-platform library that handles it for us, instead of adding custom application logic to our own code? If not that, can we have it just check a file hosted on S3 that we update with a shell script?

uforic:
Yo! Some thoughts:

The electron auto-updater that is standard is documented here:
https://medium.com/@svilen/auto-updating-apps-for-windows-and-osx-using-electron-the-complete-guide-4aa7a50b904c#.udj00852c

Long term, we want to be on the Mac Store. According to King, there isn't a Test-Flight like app for the App macOS store, so once you are out there, you are out there. 

This endpoint was going to be for macOS updating only, for the period of time in between now and when we are on the App Store for macOS. There isn't a great auto-update process for Linux, and `Windows does not require an API endpoint to check for new versions — all it needs is a file server, so you can simply drop your files in an Amazon S3 bucket.`

Once we are on the macOS App Store, we can remove this endpoint. I just figured that it could and would likely be be >2 weeks, and thus a good idea to have this endpoint in the interrim period for internal users and beta program.

Another solution would be to develop gating authentication, submit on the app store, and just not allow anyone to use it without a beta login. Or just let everyone in.

Regardless, I suspect the earliest we could get it in the App Store would be end of next week (review time of 1 day, but King mentioned that there is usually a bit of back and forth for the first submission, and we don't have our Sourcegraph organization developer account yet).

Let us know what you think of ^^

services/backend/sourcegraph_desktop.go
sqs: That API client does use a caching HTTP API client; see https://sourcegraph.com/github.com/sourcegraph/sourcegraph@e456ef0291950b1e9faeed85c33511214bebdeec/-/blob/pkg/githubutil/client.go#L45.

So we're good to go there.

sqs:
Makes sense!

PR #193, opened 2016-06-30
metrics: Measure number of running grpc requests

keegancsmith:
Also removed the heartbeat, we never use it. This should help us in future when
we have a deadlock, and also help us understand long running grpc methods
better.

### DIFF
--- a/services/backend/internal/middleware/inner/trace/trace.go
+++ b/services/backend/internal/middleware/inner/trace/trace.go
@@ -44,6 +44,7 @@ func Before(ctx context.Context, server, method string, arg interface{}) context
 		spanID = appdash.NewSpanID(parent)
 	}
 	ctx = traceutil.NewContext(ctx, spanID)
+	requestGauge.WithLabelValues(server + "." + method).Inc()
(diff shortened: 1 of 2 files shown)


services/backend/internal/middleware/inner/trace/trace.go
```diff
@@ -61,17 +62,17 @@ var requestDuration = prometheus.NewHistogramVec(prometheus.HistogramOpts{
 	Help:      "Total time spent on grpc endpoints.",
 	Buckets:   statsutil.UserLatencyBuckets,
 }, metricLabels)
-var requestHeartbeat = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
neelance: We don't need `requestHeartbeat` any more?

neelance:
LGTM if we don't need `requestHeartbeat` any more.

keegancsmith:
Mentioned in the commit message we never use it. We added it a long time ago at request from a client.

neelance:
@keegancsmith Oops, missed that. :)

PR #194, opened 2016-06-30
search: add all combinations of case words to the search DB

neelance:
This makes it possible to search for "buildvars" or "buildfunc" and get "BuildVarsFunc" as a result.

##### Reviewer tasks

I don't see a negative perf impact by this change, do you?
Should I add unit tests for `allCombinations` and/or `splitCaseWords`?


services/backend/internal/localstore/defs.go
```diff
@@ -645,16 +645,17 @@ func toTextSearchTokens(def *graph.Def) (aToks []string, bToks []string, cToks [
 	defParts := delims.Split(def.Path, -1)
 	for _, w := range defParts {
 		bToks = appendRepeated(bToks, w, 2)
+		for _, c := range allCombinations(splitCaseWords(w)) {
```
keegancsmith: I'm surprised `toTextSearchTokens` doesn't have a test case, it is very testable. Just mentioning that since you are touching this code, and it would be great to see validation of the allCombinations change.

services/backend/internal/localstore/defs.go
```diff
@@ -676,6 +677,21 @@ func splitCaseWords(w string) []string {
 	return camelcase.Split(w)
 }
 
+func allCombinations(s []string) []string {
```
keegancsmith: It isn't obvious (to me atleast) from the name what exactly this is. My first thought was powerset, which is somewhat right. It seems to be the powerset, except with each set concatted to a string in the order they appear in `s`. That isn't the best description, but it is an important property that probably deserves some explanation.

this name isn't great, at first I assumed this was something like a powerset since that is usually what you get, but it is all combinations but keeping the original ordering of strings. I can't really think of a better name right now, but maybe a comment explaining that would be great.

neelance:
Added a test for `toTextSearchTokens`, thus indirectly testing `allCombinations` and `splitCaseWords`.

keegancsmith:
LGTM.

In theory something that tokenizes into a lot of elements (which other people can control) would cause us potential issues. But I don't think it is something to worry about + we can easily add in now or later a simple cap on number of tokens.

PR #196, opened 2016-06-30
Sanitize gRPC errors with "unknown" and "internal" error codes to non-admin users

beyang:


services/backend/internal/middleware/inner/gen_middleware.go
```diff
@@ -66,8 +67,15 @@ func Services() svc.Services {
...
-				err = grpc.Errorf(codes.Internal, "<<<$service.Name>>>.<<<.Name>>> returned nil, nil")
+			if err != nil {
+				actor := auth.ActorFromContext(ctx)
+				code := grpc.Code(err)
+				if !actor.Admin && (code == codes.Unknown || code == codes.Internal) {
+					// Sanitize, because these errors should not be user visible.
+					err = grpc.Errorf(code, "<<<$service.Name>>>.<<<.Name>>> failed with internal error. Please report this to Sourcegraph support.")
+				} else if res == nil && err == nil {
+					err = grpc.Errorf(codes.Internal, "<<<$service.Name>>>.<<<.Name>>> returned nil, nil")
+				}
```
keegancsmith: can you put this in a helper function (like we do for trace.Before and trace.After)? Easier to read + less duplication.

services/backend/internal/middleware/inner/gen_middleware.go
```diff
@@ -66,8 +67,15 @@ func Services() svc.Services {
...
 			res, err = backend.Services.<<<$service.Name>>>.<<<.Name>>>(ctx, param)
-			if res == nil && err == nil {
-				err = grpc.Errorf(codes.Internal, "<<<$service.Name>>>.<<<.Name>>> returned nil, nil")
+			if err != nil {
+				actor := auth.ActorFromContext(ctx)
+				code := grpc.Code(err)
+				if !actor.Admin && (code == codes.Unknown || code == codes.Internal) {
+					// Sanitize, because these errors should not be user visible.
+					err = grpc.Errorf(code, "<<<$service.Name>>>.<<<.Name>>> failed with internal error. Please report this to Sourcegraph support.")
+				} else if res == nil && err == nil {
```
keegancsmith: this else if is in the wrong place. At this point you never have `err == nil`.

services/backend/internal/middleware/inner/gen_middleware.go
shurcooL: I'm not sure if this is the best place. We already do similar sanitization in a few other places. Grep for `handlerutil.DebugMode`.

Also, it's better to follow similar logic as `handlerutil.DebugMode` (except maybe you don't pass `*http.Request` as input), and also check for `os.Getenv("DEBUG")` env var.

Final thought, outside of scope, but `os.Getenv("DEBUG")` can be factored out of `DebugMode` into a private top level variable in `handlerutil`. `DebugMode` can be called often, and there's no need to call `os.Getenv("DEBUG")` per request; once on process startup is enough.

keegancsmith:
One inline comment requires a fix, the other is optional (optional is helper function). Ping me when you update and I'll take a quick look again.

beyang:
@keegancsmith PTAL

beyang:
@keegancsmith follow-up issue here: https://app.asana.com/0/87040567695724/150412207808897

app/internal/tmpl/tmpl.go
```diff
@@ -195,7 +194,7 @@ func Exec(req *http.Request, resp http.ResponseWriter, name string, status int,
 
 			CurrentSpanID:    traceutil.SpanID(req),
 			CurrentRouteVars: mux.Vars(req),
-			Debug:            handlerutil.DebugMode(req),
+			Debug:            auth.DebugMode(ctx),
```
keegancsmith: `auth` seems like a strange place for this function to live

keegancsmith:
LGTM

pkg/auth/context.go
```diff
@@ -16,3 +19,13 @@ func ActorFromContext(ctx context.Context) Actor {
 func WithActor(ctx context.Context, a Actor) context.Context {
 	return context.WithValue(ctx, actorKey, a)
 }
+
+func DebugMode(ctx context.Context) bool {
```
shurcooL: `DebugMode` was previously documented, but it's not here. Why not?

I think it's very important to document it because misusing this function (or a bug introduced inside it) can lead to security issues.

pkg/conf/env/env.go
```diff
@@ -9,8 +9,15 @@ import (
 	"os/user"
 	"path/filepath"
 	"runtime"
+	"strconv"
 )
 
+var Debug bool
```
shurcooL: Does it really need to be exported? There's only one place that uses it. Why not just place it in the same package as `DebugMode` and make it unexported.

services/httpapi/httpapi.go
```diff
@@ -172,7 +172,7 @@ func handleError(w http.ResponseWriter, r *http.Request, status int, err error)
 	errBody := err.Error()
 
 	var displayErrBody string
-	if handlerutil.DebugMode(r) {
+	if handlerutil.DebugMode {
 		// Only display error message to admins or locally, since it
```
shurcooL: This comment says "// Only display error message to admins or locally", is it still true?

app/internal/oauth2client/github.go
```diff
@@ -143,8 +143,10 @@ func serveGitHubOAuth2Receive(w http.ResponseWriter, r *http.Request) (err error
 	}()
 
 	ctx, cl := handlerutil.Client(r)
-
-	actor := auth.ActorFromContext(ctx)
+	authInfo, err := cl.Auth.Identify(ctx, &pbtypes.Void{})
```
beyang: @sqs can you give this a lookover? An actor is no longer added to the context object in the app layer, so this changes it to call the `cl.Auth.Identify` endpoint to get the current user.

dev/check-imports
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
beyang: @sqs Moving the entire `auth` package into `services/backend/internal` would be a much bigger change and I think we have higher priorities at the moment. This adds a static check that should prevent any dependencies from `httpapi` -> `auth` or `app` -> `auth`. The `auth` package itself should at some point be split up. The actor logic should be moved into `services/backend/internal`, but I don't want to spend any more time on this at the moment.

dev/check-imports
```diff
@@ -0,0 +1,18 @@
...
+		illegal_deps=$(go list -f '{{join .Deps "\n"}}' "$pkg" | grep -x 'sourcegraph.com/sourcegraph/sourcegraph/pkg/auth')
+		if [[ ! -z "$illegal_deps" ]]; then
+				echo "-- Found $pkg illegally depends on";
+				echo "$illegal_deps";
+				found_violation="true";
+		fi
+done;
+
+if [[ ! -z "$found_violation" ]]; then
+		exit 1;
```
keegancsmith: did you test that this actually fires if you add a dep on auth?

dev/check-imports
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# Check that the app and httpapi packages do not depend on the auth package. The
+# app and httpapi layers should be agnostic to the identity of the actor.
+found_violation=
+declare -a no_auth_pkgs=("sourcegraph.com/sourcegraph/sourcegraph/services/httpapi/..." "sourcegraph.com/sourcegraph/sourcegraph/app/...")
+for pkg in "${no_auth_pkgs[@]}"; do
+		illegal_deps=$(go list -f '{{join .Deps "\n"}}' "$pkg" | grep -x 'sourcegraph.com/sourcegraph/sourcegraph/pkg/auth')
```
keegancsmith: maybe add a check that `pkg/auth` still exists, to prevent a refactor in future making this check out of date

app/templates/scripts.html
```diff
@@ -13,10 +13,10 @@
 		{{if deployedGitCommitID}}
 			sentryTags["Deployed commit"] = "{{deployedGitCommitID}}";
 		{{end}}
-		{{if .Actor.IsAuthenticated}}
-			sentryTags["Authed"] = "{{.Actor.Login}}" ? "yes" : "no";
-			sentryTags["Authed UID"] = "{{.Actor.UID}}";
-			sentryTags["Authed user"] = "{{.Actor.Login}}";
+		{{if .AuthInfo.UID}}
```
keegancsmith: `IsAuthenticated` was clearer

app/templates/scripts.html
```diff
@@ -13,10 +13,10 @@
...
+			sentryTags["Authed"] = "{{.AuthInfo.Login}}" ? "yes" : "no";
```
keegancsmith: somewhat unrelated: why would this ever be no?

pkg/eventsutil/events.go
```diff
@@ -38,7 +36,6 @@ func LogEvent(ctx context.Context, event string, eventProperties map[string]stri
 
 	Log(&sourcegraph.Event{
 		Type:            fmt.Sprintf("Server%s", event),
-		UserID:          login,
```
keegancsmith: surely this was important information? Or is this event information we no longer actually consume (ie the whole/part of the pkg should be deleted?)

pkg/handlerutil/debug.go
```diff
@@ -1,22 +1,15 @@
...
-		return true
-	}
-	return false
+// DebugMode is true if and only if the application is running in debug mode. In
+// this mode, the application should display more verbose and informative errors
+// in the UI. Debug should NEVER be true in production.
+var DebugMode bool
+
+func init() {
+	DebugMode, _ = strconv.ParseBool(os.Getenv("DEBUG"))
```
keegancsmith: unrelated but this seems to only be used by the UI part of the code. A less generic name is likely better to prevent us accidentally setting it in production.

services/backend/internal/middleware/inner/gen_middleware.go
```diff
@@ -69,6 +69,12 @@ func Services() svc.Services {
 			if res == nil && err == nil {
 				err = grpc.Errorf(codes.Internal, "<<<$service.Name>>>.<<<.Name>>> returned nil, nil")
 			}
+			if err != nil && !DebugMode(ctx) {
+				if code := grpc.Code(err); (code == codes.Unknown || code == codes.Internal) {
+					// Sanitize, because these errors should not be user visible.
+					err = grpc.Errorf(code, "<<<$service.Name>>>.<<<.Name>>> failed with internal error.")
```
keegancsmith: should we log this so we can at least know about it? Or will that just be log spam

services/backend/internal/middleware/inner/gen_middleware.go
```diff
@@ -69,6 +69,12 @@ func Services() svc.Services {
 			if res == nil && err == nil {
 				err = grpc.Errorf(codes.Internal, "<<<$service.Name>>>.<<<.Name>>> returned nil, nil")
 			}
+			if err != nil && !DebugMode(ctx) {
+				if code := grpc.Code(err); (code == codes.Unknown || code == codes.Internal) {
+					// Sanitize, because these errors should not be user visible.
```
keegancsmith: nit: template comment rather

services/httpapi/repos.go
```diff
@@ -117,17 +116,6 @@ func serveRepos(w http.ResponseWriter, r *http.Request) error {
...
-	// The only locally hosted repos are sourcegraph repos. We want
-	// to prevent these repos showing up on a users homepage, unless they
-	// are Sourcegraph staff. Only Sourcegraph staff have write
-	// access. This means that only we will see these repos on our
-	// dashboard, which is the purpose of this if-statement. When we have
-	// a fuller security model or user-selectable repo lists, we can
-	// remove this.
-	if !authpkg.ActorFromContext(ctx).HasWriteAccess() {
-		return writeJSON(w, &sourcegraph.RepoList{})
-	}
```
keegancsmith: have you browsed around the site? I'm not exactly sure off-hand where we expose the list of repos anymore (do we?), but this is quite a major change to non-admins. I assume you can likely just delete this whole route.

rothfels:
@beyang ping, can we merge this?

PR #197, opened 2016-06-30
ci: Remove verbose flag from go test

keegancsmith:
If we have a failure it is hard to find out what it is.
Some tests also include a lot of verbose output if `-v` is specified. If we need
it, we can run the test locally.

### DIFF
--- a/dev/circle-ci-run-tests.sh
+++ b/dev/circle-ci-run-tests.sh
@@ -33,7 +33,7 @@ for pkg in $(go list -f '{{ if or (gt (len .TestGoFiles) 0) (gt (len .XTestGoFil
 		then
 			echo "Run test with coverage for package: $pkg"
 			go install -race ./cmd/src
-			go test -race -v -timeout 5m -coverprofile=/tmp/cover.$i.out "$pkg" | tee /tmp/go-test.out
+			go test -race -timeout 5m -coverprofile=/tmp/cover.$i.out "$pkg" | tee /tmp/go-test.out
 			go-junit-report < /tmp/go-test.out >> $CIRCLE_TEST_REPORTS/junit/go-test.xml
 			covered+=("$pkg")
 		else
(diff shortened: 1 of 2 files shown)


renfredxh:
I'm somewhat in support for this, couple caveats though:

- some of the tests are difficult to run locally
- it will be more difficult to diagnose flaky tests this way (why did it fail on CI but not locally?).

keegancsmith:
Cool! A response to that:

* What test is difficult to run locally? If you can run make serve-dev, you should be able to run the test
* Have we ever diagnosed a flaky test before using the CI output?

Note that if a test does fail, `go test` does include the `t.Log` output. The only thing that will be missing is any special output that is added if a test checks `testing.Verbose()`

shurcooL:
My preference on this is as follows:

- When tests pass, I don't care about looking at their output.
- When tests fail, I want to see what failed and get as much information as possible. I want the `-v` verbosity. Why would I want to know less information?

I think the problem is that it's not easy to find _which_ test failed, and that's the problem that should be resolved.

shurcooL:
> Note that if a test does fail, `go test` does include the `t.Log` output. The only thing that will be missing is any special output that is added if a test checks `testing.Verbose()`

Hmm, if that's the case, that's reassuring. I thought a lot more of the output would be missing without `-v`.

keegancsmith:
There is only one test in our codebase where we check `testing.Verbose()`, and that test has the noisiest output.

I can make it always `t.Log` that noisy output, then literally we would only get the output if the test fails. Will push a commit for that.

renfredxh:
@keegancsmith 

- I remember having some annoying experiences with some lg tests in the past, not sure if this is still the case. We can take this on a case-by-case basis.
- Yes, I have diagnosed a flaky test by the output in CI. It's easy to see if the test failed due to an HTTP timeout or something that way.

keegancsmith:
@renfredxh yeah that was back when we had test tags + didn't require postgresql to do dev. Now running lg tests is super easy :)

I think this is safe to merge now, since we will get the same output as before, but now it will only show in the case of a failure. If anyone objects, feel free to revert.

PR #198, opened 2016-07-01
Add grpc-gateway.

shurcooL:
This is a change that adds grpc-gateway support. This will be needed to try out grpc-gateway for exposing some gRPC endpoints via the HTTP reverse proxy more easily than what is currently possible with the hand-written HTTP API layer.

See https://sourcegraph.quip.com/yatzAJGOzCy7 (internal document) for background and motivation.

This is a largely additive change that will be easy to remove if the grpc-gateway experiment proves to be unfit for our needs or unhelpful.

The change is best reviewed in separate commits It:

- Vendors grpc-gateway and its dependencies, purely additive.
- Updates protobuf repo to latest version, since there are additions that are required.
- Updates generation code to also perform grpc-gateway generation of .pb.gw.go files (currently not generated because there are no uses).
- Creates the grpc-gateway router.
    - It is not yet hooked up, that will follow up.

##### Reviewer tasks

##### Test plan

Tested locally (by using the site, and by running tests). Waiting on CI to pass.

This change is largely additive so far.


shurcooL:
I've discussed this with @sqs and we're going to make an early no decision on this experiment.

It doesn't appear that anyone would actually use or benefit from this. The key limitation is that grpc-gateway generates HTTP endpoints that directly map to gRPC endpoints without any custom flexibility, compared to what have when you write the HTTP endpoint manually (where you can have full control over status codes returned, headers, caching behavior, etc.).

Closing this for that reason.

PR #200, opened 2016-07-01
Garbage collect rows for old repositories when updating defs2 table

beyang:


services/backend/internal/localstore/defs.go
```diff
@@ -621,6 +621,21 @@ func (s *defs) Update(ctx context.Context, op store.DefUpdateOp) error {
...
 	}
+
+	// Garbage-collect old def and repo_revs rows
+	rrOld, err := getRepoRevsOld(ctx, dbh, repo.URI)
+	if err != nil {
+		return err
+	}
+	for _, rr := range rrOld {
+		if _, err := dbh.Exec(`DELETE FROM repo_revs WHERE id=$1 and state=2`, rr.ID); err != nil {
+			return err
```
shurcooL: If there is an error deleting one of the old defs, do you really want to make it a fatal error for the entire `Update` and not even try to remove the other old defs? Would it be better to log it and continue anyway?

neelance: I think the `repo_revs` entry should be deleted **after** the `defs2` table has been cleaned up. That way if the cleanup fails it will be tried again the next time the repo gets updated.

keegancsmith: we are also missing instrumentation on the time this takes here.

beyang: @keegancsmith will follow up on that shortly.

keegancsmith:
@beyang any updates?

beyang:
@keegancsmith no pending updates. If you want to address this as part of your instrumentation work, that'd be great. Otherwise I can revisit this later in the week.

PR #201, opened 2016-07-01
async: Block until job is complete for RefreshIndexes

keegancsmith:
RefreshIndexes is fast now, so we want to only show a build as complete once we
have refreshed its indexes. However, in future keeping the async interface is
useful for scaling. So this commit adds a simple boolean to toggle the
behaviour.


keegancsmith:
@beyang I'm not actually sure this is the best idea, even at our current scale. For example refreshindexes on golang/go takes 13minutes for a random job I was recently looking at. Do we really want a builder doing nothing for 13min each time we index a golang/go commit (which is fairly regularly). I can dive deeper into the data to find out how many minutes this would add a day to builders.

keegancsmith:
More evidence that we should move the stuff that the new indexing that has been added into srclib-import should be also put into the async queue. https://sourcegraph.com/github.com/kubernetes/kubernetes/-/builds/404

keegancsmith:
I think we should be moving in the direction of more stuff happening asynchronously. So abandoning.

beyang:
@keegancsmith how can we guarantee that the async jobs complete in a reasonable amount of time so that we don't have, for instance, linked repositories without usage examples?

shurcooL:
Just an observation, I think that changing whether the jobs run asynchronously or synchronously will not help the jobs complete faster or more reliably. It will help with visibility into that, though.

If it's a problem that they take too long, or are too unreliable when async, then making them sync will not actually improve the end result, which is that we want usage examples to appear relatively quickly. Instead, it'll just make the entire build take longer/fail more often. That is good for visibility into the problem, but it doesn't really help resolve the core issue.

I agree with @keegancsmith that more stuff being async is the better general direction, because it can be more efficient (less waiting/blocking), but we need those jobs to be quick/reliable for it to be viable and problem-free.

So going sync may be worth it in short term to help with the visibility problem, but I just want to point out it won't really resolve the perf/reliability problem. For that, we should find the bottlenecks and optimize.

keegancsmith:
@shurcooL 

I agree that we still have an underlying issue in that indexing is slower or not as reliable as we like. But...

> I think that changing whether the jobs run asynchronously or synchronously will not help the jobs complete faster or more reliably.

That is in fact false! :P Synchronously introduces an HTTP client connection failures (network, etc) and deploy interruptions into the mix. Add in the fact that we do retries in the HTTP client, but no cancellations in our server (even in the case we could notice the client disappearing) leading to multiple uploads for a single gce bucket -> 5xx errors from our storage.

Async gives us the ability to monitor and understand what in the pipeline is the issue, and react accordingly. It also allows us to add more sensible retry logic, dedupping, etc. And to me the biggest benefit is we don't have something with considerable resources (a worker) just sitting arounding essentially doing nothing (other than waiting for an HTTP request to end).

To me the only current downside in the approach is we show no indication of the other jobs we do on the builds page (but that is completely fixable). Otherwise every other system we have only actually cares that the actual srclib import has happened, but that is the quickest part of all the extra stuff we now do in that handler.

PR #205, opened 2016-07-01
app: fix fatal flaw in Beta signup page (languages and editors fields flipped)

slimsag:
Prior to this change the Language and Editor fields were accidentally swapped (only for the `/beta` page, which AFAIK nobody has yet signed up through /cc @uforic).

##### Reviewer tasks

Not much, mostly sent for posterity.

##### Test plan

Manually tested and confirmed in Mailchimp:

image


uforic:
Phewph, close call on this one. Bullet dodged, pretty much feel like this guy:
https://www.youtube.com/watch?v=ujf50wsFFc8

Separately - looks good to me!

slimsag:
😮 That's crazy!

PR #207, opened 2016-07-01
srclib-python: install python2.7 in Docker

Unknwon:
The change is not ideal, someone please help make some comments or propose changes to srclib-python/Makefile.


services/worker/dockerfiles/Dockerfile.srclib-python
```diff
@@ -20,6 +20,10 @@ RUN mkdir -p $SRCLIBPATH/sourcegraph.com/sourcegraph && mv /toolchain/t $SRCLIBP
 WORKDIR $SRCLIBPATH/sourcegraph.com/sourcegraph/srclib-python
 RUN make install
 
+RUN virtualenv -p python2.7 .env \
```
beyang: How about doing it in the Makefile like this instead: https://github.com/sourcegraph/srclib-python/commit/ea16c866f16583a04821bd2355e0c194434cc29b

Unknwon: The problem is we need to install once for Python3 and another one for Python2, if we set `.env` for them both time, only the last one is installed (Python3 in this case), vice versa.

That's why we need to change `.env` only after one version is installed.

beyang:
@Unknwon see https://github.com/sourcegraph/sourcegraph/commit/d76af33730606b23e4fb20c051e12b1d20f88e2b. This is the only change necessary now that the srclib-python Makefile installs both versions of Python.

beyang:
Lgtm

PR #208, opened 2016-07-01
added password complexity rules to signup page and reset page

ggilmore:


app/web_modules/sourcegraph/util/constants/PasswordConstants.js
```diff
@@ -0,0 +1,11 @@
+// Password Constants
+
+// PASSWORD_PATTERN is the regex that represents the allowed passwords for the join and reset password pages.
+export const PASSWORD_PATTERN = "^(?!.*(.)\\1{2})(?=.*[^A-Za-z\\d])(?=.*\\d)(?=.*[a-z])(?=.*[A-Z]).{10,128}$";
+
+// PASSWORD_DESCRIPTION is the "english" description of the password requirements enforced by PASSWORD_PATTERN.
+export const PASSWORD_DESCRIPTION = `Your password should be between 10 and 128 characters (with no more than two identical characters in a row), and should include at least:
```
rothfels: Personally, I've never been fully bought into the "no >2 duplicate characters" rule, but so be it. 

Can we please write some tests for this?

app/web_modules/sourcegraph/user/Signup.js
```diff
@@ -118,7 +119,9 @@ export class SignupForm extends Container {
 						domRef={(e) => this._passwordInput = e}
 						tabIndex="4"
 						block={true}
-						required={true} />
+						required={true}
+						pattern = {PASSWORD_PATTERN}
```
rothfels: For consistency, remove whitespace around the `=` (and on next line).

app/web_modules/sourcegraph/util/constants/PasswordConstants.js
keegancsmith: Why do we have an upper limit? I suppose it is big enough to prevent people doing crazy lengths.

Why do we enforce complexity requirements? Do other dev tools sites do that? This could be something which prevents people signing up due to them just being super lazy.

app/web_modules/sourcegraph/util/constants/PasswordConstants.js
```diff
@@ -0,0 +1,11 @@
+// Password Constants
+
+// PASSWORD_PATTERN is the regex that represents the allowed passwords for the join and reset password pages.
+export const PASSWORD_PATTERN = "^(?!.*(.)\\1{2})(?=.*[^A-Za-z\\d])(?=.*\\d)(?=.*[a-z])(?=.*[A-Z]).{10,128}$";
```
keegancsmith: given we just hash it on the backend, why bother with this?

app/web_modules/sourcegraph/util/constants/PasswordConstants.js
```diff
@@ -0,0 +1,11 @@
...
+
+// PASSWORD_DESCRIPTION is the "english" description of the password requirements enforced by PASSWORD_PATTERN.
+export const PASSWORD_DESCRIPTION = `Your password should be between 10 and 128 characters (with no more than two identical characters in a row), and should include at least: 
+- 1 uppercase character 
+- 1 lowercase character
+- 1 digit
+- 1 special character`;
```
shurcooL: Yeah, I agree this isn't the best UI, especially for a developer-oriented site.

You can have a very secure password that breaks some rule, which can be very frustrating. You can also have a very insecure password that doesn't break any rules.

I think a better strategy is to offer hints or guidance on how to come up with a password, but not create hard restrictions since they can't really guarantee anything (except possible frustration for users). It's also potentially helpful to see a "gauge bar" showing how secure your current password appears to be (but estimating that is very hard... you can do it mostly based on password length).

Just wanted to leave some thoughts. Is there a design document or issue describing this feature?

app/web_modules/sourcegraph/util/constants/PasswordConstants.js
samlavery: I think github can afford to have less friction around password strength because they have a 2FA option available.  These are just some slight reshuffled best practices to get users to enter reasonable passwords.  I too have no idea why people have been surfacing huge max char limits, but my bank recently told me they have a 200 char password limit.

PR #210, opened 2016-07-02
beta: finish migrating the frontend + backend (remove dead code)

slimsag:
- Replace all usages of `InterestForm` with the new `BetaInterestForm`.
- Remove `EmailSubscription` endpoint (new endpoint is `BetaSubscription`, so the old one is unused).
- Always persist the Beta registration form in local storage.
- See screenshots in the two `app:` commit messages.

##### Test plan

Tested locally / manually.


shurcooL:
I've looked it over and did not spot any issues, so LGTM. But I'm not very familiar with the context.

slimsag:
Thanks for the review @shurcooL! I'm gonna wait to merge since Quinn or Chelsea may have design feedback, and merging would mean this would go live. Also, I've pushed one more commit that handles persistence of the form (which was my last TODO).

slimsag:
Rebased against master.

services/httpapi/user.go
```diff
@@ -125,28 +125,3 @@ func serveBetaSubscription(w http.ResponseWriter, r *http.Request) error {
...
-	if newErr != nil {
-		return newErr
-	}
-
-	if mailchimpKey == "" {
-		return errors.New("mailchimp authorization key only available on production environments")
-	}
-
-	client := &http.Client{}
-	req, err := http.NewRequest("POST", "https://us8.api.mailchimp.com/3.0/lists/dd6c4706a1/members", bytes.NewReader(newBody))
```
rothfels: I notice that the new betaSubscription endpoint uses the same mailchimp list. Is that intentional, and have you verified that the new data fits into the old data schema?

app/web_modules/sourcegraph/components/CheckboxList.js
```diff
@@ -24,11 +25,18 @@ class CheckboxList extends React.Component {
 		return selected;
 	}
 
+	_isDefaultValue(s: string): bool {
+		for (let d of this.props.defaultValues) {
```
rothfels: What if `this.props.defaultValues` is undefined?

Simpler: replace the function body with `this.props.defaultValues && this.props.defaultValues.indexOf(s) !== -1`

app/web_modules/sourcegraph/home/BetaInterestForm.js
```diff
@@ -49,6 +62,17 @@ class BetaInterestForm extends React.Component {
 		}
 	}
 
+	_onChange: OnChangeListener;
+	_onChange() {
+		window.localStorage["beta-interest-form"] = JSON.stringify({
```
rothfels: What's the reason to keep this in local storage rather than in memory (as React component state)? The latter is more idiomatic, and local storage doesn't buy us anything unless we expect the user to pull open the modal, start editing, leave, and come back (which I find exceedingly unlikely / not a use case I think is important to support).

rothfels: The code would be a little simpler if you just used react component state.

app/web_modules/sourcegraph/home/BetaInterestForm.js
```diff
@@ -8,18 +8,28 @@ import Dispatcher from "sourcegraph/Dispatcher";
...
+import {langName} from "sourcegraph/Language";
 import * as AnalyticsConstants from "sourcegraph/util/constants/AnalyticsConstants";
 
+type OnChangeListener = () => void;
+
 class BetaInterestForm extends React.Component {
 
 	static propTypes = {
 		onSubmit: React.PropTypes.func,
+		formClass: React.PropTypes.string,
```
rothfels: `s/formClass/className` (which is the canonical way to pass styles)

rothfels:
@slimsag I checked the design, nothing blocking this from getting into production. But left a few comments, please resolve those before merging.

services/httpapi/user.go
slimsag: Yes, intentional. The schema matches fine and I have a quip document to describe all of this which I'll release later.

app/web_modules/sourcegraph/home/BetaInterestForm.js
slimsag: Use of local storage here is intentional. We expect users to come back at a later time (after submitting the form once) and update their preferences (which languages / editors they are interested in), this is all in the quip doc/spec. I don't think local storage here complicates things very much, and it can be quite useful for returning users IMO. I think the benefits outweigh the cost here.

rothfels: SGTM

services/httpapi/user.go
rothfels: 👍

PR #211, opened 2016-07-04
don't run e2e tests in parallel because it causes errors with Selenium

neelance:
This makes all those errors in the log of the selenium-server container go away.


shurcooL:
LGTM.

PR #214, opened 2016-07-05
remove Repos.ListRemote, consolidate with Repos.List

sqs:
The dual APIs for listing repos was adding complexity to any client
that needed to know if a remote repo already existed locally. The
combining of the 2 lists should occur on the server side, not be
pushed to the client, to centralize the merging logic and reduce the
amount of data that needs to be transferred. 

See individual commit messages.

On https://staging.sourcegraph.com


sqs:
@rothfels This branch (rebased against master recently, merge conflicts are probably minimal) is what will make it easier for us to enable/disable repos and scope searches. It removes the additional complexity of Repos.ListRemote and makes listing repos that meet certain characteristics (e.g., enabled repos only) possible.

PR #215, opened 2016-07-05
WIP: support enabling/disabling repos

sqs:
(determines whether they are included in search results)


api/sourcegraph/sourcegraph.proto
```diff
@@ -177,6 +177,26 @@ message Repo {
...
+	RepoPermissions Permissions = 21;
+}
+
+// RepoPermissions describes the actions that a user may perform on a
+// repo. Currently, the definition of these permissions directly maps
+// to GitHub permissions, except for "Pull", which means read access.
+message RepoPermissions {
+	bool Pull = 1;
+	bool Push = 2;
+	bool Admin = 3;
```
shurcooL: Why name these push/pull instead of read/write? The latter seems simpler and more readable (push/pull are both 4 letters, start with `p`, and seem harder to visually tell the difference between to me). Is there not a good mapping between pull/read and push/write?

sqs: Because I wanted to make it clear it maps directly to GitHub's permissions (which use those terms). When we create our own permissions system, we should use read/write/admin/whatever. Think that's a good enough reason?

shurcooL: That's reasonable.

PR #217, opened 2016-07-05
limit word combinations for fulltext search to first 6 words

neelance:
This is to fix the "string is too long for tsvector" error (and not waste performance).

### DIFF
--- a/services/backend/internal/localstore/defs.go
+++ b/services/backend/internal/localstore/defs.go
@@ -666,7 +666,7 @@ func toTextSearchTokens(def *graph.Def) (aToks []string, bToks []string, cToks [
 	pathParts := delims.Split(def.Path, -1)
 	for _, w := range pathParts {
 		bToks = appendRepeated(bToks, w, 2)
-		for _, c := range allCombinations(splitCaseWords(w)) {
+		for _, c := range boundedCombinations(splitCaseWords(w), 6) {
 			if c != "" {
 				cToks = appendRepeated(cToks, c, 1)
 			}
@@ -674,7 +674,7 @@ func toTextSearchTokens(def *graph.Def) (aToks []string, bToks []string, cToks [
 	}
 	lastPathPart := pathParts[len(pathParts)-1]
 	aToks = appendRepeated(aToks, lastPathPart, 3) // mega extra points for matching the last component of the def path (typically the "name" of the def)
-	for _, c := range allCombinations(splitCaseWords(lastPathPart)) {
+	for _, c := range boundedCombinations(splitCaseWords(lastPathPart), 6) {
 		if c != "" {
 			aToks = appendRepeated(aToks, c, 1) // more points for matching last component of def path
 		}
@@ -701,6 +701,13 @@ func splitCaseWords(w string) []string {
 		return strings.Split(w, "_")
 	}
 	return camelcase.Split(w)
+}
+
+func boundedCombinations(s []string, bound int) []string {
+	if len(s) > bound {
+		return append(allCombinations(s[:bound]), s[bound:]...)
+	}
+	return allCombinations(s)
 }
 
 // allCombinations returns all strings that can be built by concatenating a subset of the given strings without reordering.
--- a/services/backend/internal/localstore/defs_test.go
+++ b/services/backend/internal/localstore/defs_test.go
@@ -11,7 +11,7 @@ func TestToTextSearchTokens(t *testing.T) {
 		DefKey: graph.DefKey{
 			Repo: "repo1/repo2",
 			Unit: "unit1/unit2",
-			Path: "path_x/pathFooBar",
+			Path: "path_x/pathFooBarHelloWorldThisIsLong",
 		},
 		File: "file1/file2",
 		Name: "name",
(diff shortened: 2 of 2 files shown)


services/backend/internal/localstore/defs.go
```diff
@@ -664,21 +664,24 @@ func toTextSearchTokens(def *graph.Def) (aToks []string, bToks []string, cToks [
 	bToks = appendRepeated(bToks, unitParts[len(unitParts)-1], 2)
 
 	pathParts := delims.Split(def.Path, -1)
-	for _, w := range pathParts {
+	for i, w := range pathParts {
 		bToks = appendRepeated(bToks, w, 2)
-		for _, c := range allCombinations(splitCaseWords(w)) {
+
+		words := splitCaseWords(w)
+		if len(words) > 6 { // limit max number of combinations to 2**6 = 64
```
keegancsmith: Should we be even more conservative. IE should we care about combinations of a path part that isn't the last path part.

services/backend/internal/localstore/defs.go
```diff
@@ -664,21 +664,24 @@ func toTextSearchTokens(def *graph.Def) (aToks []string, bToks []string, cToks [
...
+			cToks = append(cToks, words[6:]...)
+			words = words[:6]
+		}
+		for _, c := range allCombinations(words) {
 			if c != "" {
 				cToks = appendRepeated(cToks, c, 1)
+				if i == len(pathParts)-1 {
+					aToks = appendRepeated(aToks, c, 1) // more points for matching last component of def path
+				}
```
keegancsmith: having `cToks` and `aToks` dealt with here feels less understandable than before. You also don't do `append(aToks, words[6:]...)` like you do for `cToks`

Maybe a wrapper `boundedCombinations(words []string, bound int)` which returns `append(allCombinations(words[:6]), words[6:]...)` would allow you to move this back outside of this loop + include `words[6:]...`.

keegancsmith:
Sorry, I even thought about this and somewhat commented on it in the last review, but obviously didn't think properly about it given the `O(n^2)` output.

Shall we add some sort of sanity check at the end of the function to ensure we never have crazy large tsvectors?

Once you have responded to feedback, feel free to merge in.

services/backend/internal/localstore/defs.go
neelance: Performance wise I am not worried about that, because it only scales linearly with the length of the path. In terms of results quality I'd like to see a case of bad results before removing it.

keegancsmith: I'm more talking about indexing unnecessary amounts of data in postgres, without really improving search. But sure, we can evaluate afterwards.

services/backend/internal/localstore/defs.go
neelance: Good idea, did that. More readable now?

neelance:
Where would you put the bound of the sanity check?

keegancsmith:
not sure, not 100% familiar with the code. The nicer way would be to have the sanity check in the caller. But if it is called often, or it doesn't feel right in the caller, at the end of the function seems good to me.

services/backend/internal/localstore/defs.go
keegancsmith: ooh, much more readable. Thanks!

neelance:
Hrhr, miscommunication. With "where" I meant "at which size of tsvector". I can't decide, feels arbitrary to me.

keegancsmith:
Maybe take a look at the distribution of tsvector sizes in the db as it stands and take a guess?

neelance:
The column itself does not store the initial data calculated by this function. Postgres itself has this 1MB sanity check. It's relatively high...

keegancsmith:
@neelance lets just merge this in and follow-up with the sanity check. As it is this will fix an issue in prod.

beyang:
@keegan @richard why do we need all permutations. Why not just use the subset of permutations that correspond to a prefix of the original token? E.g., `oneTwoThree` -> {`one`, `oneTwo`, `oneTwoThree`}.

If a user searches for `oneThree`, we don't even necessarily want to return `oneTwoThree` in the search results. All we want is prefix search (e.g., `one` or `oneTwo`). We have the single-word lexemes (`one`, `two`, `three`) for fuzzier matching anyway.

neelance:
Remark: Searching for `one three` is much more fuzzier than searching for `onethree`. The words may even be at different places of the doc comment.

beyang:
This issue might be fixed for now, but in the future it could crop up again in terms of reduced search performance because we are indexing many extra tokens per row. What's a real-world case where you'd actually search for `onethree` and expect `oneTwoTree` as a result? Typically you are either searching for a rough match (in which case you'd probably separate search terms by spaces) or an exact name match (in which case you'd be typing out the exact name).

beyang:
Now the main search index is half the size of the table as a whole. When you try queries in production, they are insanely slow...

neelance:
With this fix applied the max number of tokens generated is 64. This should not have much negative impact. I see benefit in having those results when looking for something that you don't remember the exact name for. I think it's not unreasonable to type `buildfunc` when you're looking for `BuildVarsFunc`.

PR #218, opened 2016-07-05
change vertical align for dropdown to realign translate button

uforic:
Previously: 
screen shot 2016-07-05 at 1 48 42 pm
Now:
screen shot 2016-07-05 at 1 48 54 pm

##### Reviewer tasks

@chexee , mind taking a look? I checked in our code base, and I don't see the <Dropdown used anywhere else except this instance in DefInfo. 

Also, @chexee , if vertical-align isn't specified, this fixes it as well. maybe cleaner than forcing it?

##### Test plan

Tested locally, and is fixed. Checked if anyone else uses Dropdown.css, and couldn't find any users.


rothfels:
LGTM, merging.

PR #220, opened 2016-07-05
app/internal/gzipfileserver: Serve files only, not directories.

shurcooL:
There's no point in serving directory listings in gzipfileserver. It's
used to serve assets.

Fixes sourcegraph/sourcegraph-issues#13.


beyang:
LGTM

PR #223, opened 2016-07-06
Prevent unauthorized repo api access from leaking info. 401 -> 404

samlavery:
Small change to swallow errors that would indicate that a valid repo exists at a URL, preventing repo enumeration attacks.  We simply always return a 404 externally in the case of 404 or 401 Unauthorized.

### DIFF
--- a/services/backend/internal/localstore/repos.go
+++ b/services/backend/internal/localstore/repos.go
@@ -194,6 +194,10 @@ func verifyAccessAndSetAllFields(ctx context.Context, repo *sourcegraph.Repo) (*
 	// accesscontrol.VerifyUserHasReadAccess calls (*repos).Get).
 	if strings.HasPrefix(strings.ToLower(repo.URI), "github.com/") {
 		if err := accesscontrol.VerifyActorHasGitHubRepoAccess(ctx, auth.ActorFromContext(ctx), "Repos.Get", repo.ID, repo.URI); err != nil {
+			thrown := grpc.Code(err)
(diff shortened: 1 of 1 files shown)


services/backend/internal/localstore/repos.go
```diff
@@ -194,6 +194,11 @@ func verifyAccessAndSetAllFields(ctx context.Context, repo *sourcegraph.Repo) (*
 	// accesscontrol.VerifyUserHasReadAccess calls (*repos).Get).
 	if strings.HasPrefix(strings.ToLower(repo.URI), "github.com/") {
 		if err := accesscontrol.VerifyActorHasGitHubRepoAccess(ctx, auth.ActorFromContext(ctx), "Repos.Get", repo.ID, repo.URI); err != nil {
+			thrown := grpc.Code(err)
+			if codes.Code(thrown) == codes.Unauthenticated {
+				log15.Debug("Suppression of error message: 401 not authorized REPLACED by 404 not found) for User ", "UID", string(auth.ActorFromContext(ctx).UID))
```
sqs: unclosed close paren after "found"

sqs: No need to wrap the int in `string(...)`.

sqs: In the log message, include which repo.

Actually, is this worth logging at all? It doesn't seem logworthy. It seems like we need to suppress this as a matter of core functionality, not to handle a hacky/bug case.

PR #224, opened 2016-07-06
Add platform to event properties

KattMingMing:
##### Description

Adds a platform for analytics calls that matches Platform for ChromeExtension. This also means event properties can be Web, Chrome ext, integration and also pass in additional props that let us know how the integration was kicked off i.e Platform: Desktop that came from Sublime.

### DIFF
--- a/app/web_modules/sourcegraph/util/EventLogger.js
+++ b/app/web_modules/sourcegraph/util/EventLogger.js
@@ -43,17 +43,15 @@ export class EventLogger {
 		// with the relevant event properties.
 		this._dispatcherToken = Dispatcher.Stores.register(this.__onDispatch.bind(this));
 
-		document.addEventListener("sourcegraph:platform:initalization", this._initializeForSourcegraphPlatform.bind(this));
+		if (typeof document !== "undefined") {
+			document.addEventListener("sourcegraph:platform:initalization", this._initializeForSourcegraphPlatform.bind(this));
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


uforic:
LGTM - only other thing that comes to mind - does the event logger on the Chrome Extension need to change at all because of this? Can we rip out some of the logic on this code:
```if (location.query && location.query["utm_source"] === "integration" && location.query["type"]) {
				eventProps = {
					// Alfred, ChromeExtension, FireFoxExtension, SublimeEditor, VIMEditor.
					referred_by_integration: location.query["type"],
					url: location.pathname,
				};
			} else if (location.query && location.query["utm_source"] === "chromeext") {
				// TODO:matt remove this once all plugins are switched to new version
				// This is temporarily here for backwards compat
				eventProps = {
					referred_by_browser_ext: "chrome",
					url: location.pathname,
				};
			} else if (location.query && location.query["utm_source"] === "browser-ext" && location.query["browser_type"]) {
				eventProps = {
					referred_by_browser_ext: location.query["browser_type"],
					url: location.pathname,
				};
			} else if (location.query && location.query["utm_source"] === "sourcegraph-editor" && location.query["editor_type"]) {
				eventProps = {
					url: location.pathname,
					referred_by_sourcegraph_editor: location.query["editor_type"],
				};
			} else {
				eventProps = {
					url: location.pathname,
				};
			}``` by just setting platform instead of these referred by properties?

PR #226, opened 2016-07-06
Fix search for single tokens

neelance:
See commit messages for explanations.


neelance:
This may have a performance impact, but I see no reason to hold back the change, because it is a necessary improvement. Local testing seemed fast enough.

keegancsmith:
LGTM

I don't understand the old behaviour easier, since it is very surprising. We should be relying on ranking. If there is a good reason though, we can easily revert. So please merge in!

PR #227, opened 2016-07-06
Misc changes to UI and code cleanup

sqs:
These were misc. changes I encountered while going through the UI code.

### DIFF
--- a/app/web_modules/sourcegraph/delta/FileDiff.js
+++ b/app/web_modules/sourcegraph/delta/FileDiff.js
@@ -106,7 +106,7 @@ class FileDiff extends Component {
 			<div styleName="container" id={this.props.id || ""}>
 				<header styleName="header">
 					<div styleName="info">
-						<DiffStatScale Stat={diff.Stats} />
+						<DiffStatScale Stat={diff.Stats} styleName="scale" />
 						<span styleName="name">
 							<span>{isDevNull(diff.OrigName) ? diff.NewName : diff.OrigName}</span>
 							{diff.NewName !== diff.OrigName && !isDevNull(diff.OrigName) && !isDevNull(diff.NewName) ? (
--- a/app/web_modules/sourcegraph/delta/FileDiff.js
+++ b/app/web_modules/sourcegraph/delta/FileDiff.js
@@ -88,7 +88,7 @@ class FileDiff extends Component {
 
 	render() {
 		let diff = this.props.diff;
-		let hunkAnns = this._groupAnnotationsByHunk(diff.Hunks);
+		let hunkAnns = diff.Hunks ? this._groupAnnotationsByHunk(diff.Hunks) : null;
 		return (
 			<div styleName="container" id={this.props.id || ""}>
 				<header styleName="header">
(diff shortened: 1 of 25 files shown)


app/web_modules/sourcegraph/delta/FileDiff.js
```diff
@@ -88,7 +88,7 @@ class FileDiff extends Component {
 
 	render() {
 		let diff = this.props.diff;
-		let hunkAnns = this._groupAnnotationsByHunk(diff.Hunks);
+		let hunkAnns = diff.Hunks ? this._groupAnnotationsByHunk(diff.Hunks) : null;
```
keegancsmith: I don't know our JS codebase well. Why can't we avoid null in this case and have the model/store/whatever layer return an empty list to avoid null checks sprinkled everywhere.

app/web_modules/sourcegraph/delta/Hunk.js
```diff
@@ -1,11 +1,16 @@
...
 import CSSModules from "react-css-modules";
 import {atob} from "abab";
 import BlobLine from "sourcegraph/blob/BlobLine";
 import fileLines from "sourcegraph/util/fileLines";
 
-class Hunk extends React.Component {
+class Hunk extends Component {
+	reconcileState(state, props) {
+		Object.assign(state, props);
+	}
```
keegancsmith: I'm interested why we need to implement this now? I'll probably ask some more noobs questions on this PR in the interest of me learning the frontend.

app/web_modules/sourcegraph/delta/FileDiff.js
sqs: Because the empty array literal is a new object that is not equal to any other empty array literal. Therefore React thinks the state is dirty and rerenders things unnecessarily.

services/backend/vcs_util.go
```diff
@@ -23,10 +23,24 @@ func blameFileByteRange(r vcs.Repository, path string, opt *vcs.BlameOptions, st
 	var hunks2 []*vcs.Hunk // filtered
 	for _, h := range hunks {
 		if h.StartByte <= endByte && h.EndByte > startByte {
+			copied := false
```
keegancsmith: nice catch. I assume this actually led to some subtle and rare bugs?

keegancsmith:
Not that I can review this completely properly, but the changes LGTM

app/web_modules/sourcegraph/delta/Hunk.js
rothfels: @keegancsmith we have two subclasses of `React.Component` called `Component` and `Container` that are documented in `ui-guide.md`.

rothfels:
This LGTM, merging.

app/web_modules/sourcegraph/delta/FileDiff.js
keegancsmith: that is truly awful. I couldn't believe it, so had to confirm, unfortunately you are right :'(
```
JSON.parse('{}') == JSON.parse('{}');
false
JSON.parse('{}') === JSON.parse('{}');
false
```

PR #231, opened 2016-07-07
all: final changes for Sourcegraph beta program

slimsag:
- Add a link to `/beta` from GlobalNav (otherwise users can never land on that page,
  as there are no other links to it anywhere / all other uses are inside embedded
  frames).
- Remove the "Email Address" field from the BetaInterestForm; use the account
  primary email instead. This solves a tougher technical aspect of keeping Mailchimp
  in sync, otherwise we would need a bit on each user email address specifying if it
  was the one they registered into Mailchimp or not.
- Add CLI flags to properly manage beta users in all meaningful ways:
  - `src user list [--all-betas=...]`
  - `src user update [--add-betas=...] [--rm-betas=...] [--set-betas=...] <user>`
  - `src user get <user>` now shows betas a user is in.
  - See Quip document below for more details.
- Add wording to the beta page to explain why the form still shows (see BetaInterestForm.js:144-149).
- Tell the user which email address (always their account primary email address) that
  we will be emailing them at.
- When opening the beta form via a modal, always display the success text after
  form submission, and display a close button to exit the modal instead of closing
  it immediately upon submission (which did not show potential errors, either).
- Implement the betas backend, which properly keeps Mailchimp in sync + stores which
  betas a user is part of in Postgres.
- Implement helpers for detecting whether or not a user is in a given beta in both
  Go and JS, see [relevant Quip document describing beta user management](https://sourcegraph.quip.com/KTaTA0pLkt8v) for more details.

Users will need to run the following to migrate to the new DB schema:

```
ALTER TABLE users ADD COLUMN betas text ARRAY;
ALTER TABLE users ADD COLUMN beta_registered boolean;
```

(within the app DB)

##### Reviewer tasks

- I've spent a lot of time making sure this is the _right_ implementation, so please raise any concerns you might have. I've also tested pretty extensively in many different situations.
- Please also briefly review [the Quip document describing beta user management](https://sourcegraph.quip.com/KTaTA0pLkt8v) and comment on anything that stands out.
- I'm very confident it works, but please checkout `sg/beta-cli-server` and run the above DB migration command in `psql`. You can `ALTER TABLE users DROP COLUMN betas;` after to revert the migration. Confirm that nothing weird happens.

##### Test plan

Testing has been quite extensive, but manual. Much of it relies on external services and as such it seems difficult to test under traditional unit tests.

To deploy, I will:
  1. Pin the sourcegraph.com version to master.
  2. Merge this branch to `master`.
  3. Wait for CI build.
  4. Perform the DB migration in prod.
  5. Deploy the new `master` to prod.
  6. Confirm everything looks OK and is working. Any issue can be fixed immediately by rolling back after running the above revertion SQL.


slimsag:
@rothfels if you can review at least the user experience side of this / confirm the new wording / styling looks OK?

If you don't have time for a full code review of this, I can assign someone else.

keegancsmith:
> 4. Perform the DB migration in prod.
Can't you do that as your first step? Then you should be able to just merge into master without the extra steps of pining.

rothfels:
@slimsag are you aware that we don't always have a user email address? If you create an account w/ GitHub login and all your emails are listed as private, there will be no entry in the `user_email` table for that user. And (afaict right now) there's no way to subsequently add an email to that table. In this case, we might need to still ask for the user's email address on the form. (And, maybe, we should go ahead and save that email address to the emails table, though if we do we should be careful not to fail the beta form POST if the user chooses an email address which is already taken). I agree that in the normal case it is better to use the email address we already have saved for the user. 👍

api/sourcegraph/sourcegraph.proto
```diff
@@ -978,6 +978,28 @@ message User {
...
+	// Betas is a list of betas which the user is enrolled in. A user may be
+	// granted access to any arbitrary beta string, but "pending" and "accepted" are
+	// reserved. The three states are:
+	//
+	//  1. `len(Betas) == 0` (user is in no betas and has not requested beta access).
+	//  2. `len(Betas) == 1 && InBeta("pending")` (user has requested beta access).
+	//  4. `len(Betas) > 1 && InBeta("accepted")` (user granted access to one or more betas).
+	//
+	// The "pending" and "accepted" statuses are implicitly set as you would
+	// expect, thus their present in the list can generally be thought of as
```
rothfels: s/present/presence

api/sourcegraph/sourcegraph.proto
```diff
@@ -978,6 +978,28 @@ message User {
...
+	// optional.
```
rothfels: It doesn't look optional, given your comments above.

api/sourcegraph/sourcegraph.proto
```diff
@@ -978,6 +978,28 @@ message User {
 	// Admin is whether the user is a site admin for the site.
 	bool Admin = 12;
 
+	// Betas is a list of betas which the user is enrolled in. A user may be
+	// granted access to any arbitrary beta string, but "pending" and "accepted" are
+	// reserved. The three states are:
+	//
+	//  1. `len(Betas) == 0` (user is in no betas and has not requested beta access).
+	//  2. `len(Betas) == 1 && InBeta("pending")` (user has requested beta access).
+	//  4. `len(Betas) > 1 && InBeta("accepted")` (user granted access to one or more betas).
```
rothfels: s/4/3

api/sourcegraph/sourcegraph.proto
```diff
@@ -978,6 +978,28 @@ message User {
...
+	//
+	// To add a beta status, simply append it. The "pending" status will be
+	// removed and the "accepted" status added.
+	//
+	// To remove a beta status, simply remove it from the slice. If the user is
+	// in no more betas, the "accepted" status will be removed and "pending"
+	// added.
+	//
+	// Only admins may set a list of non-"pending" betas.
+	repeated string Betas = 14;
```
rothfels: Was there a reason you didn't break this into two fields, `BetaStatus` and `Betas`?

api/sourcegraph/sourcegraph.proto
```diff
@@ -1172,6 +1198,38 @@ message NewAccount {
 	int32 UID = 4;
 }
 
+// BetaRegistration represents the user information needed to register for the
+// beta program.
+message BetaRegistration {
+	// Email is the email to register for the beta program. If it is an empty
+	// string, the account's primary email address will be used.
+	//
+	// TODO(slimsag): this field is not currently supported. Do not use it.
```
rothfels: I think we need to support this; see my top-level comment on the PR.

api/sourcegraph/sourcegraph.proto
```diff
@@ -1172,6 +1198,38 @@ message NewAccount {
...
+
+	// LastName is the last name of the user.
+	string LastName = 3;
+
+	// Languages is a list of programming languages the user is interested in.
+	repeated string Languages = 4;
+
+	// Editors is a list of editors the user is interested in.
+	repeated string Editors = 5;
+
```
rothfels: We had suggested several other important fields to add: team size, role. Do you want to add those now or in a follow-on?

app/web_modules/sourcegraph/home/BetaInterestForm.js
```diff
@@ -61,11 +60,22 @@ class BetaInterestForm extends React.Component {
 		}
 	}
 
+	_didRegister() {
+		return this.context.user && this.context.user.Betas;
```
rothfels: Isn't this going to be true because you don't `omitempty` on the `Betas` field? In JavaScript, `Boolean([])` => `true`

app/web_modules/sourcegraph/home/BetaInterestForm.js
```diff
@@ -61,11 +60,22 @@ class BetaInterestForm extends React.Component {
...
+	}
+
+	_betasWithoutStatuses() {
```
rothfels: Wouldn't be necessary if you broke out beta status separately.

app/web_modules/sourcegraph/home/BetaInterestForm.js
```diff
@@ -106,18 +114,24 @@ class BetaInterestForm extends React.Component {
 
 	render() {
 		if (this.state.resp && !this.state.resp.Error) {
-			return (<p>Thank you for registering. You will hear from us soon.</p>);
+			// Display a "Close" button if there is an onSubmit handler.
+			return (<span>
+				<p>Success! Return to this page any time to update your favorite editors / languages!</p>
+				<p>We'll contact you at <strong>{this.state.resp.EmailAddress}</strong> once a beta program has begun.</p>
```
rothfels: remove "program"

app/web_modules/sourcegraph/home/BetaInterestForm.js
```diff
@@ -126,30 +140,35 @@ class BetaInterestForm extends React.Component {
...
-						<Button block={true} type="submit" color="purple">Participate in the beta</Button>
-					</div>
-					<div styleName="row" className={base.pb4}>
-						{this.state.formError && <strong>{this.state.formError}</strong>}
-						{this.state.resp && this.state.resp.Error && <div>{this.state.resp.Error.body}</div>}
-					</div>
-			</form>
+			<div>
+				{didRegister && betas.length === 0 && <span>
+					<p>You've already registered. We'll contact you once a beta program matching your interests has begun.</p>
```
rothfels: Remove "program"

app/web_modules/sourcegraph/home/BetaInterestForm.js
```diff
@@ -126,30 +140,35 @@ class BetaInterestForm extends React.Component {
...
+				</span>}
+				{betas.length === 1 && <p>You've been granted access to the <em>{betas[0]}</em> beta! Check your email for more information.</p>}
+				{betas.length > 1 && <p>You've been granted access to the following betas: <em>{betas.join(", ")}</em>. Check your email for more information.</p>}
```
rothfels: I don't think we need to tell the user right now precisely which betas they've been added to. Let's just tell the user in all cases once they've registered that we'll contact them by email.

app/web_modules/sourcegraph/user/index.js
```diff
@@ -7,8 +7,15 @@ import type {SearchSettings} from "sourcegraph/search";
 export type User = {
 	UID: number;
 	Login: string;
+	Betas?: string[];
```
rothfels: I think the type annotation should be `Array<string>`

services/backend/users.go
```diff
@@ -54,3 +60,59 @@ func (s *users) List(ctx context.Context, opt *sourcegraph.UsersListOptions) (*s
...
+	primary, err := emails.Primary()
+	if err != nil {
+		return nil, err
+	}
+	opt.Email = primary.Email
+
+	// If the account does not already have any betas, give them pending beta
+	// access.
+	user, err := svc.Users(ctx).Get(ctx, &userSpec)
+	if len(user.Betas) == 0 {
```
slimsag: I missed error handling here; will fix.

api/sourcegraph/sourcegraph.proto
slimsag: There were at the time I implemented this portion. But I agree that now, looking at the entire picture, it's better separated. Done.

api/sourcegraph/sourcegraph.proto
slimsag: Looking into it.

api/sourcegraph/sourcegraph.proto
slimsag: Yes, these were in the quip doc / specification. However, I did add these in one of my previous PRs and Quinn suggested (over Slack) that we remove them:

> let's remove team size and team role, since those can be hard to answer and we can usually find it out on linkedin

I also agree with this.

rothfels: Sounds good.

app/web_modules/sourcegraph/user/index.js
slimsag: Are you saying for correctness, or for style? Because `string[]` and `Array<string>` are identical, it's just a short-hand: See https://flowtype.org/docs/arrays.html#_

> The shorthand syntax T[] is equivalent to Array<T>.

slimsag:
All comments addressed except your top level concern; which I've ping'd you on Slack about (pretty sure it's impossible to not get an email address from our GitHub users)

slimsag:
@keegancsmith

> Can't you do that as your first step? Then you should be able to just merge into master without the extra steps of pining.

Nope, the ORM we're using is pretty inflexible in this manner, and for example breaks account registration if I attempt to do that:

```
EROR[07-08|20:42:12] Error during GitHub account creation (suppressing HTTP 500 and returning redirect to non-GitHub signup form). err="rpc error: code = 2 desc = gorp: No fields [betas beta_registered] in type dbUser" github-login=slimsag
```

slimsag:
This is ready to ship. Unless someone objects I will merge + deploy late Sunday night / early Monday morning.

api/sourcegraph/sourcegraph.proto
```diff
@@ -978,6 +978,24 @@ message User {
...
+	//  pkg/betautil/betautil.go
+	//  app/web_modules/sourcegraph/util/betautil.js
+	//
+	// Only admin users may set this field.
+	repeated string Betas = 14;
+
+	// BetaRegistered is whether or not the user has registered for beta access.
+	//
+	// If a user has registered for beta access, this field will always be true
+	// even if they are participating in a beta (len(User.Betas) > 0).
```
rothfels: I think you mean even if they are *not* participating in a beta (len == 0)

app/web_modules/sourcegraph/user/index.js
```diff
@@ -7,8 +7,29 @@ import type {SearchSettings} from "sourcegraph/search";
 export type User = {
 	UID: number;
 	Login: string;
+	Betas: string[];
+	BetaRegistered: bool;
 };
 
+// InBeta tells if the given user is a part of the given beta program.
+export function InBeta(u?: User, b: string): boolean {
```
rothfels: Are you sure you don't want `u: ?User`?

rothfels: For stylistic consistency, these functions should be fully camel case (i.e. `inBeta`). We are not in Go. :P

rothfels: (Usually in JavaScript capitalized names are for types/classes)

app/web_modules/sourcegraph/util/betautil.js
```diff
@@ -0,0 +1,12 @@
...
+//  pkg/betautil/betautil.go
+//  app/web_modules/sourcegraph/util/betautil.js
+//
+// Beta strings should be human-readable, and should not be suffixed with
+// "Beta". Add a constant below:
+//
+// Example const:
+//
+//  export const AWESOME_STUFF = "Awesome Stuff";
+//
```
rothfels: Can we commit this with a couple constants? We should have the following:

- JavaScript
- Python
- Desktop

cli/user_cmds.go
```diff
@@ -116,7 +117,10 @@ func (c *userCreateCmd) Execute(args []string) error {
 }
 
 type userListCmd struct {
-	Args struct {
+	AllBetas       string `long:"all-betas" description:"only users participating in all the given betas"`
+	RegisteredBeta bool   `long:"registered-beta" description:"filter by users who have registered for beta access"`
+	HaveBeta       bool   `long:"have-beta" description:"filter by users who have access to at least one beta"`
+	Args           struct {
 		Query string `name:"QUERY" description:"search query"`
```
rothfels: Looks like a `gofmt` issue.

cli/user_cmds.go
```diff
@@ -224,9 +241,76 @@ func (c *userUpdateCmd) Execute(args []string) error {
...
 
+	updateBetas := func(betas []string) error {
+		oldBetas := user.Betas
+		user.Betas = betas
+		if _, err := cl.Accounts.Update(cliContext, user); err != nil {
+			return err
+		}
+		// Get the user again, because Update implicitly manages the pending
+		// and accepted beta statuses.
+		user, err = cl.Users.Get(cliContext, &userSpec)
```
rothfels: I don't see why you need to do this. Is it just for the `fmt.Printf`? Because if so you can use the function argument for the printout.

services/backend/accounts.go
```diff
@@ -83,10 +87,80 @@ func (s *accounts) Create(ctx context.Context, newAcct *sourcegraph.NewAccount)
...
+			if !betautil.Betas[beta] {
+				return nil, fmt.Errorf("no such beta %q", beta)
+			}
+		}
+
+		// If there is at least one beta, ensure the BetaRegistered field is also set.
+		in.BetaRegistered = true
+	}
+
 	if err := store.AccountsFromContext(ctx).Update(ctx, in); err != nil {
```
rothfels: Does this require admin permissions? If not, users could accept themselves.

app/web_modules/sourcegraph/home/BetaInterestForm.js
```diff
@@ -126,30 +165,36 @@ class BetaInterestForm extends React.Component {
...
+				{betaRegistered && <span>
+					<p>You've already registered. We'll contact you once a beta matching your interests has begun.</p>
+					<p>Feel free to update your favorite editors / languages using the form below.</p>
+				</span>}
+				<form styleName="form" className={className} onSubmit={this._sendForm.bind(this)} onChange={this._onChange}>
+						<div styleName="row">
+							<Input domRef={(c) => this._fullName = c} block={true} type="text" name="fullName" placeholder="Name" required={true} defaultValue={defaultFullName} />
+						</div>
+						{(!emails || emails.length === 0) && <div styleName="row">
+							<Input domRef={(c) => this._email = c} block={true} type="email" name="email" placeholder="Email address" required={true} defaultValue={defaultEmail} />
```
rothfels: Do you want to disable this field if the user already has an email?

rothfels:
Small comments otherwise LGTM

app/web_modules/sourcegraph/user/index.js
slimsag: I'll try to read online about the difference between `u?: User` and `u: ?User` ... I forget what the difference is.

api/sourcegraph/sourcegraph.proto
slimsag: No. The point is, `User.BetaRegistered` says whether or not the user has at any point in the past said "I want beta access" and from that point onward it never changes, regardless of whether or not they are participating in beta programs actively (`len(User.Betas) > 0`).

If `len(User.Betas) == 0` then the user is not actively participating in beta programs, and `BetaRegistered` may be either `true` or `false`, depending on whether or not the user has said "I want beta access" -- there is no relationship between length 0 and the betas field.

I only mentioned `even if they are participating in a beta (len(User.Betas) > 0).` because one could become confused when reading this thinking that

> "a users BetaRegister == true ('I want beta access!') status may be removed after they are placed into a beta program"

And thus we especially point out that is _not_ the case here.

app/web_modules/sourcegraph/util/betautil.js
slimsag: Perfect, I wanted to add these instead of the "example" comments but wasn't sure which ones we'd actually use. I will add these.

cli/user_cmds.go
slimsag: Looks fine to me. I don't think it's an issue. The type is `struct` and thus `gofmt` tabs it over by one column.

cli/user_cmds.go
slimsag: You're right, it's no longer needed. I should've removed it.

cli/user_cmds.go
shurcooL: I don't see a `gofmt` issue here.

Did you miss that `Query` is inside an anonymous struct `Args`? It's not a direct field of `userListCmd` struct like the others. Look at `userCreateCmd` above, it has the same pattern for `Args`.

app/web_modules/sourcegraph/home/BetaInterestForm.js
slimsag: I'd rather not display it. If we don't display the field, users will assume we're getting the email address from somewhere else (their account info), but if it's disabled they might find it confusing and contact us with "I can't fill out the email address field.. looks broken to me" and we'll have to explain why it is there but not usable.

keegancsmith:
@slimsag That message sounds like it can't find fields in the DB, but I am suggesting just adding the fields to the DB now without the code knowing about it. Gorp shouldn't complain since it generates queries/insertions with specifying the fields (ie it doesn't use `*` and assume ordering).

slimsag:
@keegancsmith the error I posted is from doing exactly that! 😄 I would have thought it would work too, but it doesn't. I guess it does use `*` and assume the ordering! To be clear, what I did:

1. Using `master` (without these changes) run `ALTER TABLE users ADD COLUMN betas text ARRAY;` and `ALTER TABLE users ADD COLUMN beta_registered boolean;`
2. Run `make serve-dev`
3. Create a new account, get the errors above!

keegancsmith:
@slimsag doh! that sucks :P Maybe it depends how we use it, cause I've done it before (I guess most of my stuff avoids most of the ORM and does raw sql).

slimsag:
yeah.. I think we maybe should look at dropping the ORM; in my experience it's more limiting than plain sql :) :)

PR #232, opened 2016-07-07
search: remove "Latest" option

neelance:
This option was always true. Setting it to false did not work any more.


keegancsmith:
LGTM

PR #234, opened 2016-07-07
Various changes related to search

neelance:


cli/search_cmd.go
```diff
@@ -30,23 +30,18 @@ type searchCmd struct {
...
-		})
-		if err != nil {
-			return err
-		}
-	} else if c.Refresh != "" {
-		log.Printf("Def.RefreshIndex")
-		res, err := cl.Repos.Resolve(cliContext, &sourcegraph.RepoResolveOp{Path: c.Refresh})
-		if err != nil {
+
+		if c.RefreshAsync {
```
keegancsmith: this is not the same now. We will both put a job on the queue and then run call refresh and block. Currently refreshasync does do something different, but I want to change that soon.

neelance: This is not the same as before the commit? Why?

services/backend/internal/localstore/repos.go
```diff
@@ -271,6 +271,11 @@ func (s *repos) List(ctx context.Context, opt *sourcegraph.RepoListOptions) ([]*
 	if opt == nil {
 		opt = &sourcegraph.RepoListOptions{}
 	}
+	if opt.All {
```
keegancsmith: the paranoid in me wants a simple test ensuring that you have to be admin for all to work. Don't want a refactor to result in a security leak.

services/backend/internal/localstore/repos.go
```diff
@@ -507,6 +512,10 @@ func (s *repos) Search(ctx context.Context, query string) ([]*sourcegraph.RepoSe
 var errOptionsSpecifyEmptyResult = errors.New("pgsql: options specify and empty result set")
 
 func (s *repos) listSQL(opt *sourcegraph.RepoListOptions, allowGitHubRepoIDs []string) (string, []interface{}, error) {
+	if opt.All {
+		return "SELECT repo.* FROM repo", nil, nil
+	}
```
keegancsmith: in fact, doing the All check here would be best? but it would be a shame to introduce ctx here.

services/backend/internal/localstore/repos.go
```diff
@@ -271,6 +271,11 @@ func (s *repos) List(ctx context.Context, opt *sourcegraph.RepoListOptions) ([]*
 	if opt == nil {
 		opt = &sourcegraph.RepoListOptions{}
 	}
+	if opt.All {
+		if err := accesscontrol.VerifyUserHasAdminAccess(ctx, "Repos.List"); err != nil {
+			return nil, err
+		}
+	}
```
keegancsmith: relatedly, I I guess the feeling this will hit github in our current architecture (in the backend layer, not the store layer currently). That will result in guaranteed downtime due to rate-limit exhaustion.

cli/search_cmd.go
```diff
@@ -20,7 +20,8 @@ func init() {
 }
 
 type searchCmd struct {
-	Refresh      string `long:"refresh" description:"repository URI for which to update the search index and counts (implies --refresh-counts)"`
+	Refresh      string `long:"refresh" description:"repository URI for which to update the search index and counts"`
+	RefreshAll   bool   `long:"refresh-all" description:"update the search index and counts for all repos"`
```
keegancsmith: do we really want to make it that easy for someone to run that? Maybe just a little bit harder and make this take a list of URI and then they have to use `xargs`.

keegancsmith:
I have issues with only some of the commits. Most of them I don't though, and they make changes to code I wanna hack on as well. Do you mind merging just those in? I can also cherry-pick and merge the ones I am happy with manually now if ya like.

services/backend/internal/localstore/repos.go
neelance: There's already a `VerifyUserHasReadAccess` above this. Do you think this will make a negative difference? Do I need to change this?

cli/search_cmd.go
neelance: I think `xargs` may fail with this big list. Also, then we would need to either accept integer IDs or resolve each of them. What do you want to prevent here?

services/backend/internal/localstore/repos.go
keegancsmith: This is a problem. After calling list from the repos store we then speak to github. https://github.com/sourcegraph/sourcegraph/blob/master/services/backend/repos.go#L90-L110

services/backend/internal/localstore/repos.go
neelance: I'll add a test.

cli/search_cmd.go
keegancsmith: resolve each of them. `xargs` has the `-n` argument for large lists.

services/backend/internal/localstore/repos.go
neelance: Why isn't this a problem before my change?

keegancsmith: we never returned all repos before in List. In fact we had some downtime recently thanks to a change which allowed us to do that.

services/backend/internal/localstore/repos.go
neelance: Done.

services/backend/internal/localstore/repos.go
neelance: I moved the check to there. It's safer.

cli/search_cmd.go
keegancsmith: before we would only run async.refreshindexes if  c.RefreshAsync was true.

services/backend/internal/localstore/repos.go
```diff
@@ -506,7 +506,15 @@ func (s *repos) Search(ctx context.Context, query string) ([]*sourcegraph.RepoSe
 
 var errOptionsSpecifyEmptyResult = errors.New("pgsql: options specify and empty result set")
 
-func (s *repos) listSQL(opt *sourcegraph.RepoListOptions, allowGitHubRepoIDs []string) (string, []interface{}, error) {
+func (s *repos) listSQL(ctx context.Context, opt *sourcegraph.RepoListOptions, allowGitHubRepoIDs []string) (string, []interface{}, error) {
+	if opt.All {
+		if err := accesscontrol.VerifyUserHasAdminAccess(ctx, "Repos.List"); err != nil {
+			return "", nil, err
+		}
+
```
keegancsmith: this still has the issue that in the backend layer we will speak to github -> rate limit depleted.

keegancsmith:
LGTM except the list all implementation.

services/backend/internal/localstore/repos.go
neelance: I still don't get this. Is `VerifyUserHasAdminAccess` talking to GitHub and `VerifyUserHasReadAccess` is not?

Also the `All` flag should only be set by the admin operation, which is rare.

What change do you want to see here?

cli/search_cmd.go
neelance: There is still an `if c.RefreshAsync` around the call to `cliClient.Async.RefreshIndexes`. I don't see any changed behavior.

services/backend/internal/localstore/repos.go
neelance: Do you think even a single invocation for listing all by the admin tool would be bad?

services/backend/internal/localstore/repos.go
keegancsmith: See this comment https://github.com/sourcegraph/sourcegraph/pull/234#discussion_r69906433

The backend layer makes an explicit call to github after calling the store, and unrelated to access checks. It is calling github to fetch fields like `Fork`, `Private`, etc.

PR #236, opened 2016-07-07
WIP: scroll to correct location on blob view on "back" navigation

rothfels:


app/web_modules/sourcegraph/blob/Blob.js
```diff
@@ -102,10 +98,10 @@ class Blob extends Component {
 		highlightedDef: ?string;
 		contentsOffsetLine: number;
 		expandedRanges: Range[];
-		visStartLine: number;
```
rothfels: These are removed because they're not being used anywhere, right?

app/web_modules/sourcegraph/blob/Blob.js
```diff
@@ -206,6 +202,20 @@ class Blob extends Component {
...
 		}
+		if (state.startLine && oldStartLine !== state.startLine) {
+			let scrollThunk = () => {
+				if (this.state.scrollToStartLine) {
+					this._scrollTo(state.startLine);
+				}
+			};
+			let isVisible = this._lineIsVisible(state.startLine);
+			if (isVisible === null) {
+				state.scrollTarget = state.startLine;
```
rothfels: Do these properties need to be on state? Why can't you just call scrollThunk() and pass the curr state to scroll do once isVisible === false?

app/web_modules/sourcegraph/blob/Blob.js
```diff
@@ -206,6 +202,20 @@ class Blob extends Component {
...
+				state.scrollCallback = scrollThunk;
+			} else if (isVisible === false) {
+				scrollThunk();
```
rothfels: Instead of `srcollThunk` here, could you just replace it with the body of scroll thunk?

```
if (state.scrollToStartLine) {
   this._scrollTo(state.startLine)
```

app/web_modules/sourcegraph/blob/Blob.js
ggilmore: Yes, these values didn't seem to be used in the rest of the file.

app/web_modules/sourcegraph/blob/Blob.js
ggilmore: Discussed offline.

app/web_modules/sourcegraph/blob/Blob.js
ggilmore: Discussed offline.

rothfels:
LGTM

PR #240, opened 2016-07-07
all: Call runtime.GOMAXPROCS(0) once on startup.

shurcooL:
Instead of calling it once per request.

Its current performance is not as bad as I might've imagined, but it
still seems scary to call a runtime method that often. Its value
doesn't change mid-process, so might as well store it. This way,
accessing the value does not need any locks.


shurcooL:
Hmm, I just realized this might be invalid if any code actually calls `runtime.GOMAXPROCS` with a > 1 value and *changes* its value, since `conf.GoMaxProcs` would be out of date then.

Perhaps we should be using `runtime.NumCPU` instead?

My goal was to avoid calling sometime from `runtime` on a per-request basis.

shurcooL:
Closing as invalid.

> My goal was to avoid calling sometime from `runtime` package on a per-request basis.

I'll try a different take on that, just use reasonable constants.

cli/internal/middleware/status.go
```diff
@@ -29,7 +29,7 @@ func statusHandler(w http.ResponseWriter, r *http.Request) {
 	fmt.Fprintln(w, "Hostname:   ", hostname)
 	fmt.Fprintln(w, "Git commit: ", envutil.GitCommitID)
 	fmt.Fprintln(w, "Uptime:     ", time.Since(sgxStarted))
-	fmt.Fprintln(w, "GOMAXPROCS: ", runtime.GOMAXPROCS(0))
```
keegancsmith: @shurcooL I'd vote to just remove this from the status handler. It's information we don't actually care about.

shurcooL: If it's not needed, I have no problem with removing it.

Made https://github.com/sourcegraph/sourcegraph/pull/246.

PR #242, opened 2016-07-08
Periodically call debug.freeOSMemory() to prevent accidental OOMKilling

samlavery:
See https://app.asana.com/0/87040567695724/151639107832268 for GC logs from production showing the issue.


rothfels:
I'm still a little shocked that the go gc needs this advisory. Im not the go expert here but this concept looks fine to me.

rothfels:
Add a link to external docs where you found the recommendation to use this method.

samlavery:
The recommended course of action is refactor your codebase to use pools of pre-allocated buffers, but it seems the bulk of the garbage generated in this case comes from json marshalling/unmarshalling, which is stdlib code.  The alternative freeOSMemory call is brought up in the comments of this blog posting: https://blog.cloudflare.com/recycling-memory-buffers-in-go/

slimsag:
I'm not trying to be rude or say that this doesn't mitigate the issue we're currently facing, but I don't think this is a good change and I think it should be reverted.

First, notice that `FreeOSMemory` is defined in package _debug_ (it's not in `runtime` package _for a reason_):

> Package debug contains facilities for programs to debug themselves while they are running.

Second, the docs say:

> FreeOSMemory forces a garbage collection followed by an attempt to return as much memory to the operating system as possible. **(Even if this is not called, the runtime gradually returns memory to the operating system in a background task.)**

You can find where the runtime GC triggers the scavenger [here](https://sourcegraph.com/github.com/golang/go/-/blob/src/runtime/proc.go#L3636-3641) and [here](https://sourcegraph.com/github.com/golang/go/-/blob/src/runtime/proc.go#L3554-3556). In particular, the comment:

> If a heap span goes unused for 5 minutes after a garbage collection,
> we hand it back to the operating system.

And there are serious side effects to running `FreeOSMemory`, too. Because of this change [we'll now trigger a STW (Stop The World) garbage collection](https://sourcegraph.com/github.com/golang/go/-/blob/src/runtime/mheap.go#L953-957) every 60s, and we're also going to lose out on performance where the runtime can reuse memory that it has allocated.. a simplified explanation is that it's a bit like we're restarting our process every 60s with this change from Go's perspective.

Maybe I'm off base here? I don't know. What do others think (@shurcooL, @neelance)? I think we shouldn't be so quick to merge changes like this.

shurcooL:
I agree with @slimsag that we should not merge a change of this magnitude (i.e., effect on the site) this easily, but I think we should be able to experiment with such changes, as long as it's done responsibly.

About the change, I agree this is not something I would easily employ myself, and I agree that even if it works, it probably carries many negative side-effects and is definitely not the root cause of some issues we're seeing.

That said, I can't know the full effects of something like this without actually trying it on a larger scale. So I'm open to trying it, but only if we watch the results closely and evaluate the results after some time.

Ideally, a change like this should be merged when ready, we gather data for some time (some hours or days), then revert it (to prevent momentum effect), then compare without it again, and finally make a decision if it's worth keeping.

neelance:
I fully agree with @slimsag and @shurcooL. This change should be reverted soon.

samlavery:
Yesterday, GRPC pods were being OOMkilled fairly frequently, corrupting data and throwing 500s to the end user.  This was labelled as a P0 bug.  I turned on gctrace=1 yesterday morning and watched what was happening, I suggest tailing a grpc container and piping the output to egrep "scv|gc" yourself to get a feel for how much data is moving.  GODEBUG is still on so the GC stats are still being collected.  What appears to be happening is that given the right set of pre-conditions, the grpc nodes are generating more data than the GC can clear before the kernel steps in and terminates the process.  

I was able to observe the pod getting close to it's 20GB limit while still holding on to roughly 8GB of idle, deallocated RAM.  Prod is an environment where the GC is releasing 6GB+ every 60 seconds.  This sets the stage for something like a big multirepo.importer job to fire off and cause a spike in memory allocations.  Once this happens, the process doesn't have the memory headroom available to both alloc everything for the new workload plus a full mark, sweep, return cycle and thus it gets killed.

The go GC scavenger(routine that does the returning) runs on a constant 2 minute timer, the change I made was to advise the GC to run scavenge every 1 minute.  The idea is that if we aggressively clear old space we won't run out of the headroom needed to respond to sudden extreme demands.  I pushed the change last night, and it effectively stopped pods from being killed until this morning when the vm.overcommit was turned back on.  @shurcooL I have plenty of GC data from before the change, and plenty from over night.  I can revert the change if you like, but we should probably have a plan for how we're going to deal with the original P0.

keegancsmith:
@samlavery Thanks for the explanation. With that I can understand why you did what you did, that makes a lot of sense. Thanks! As for the underlying issue I have some thoughts:

I never thought of the fact that we would corrupt data if a gRPC pod dies, good catch. We should check when we update the version file on the srclib store. If that is done last we should never return corrupted data to the end user (excluding possible weird race conditions on updating that tiny file). If it isn't, it should be an easy fix to prevent corrupted data reaching end users.

As for the 500s end users were seeing. Was that from the corrupted data? Or something else? Do you have an example so we can investigate further? Otherwise gRPC workers should never be involved in a users request other than the fact it puts data into srclibstore and postgres.

neelance:
@samlavery I still don't see how this relates to giving back memory to the system. GC runs often and marks memory as not used any more. Already at this point the memory can be re-used by **the same** Go process. Giving back the memory to the system only makes sense for making it available to **other processes**. Is that your intention?

beyang:
This is clearly a hack meant to mitigate the memory usage issue and should be reverted as soon as possible. Sam is looking into the root cause and will hopefully get to the bottom of that soon. Until then, I'm fine keeping this in place unless we notice perf issues (which so far haven't been an issue). Sam's on SWAT so will be first responder to any memory or perf-related issues over the weekend.

slimsag:
Sounds great to me. Just to be clear, I'm not trying to cause trouble :) I only raised this issue here because the wording (`"We should remove this when Golang's GC gets less agressive about holding on to allocated RAM."`) sounded a bit like this was being committed as a long term fix / considering it as an issue with Go (which I think it isn't). I agree with what others have said here fully, though.

PR #230, opened 2016-07-07
Fixes for issues with Repo.List endpoint.

shurcooL:
This PR includes fixes for https://app.asana.com/0/30847220579580/152010984147276/f and https://app.asana.com/0/87040567695724/152148018628427/f issues.

See individual commit messages for relevant details. **Highly recommended to review commit by commit rather than all at once.**

##### Reviewer tasks

- [ ] HACKS: reviewer approves hacks introduced by this change
- [ ] SECURITY: reviewer approves the security-related changes
- [ ] UNIT-TEST: reviewer approves the unit tests (or the justified omission thereof)
- [ ] E2E-TEST: reviewer approves end-to-end tests (or the justified omission thereof)

##### Test plan

Tested locally during development. Currently pushing to staging4 instance, should be validated there.

PR #244, opened 2016-07-08
services/backend/internal/localstore: Fix List RemoteOnly skipping local repos.

shurcooL:
This is a followup to #230.

Fix issue in the logic of RemoteOnly handling that results in local repos (ones that are present in DB) to be skipped from the list of remote repos.

Populate repo ID (Sourcegraph ID from database) in returned repos even when RemoteOnly if set. This isn't used now, but will be helpful later once we rely more on repo IDs internally.

Add test.

Updates #230.


keegancsmith:
LGTM

PR #245, opened 2016-07-08
query param with better UX

rohanpai:


app/web_modules/sourcegraph/search/SearchSettings.js
```diff
@@ -97,6 +119,7 @@ class SearchSettings extends Container {
 
 		Dispatcher.Stores.dispatch(new UserActions.UpdateSettings(newSettings));
 
+
```
rohanpai: will remove this

app/web_modules/sourcegraph/app/GlobalNav.js
```diff
@@ -221,7 +222,9 @@ class SearchForm extends React.Component {
 
 	_handleSubmit(ev: Event) {
 		ev.preventDefault();
-		this.props.router.push(locationForSearch(this.props.location, this.state.query, true, true));
+		const settings = UserStore.settings.get();
+		const langs = settings && settings.search ? settings.search.languages : null;
```
rothfels: Include the "repos to include" filter in query parameters as well.

app/web_modules/sourcegraph/search/SearchSettings.js
```diff
@@ -35,10 +34,33 @@ class SearchSettings extends Container {
...
+	}
+
+	componentDidMount() {
+		super.componentDidMount();
+		const langFromQuery = this.props.location.query.lang;
+		if (langFromQuery) {
+			const langs = (typeof (langFromQuery) === "string") ? [langFromQuery] : langFromQuery;
+			let validLangParams = [];
+			for (const lang in langs) {
```
rothfels: Instead of lines 49 - 53, you can make `langs` a `let` instead of a `const` and then do the following:

`langs = langs.filter(langIsSupported)`

^^ much nicer :)

app/web_modules/sourcegraph/search/routes.js
```diff
@@ -29,23 +29,32 @@ export function urlToSearch(query?: ?string): string {
 // location and the updatePath/updateQuery parameters.
 //
 // This function modifies loc.
-export function locationForSearch(loc: RouterLocation, query: ?string, updatePath: bool, updateQuery: bool): RouterLocation {
+export function locationForSearch(loc: RouterLocation, query: ?string, lang: ?string[], updatePath: bool, updateQuery: bool): RouterLocation {
```
rothfels: s/lang/langs

app/web_modules/sourcegraph/search/routes.js
```diff
@@ -29,23 +29,32 @@ export function urlToSearch(query?: ?string): string {
...
 	if (updatePath) {
 		loc.pathname = `/${rel.search}`;
 	}
+	let langParam = [];
```
rothfels: I think your flow type annotation makes this irrelevant; langs should always be an array.

app/web_modules/sourcegraph/search/routes.js
rothfels: Maybe would be better (esp. since you'd be adding a third param with the "repos" filter) to just make this one parameter called `searchParams` which is an object w/ `query`, `langs`, and `repos`.

app/web_modules/sourcegraph/search/GlobalSearch.js
```diff
@@ -169,9 +171,9 @@ class GlobalSearch extends Container {
 			} else {
 				state._privateRepos = null;
 			}
+			this.context.router.push(locationForSearch(this.props.location, this.state.query, languages, scope, true, true));
```
rothfels: This should be a replace i think

app/web_modules/sourcegraph/search/routes.js
```diff
@@ -29,23 +29,42 @@ export function urlToSearch(query?: ?string): string {
...
 // This function modifies loc.
-export function locationForSearch(loc: RouterLocation, query: ?string, updatePath: bool, updateQuery: bool): RouterLocation {
+export function locationForSearch(loc: RouterLocation, query: ?string, lang: ?string[], scope: ?Object, updatePath: bool, updateQuery: bool): RouterLocation {
 	if (updatePath) {
 		loc.pathname = `/${rel.search}`;
 	}
+	let langParam = [];
+	if (typeof (lang) === "string") {
+		langParam = [lang];
+	} else if (typeof (lang) === "object" && lang !== null && lang.length > 0) {
```
rothfels: This stuff is too complicated. Why isn't lang always a string array?

app/web_modules/sourcegraph/search/routes.js
```diff
@@ -29,23 +29,42 @@ export function urlToSearch(query?: ?string): string {
...
 
 	if (!updateQuery) {
 		if (query && !loc.state) loc.state = {};
-		if (query) loc.state.q = query;
-		else delete loc.state.q;
+		if (query) {
+			loc.state.q = query;
+			loc.state.lang = langParam;
+			scopeNames.map(function(scopeName) {
+				if (scope[scopeName] === true) loc.state[scopeName] = true;
```
rothfels: Instead of accessing these properties directly on `loc.state`, you should use a `loc.state.scope` object.

app/web_modules/sourcegraph/search/routes.js
```diff
@@ -29,23 +29,42 @@ export function urlToSearch(query?: ?string): string {
...
 	}
 
 	if (query && !loc.query) loc.query = {};
-	if (query) loc.query.q = query;
-	else delete loc.query.q;
+	if (query) {
+		loc.query.q = query;
+		loc.query.lang = langParam;
+		scopeNames.map(function(scopeName) {
+			if (scope[scopeName] === true) loc.query[scopeName] = true;
```
rothfels: I see duplicated code here. Factor.

app/web_modules/sourcegraph/search/SearchSettings.js
```diff
@@ -35,10 +34,38 @@ class SearchSettings extends Container {
...
+		super(props);
+		this.state = {settings: UserStore.settings.get()};
+
+	}
+
+	componentDidMount() {
+		super.componentDidMount();
+		const langFromQuery = this.props.location.query.lang;
+		if (langFromQuery) {
+			const langs = (typeof (langFromQuery) === "string") ? [langFromQuery] : langFromQuery;
```
rothfels: Remove parens around `langFromQuery`

rothfels: You can also remove them around the `typeof` clause

app/web_modules/sourcegraph/search/SearchSettings.js
```diff
@@ -35,10 +34,38 @@ class SearchSettings extends Container {
...
+				...this.state.settings,
+				search: {
+					...this.state.settings.search,
+					languages: validLangParams,
+					scope: scopeParams,
+				},
+			};
+			setTimeout(() => Dispatcher.Stores.dispatch(new UserActions.UpdateSettings(newSettings), 3000));
+		} else {
+			setTimeout(() => Dispatcher.Stores.dispatch(new UserActions.UpdateSettings(this.state.settings), 3000));
```
rothfels: You shouldn't need to do this. Why is this here?

app/web_modules/sourcegraph/search/SearchSettings.js
```diff
@@ -35,10 +34,38 @@ class SearchSettings extends Container {
...
+
+			const newSettings = {
+				...this.state.settings,
+				search: {
+					...this.state.settings.search,
+					languages: validLangParams,
+					scope: scopeParams,
+				},
+			};
+			setTimeout(() => Dispatcher.Stores.dispatch(new UserActions.UpdateSettings(newSettings), 3000));
```
rothfels: Why such a long timeout? You should also leave a helpful comment so the next developer understands why this is here.

app/web_modules/sourcegraph/search/routes.js
```diff
@@ -29,26 +29,42 @@ export function urlToSearch(query?: ?string): string {
...
 
+	if (!lang) lang = [];
+
 	if (!updateQuery) {
+		loc.state = updateScopeAndLanguage(loc.state, scope, lang);
 		if (query && !loc.state) loc.state = {};
-		if (query) loc.state.q = query;
-		else delete loc.state.q;
+		if (query) {
+			loc.state.q = query;
```
rothfels: If you're going to inline the `else` condition, you can also inline this condition.

app/web_modules/sourcegraph/search/routes.js
```diff
@@ -29,26 +29,42 @@ export function urlToSearch(query?: ?string): string {
...
+		} else delete loc.state.q;
 		return loc;
 	}
 
+	loc.query = updateScopeAndLanguage(loc.query, scope, lang);
 	if (query && !loc.query) loc.query = {};
-	if (query) loc.query.q = query;
-	else delete loc.query.q;
+	if (query) {
+		loc.query.q = query;
```
rothfels: Et al.

app/web_modules/sourcegraph/search/routes.js
```diff
@@ -29,26 +29,42 @@ export function urlToSearch(query?: ?string): string {
...
 }
 
+function updateScopeAndLanguage(oldState: any, scope, lang) {
+	const scopeNames = ["public", "private", "popular", "repo"];
+	let state = {...oldState};
+	scopeNames.map((scopeName) => {
+		if (scope && scope[scopeName]) state[scopeName] = true;
+		else delete state[scopeName];
+	});
+	state.lang = lang.slice();
```
rothfels: What is this line for?

app/web_modules/sourcegraph/search/routes.js
```diff
@@ -29,26 +29,42 @@ export function urlToSearch(query?: ?string): string {
...
+	if (query) {
+		loc.query.q = query;
+	} else delete loc.query.q;
 	if (loc.state) delete loc.state.q;
-
 	return loc;
 }
 
+function updateScopeAndLanguage(oldState: any, scope, lang) {
+	const scopeNames = ["public", "private", "popular", "repo"];
```
rothfels: You've hardcoded these here, and also have new code in SearchSettings.js that iteratively looks over these properties. Somebody is bound to update one of these locations but not the other, so you should extract these scope values into a common location and import them.

rothfels: I recommend `search/index.js` where the `SearchScope` type is defined.

app/web_modules/sourcegraph/search/routes.js
rohanpai: @rothfels I need this line since otherwise I get the error `Uncaught TypeError: Cannot assign to read only property '1' of object '[object Array]'`

app/web_modules/sourcegraph/search/SearchSettings.js
```diff
@@ -11,15 +11,15 @@ import Dispatcher from "sourcegraph/Dispatcher";
...
-import {allLangs, langName} from "sourcegraph/Language";
+import {allLangs, langName, langIsSupported} from "sourcegraph/Language";
 import type {LanguageID} from "sourcegraph/Language";
 import {privateGitHubOAuthScopes} from "sourcegraph/util/urlTo";
 import {withUserContext} from "sourcegraph/app/user";
 import LocationStateToggleLink from "sourcegraph/components/LocationStateToggleLink";
 import {LocationStateModal, dismissModal} from "sourcegraph/components/Modal";
 import BetaInterestForm from "sourcegraph/home/BetaInterestForm";
 import * as AnalyticsConstants from "sourcegraph/util/constants/AnalyticsConstants";
+import {SearchScopes} from "sourcegraph/search";
```
rothfels: This should not be capitalized; reserve that for components/classes. Use `searchScopes` camel case.

app/web_modules/sourcegraph/search/SearchSettings.js
```diff
@@ -35,10 +35,35 @@ class SearchSettings extends Container {
...
+
+	componentDidMount() {
+		super.componentDidMount();
+		const langFromQuery = this.props.location.query.lang ? this.props.location.query.lang : [];
+		const query = this.props.location.query;
+		if (this.props.location.pathname === "/search" && (query.lang || query.public || query.private || query.repo || query.popular)) {
+			const langs = typeof langFromQuery === "string" ? [langFromQuery] : langFromQuery;
+			const validLangParams = langs.filter(langIsSupported);
+			let scopeParams = {};
+			SearchScopes.map(function(scopeName) {
```
rothfels: only use `.map` when you care about the return value; else use `.forEach`

Also, if you use the `=>` syntax you can avoid `.bind(this)`:

```
let scopeParams = {};
searchScopes.forEach((scopeName) => scopeParams[scopeName] = this.props.locatino.query[scopeName] === "true");
```

app/web_modules/sourcegraph/search/SearchSettings.js
```diff
@@ -35,10 +35,35 @@ class SearchSettings extends Container {
...
+
+			const newSettings = {
+				...this.state.settings,
+				search: {
+					...this.state.settings.search,
+					languages: validLangParams,
+					scope: scopeParams,
+				},
+			};
+			setTimeout(() => Dispatcher.Stores.dispatch(new UserActions.UpdateSettings(newSettings), 100));
```
rothfels: Are you sure you need a timeout of 100ms even? It may be enough to exclude the second argument entirely (which just guarantees that the callback will be invoked after the component mount cycle completes.

app/web_modules/sourcegraph/search/routes.js
```diff
@@ -29,26 +30,42 @@ export function urlToSearch(query?: ?string): string {
 // location and the updatePath/updateQuery parameters.
 //
 // This function modifies loc.
-export function locationForSearch(loc: RouterLocation, query: ?string, updatePath: bool, updateQuery: bool): RouterLocation {
+// @NOTE loc cannot have immutable properties. lang and scope might be immutable.
+//		 Ensure that if you assign new properties to loc that the properties are mutable
```
rothfels: Remove leading whitespace before Ensure, and end sentences with period.

app/web_modules/sourcegraph/search/routes.js
```diff
@@ -58,3 +75,19 @@ export function queryFromStateOrURL(loc: RouterLocation): ?string {
 	else if (loc.query && loc.query.hasOwnProperty("q")) return firstQueryValue(loc.query.q);
 	return null;
 }
+
+
+export function langFromStateOrURL(loc: RouterLocation): ?string[] {
+	if (loc.state && loc.state.hasOwnProperty("lang")) return typeof loc.state.lang === "string" ? [loc.state.lang] :loc.state.lang;
```
rothfels: why not just `loc.state.lang`?

app/web_modules/sourcegraph/search/routes.js
```diff
@@ -58,3 +75,19 @@ export function queryFromStateOrURL(loc: RouterLocation): ?string {
 	else if (loc.query && loc.query.hasOwnProperty("q")) return firstQueryValue(loc.query.q);
 	return null;
 }
+
+
+export function langFromStateOrURL(loc: RouterLocation): ?string[] {
```
rothfels: Rename `langsFromStateOrURL`

app/web_modules/sourcegraph/search/routes.js
```diff
@@ -58,3 +75,19 @@ export function queryFromStateOrURL(loc: RouterLocation): ?string {
...
+	if (loc.state && loc.state.hasOwnProperty("lang")) return typeof loc.state.lang === "string" ? [loc.state.lang] :loc.state.lang;
+	else if (loc.query && loc.query.hasOwnProperty("lang")) return typeof loc.query.lang === "string" ? [loc.query.lang] : loc.query.lang;
+	return null;
+}
+
+export function scopeFromStateOrURL(loc: RouterLocation): ?Object {
+	let scope = {};
+	searchScopes.forEach((scopeName) => {
+		if (loc.state && loc.state.hasOwnProperty(scopeName)) scope[scopeName] = loc.state.scopeName;
```
rothfels: why not just `loc.state[scopeName]`

rohanpai: loc.state might be undefined. 

Merging after passes CI tests in GH

PR #247, opened 2016-07-08
replace github.com/rogpeppe/rog-go/parallel with github.com/neelance/parallel

neelance:
This avoids messing up CPU and memory profiles by not adding its own stack frame.


keegancsmith:
LGTM

Ship it!!!!

shurcooL:
This is an improvement. LGTM.

It's a shame we couldn't end up just using plain Go code without a library like in the prototype you showed me, I thought that was nicer.

At least with this API you can see the original code doing `go func()` instead of it being hidden in `parallel.Do`.

PR #248, opened 2016-07-08
Small improvements to DefInfo – Part 1

chexee:
- Formats code blocks in docstcings
- Fixes “Sign in” button spacing
- Improved empty state for no refs or examples


beyang:
Screenshot highlighting visual changes would be great!

app/web_modules/sourcegraph/components/styles/_typography.css
```diff
@@ -47,10 +47,9 @@
 	line-height: 1.45;
 }
 
-@value code-normal-line-height 22px; /* used by JavaScript */
 .code-normal {
 	font-size: code-font-size-2;
-	line-height: code-normal-line-height;
+	line-height: 1.45;
```
sqs: @chexee This is used in FileMargin.js to position the def box in the blob view. It needs to be accessible to JavaScript as a pixel value for the computations to work. Check the top of FileMargin.js, where it uses this value.

app/web_modules/sourcegraph/def/index.js
```diff
@@ -94,6 +94,6 @@ export type RefLocationsKey = {
 
 export type ExamplesKey = {
 	repo: string;
-	commitID: string;
-	def: string;
+	commitID: ?string[];
```
sqs: Why an array (`[]`)?

app/web_modules/sourcegraph/def/index.js
```diff
@@ -94,6 +94,6 @@ export type RefLocationsKey = {
...
+	def: ?string[];
```
sqs: Same - why an array?

app/web_modules/sourcegraph/def/index.js
```diff
@@ -85,7 +85,7 @@ export function encodeDefPath(path: string): string {
 
 export type RefLocationsKey = {
 	repo: string;
-	commitID: string;
+	commitID: ?string[];
```
sqs: Same - why an array?

app/web_modules/sourcegraph/components/styles/_typography.css
sqs: If you have a better way to do this, just make sure that the def box is aligned with the selected def on the blob view after your changes.

chexee:
Empty state logged in:
screenshot 2016-07-08 10 56 32

Empty state logged out:
screenshot 2016-07-08 10 56 59

app/web_modules/sourcegraph/components/styles/_typography.css
chexee: Ah, I'll take a look – ideally we can calculate the vertical position of the highlighted line after it's rendered independently of the line height of the page.

app/web_modules/sourcegraph/def/index.js
chexee: Ah, I misunderstood the flow syntax here. It should allow for both `string` and `null`, which I think would actually just be `?string`

sqs:
"joining with GitHub" not Github in the screenshot

app/web_modules/sourcegraph/blob/BlobMain.js
```diff
@@ -180,6 +188,7 @@ export default class BlobMain extends Container {
 					<Blob
 						repo={this.state.repo}
 						rev={this.state.rev}
+						ref={(c) => { this.setState({selectionStartLine: c && c.refs && c.refs.startLineComponent ? c.refs.startLineComponent : null}); }}
```
beyang: Adding parens to make explicit the order of operations here would make this more readable. I'm actually not sure of the order myself. Is it

`(c && c.refs && c.refs.startLineComponent) ? c.refs.startLineComponent : null`

or

`c && c.refs && (c.refs.startLineComponent ? c.refs.startLineComponent: null)`

beyang:
1 minor comment, otherwise LGTM :ship:

app/web_modules/sourcegraph/blob/BlobMain.js
chexee: The first one, so parens sounds like a  great idea :)

PR #250, opened 2016-07-10
Updated header visual style

chexee:
This will eventually be paired with new “integrations” and “repository” designs.  

Things to check:
- [ ] All dropdown menu items work
- [ ] Search works
- [ ] Cursor goes to end if you click on search icon


chexee:
screenshot 2016-07-10 01 25 51
screenshot 2016-07-10 01 26 07

rothfels:
I think this is a net win...but there's some complexity with the results panel overlay which causes a scroll-view-within-a-scroll-view effect: https://cl.ly/1F0E0s2w3g0k

I still think this is better overall than the current UX, which uses a narrower-width panel on wider screens (but that is not clearly separated from the underlying view). I'll keep iterating on this from here. Thanks @chexee

PR #251, opened 2016-07-11
Remove bad metrics from prometheus

keegancsmith:
This is unfortunetly an easy mistake to make since we have no tools to prevent it, nor to warn when it is happening in production (yet). This is currently causing prometheus to fail at collecting important frontend metrics.

### DIFF
--- a/services/httpapi/internal_appdash.go
+++ b/services/httpapi/internal_appdash.go
@@ -5,10 +5,8 @@ import (
 	"net/http"
 	"time"
 
-	"github.com/prometheus/client_golang/prometheus"
 	"sourcegraph.com/sourcegraph/appdash"
 	"sourcegraph.com/sourcegraph/sourcegraph/pkg/httputil/httpctx"
-	"sourcegraph.com/sourcegraph/sourcegraph/pkg/statsutil"
 	"sourcegraph.com/sourcegraph/sourcegraph/pkg/traceutil"
 	"sourcegraph.com/sourcegraph/sourcegraph/pkg/traceutil/appdashctx"
 )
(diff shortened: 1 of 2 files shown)


neelance:
LGTM

services/repoupdater/repo_updater.go
```diff
@@ -41,7 +39,7 @@ func init() {
 // Enqueue queues a mirror repo for refresh. If asUser is not nil, that user's
 // auth token will be used for performing the fetch from the remote host.
 func Enqueue(repo int32, asUser *sourcegraph.UserSpec) {
-	enqueueCounter.WithLabelValues(strconv.Itoa(int(repo))).Inc()
```
keegancsmith: @sqs: This mistake was introduced in the PR which switched us from repo uris to ids. In future large scale refactors we should try and automate as much as possible, then leave the manual bits (which I assume this was) up for review.

services/httpapi/internal_appdash.go
```diff
@@ -57,13 +45,6 @@ func serveInternalAppdashRecordSpan(w http.ResponseWriter, r *http.Request) erro
 		return err
 	}
 
-	// Record page load duration in Prometheus histogram.
-	labels := prometheus.Labels{
-		"name": ev.Name,
-	}
-	elapsed := ev.E.Sub(ev.S)
-	pageLoadDuration.With(labels).Observe(elapsed.Seconds())
```
keegancsmith: @slimsag this may be affecting appdash as well. The `ev.Name` is user controlled, so we have lots of interesting values from hackers trying to find injections into our system.

PR #252, opened 2016-07-11
e2e: Add convenience function Click

keegancsmith:
We add the convenience function so that we can add a work-around on an issue
that affects ChromeDriver. Our assumption is that ChromeDriver will try to click
on an element, but between the time it gets the element and clicks on it, the
element moves. This happens a lot on the Usage Examples page, since we
asynchronously load in some extra annotations that appear above the `View all
references` button. This lead to internal errors from selenium, with selenium
logs like the following:

```
org.openqa.selenium.WebDriverException: unknown error:
Element is not clickable at point (810, 295).
Other element would receive the click: <div class="Blob__margin___2uZsC" style="position: relative;">...</div>
```

More information at https://github.com/seleniumhq/selenium-google-code-issue-archive/issues/2766

### DIFF
--- a/test/e2e/def_flow.go
+++ b/test/e2e/def_flow.go
@@ -19,15 +19,9 @@ func testDefFlow(t *T) error {
 	}
 
 	// Check that the def link appears
-	defLink := t.WaitForElement(selenium.ByLinkText, "View definition", MatchAttribute("href", `/github\.com/golang/go/-/def/GoPackage/net/http/-/Header/Get`))
-	if err = defLink.Click(); err != nil {
-		return err
-	}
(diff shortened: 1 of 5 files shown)


slimsag:
Did you deploy this to prod after you sent me this PR on Slack? If so, it didn't fix the issue it seems (check #e2etest).

Should all tests be using `t.Click` or just ones that encounter this bug? Perhaps we should update the other tests to use this, too?

Aside from those two comments, this LGTM.

keegancsmith:
Not deployed, will do. Will update the rest of the uses. I did test it locally though.

slimsag:
follow-up changes LGTM

PR #253, opened 2016-07-11
UX: Hide search box after result is selected

KattMingMing:
##### Description

Explicitly hides the search results box after a result is selected by setting open to false after `state.nextQuery` is set to null.

This is on staging4


uforic:
Tested on staging4 and in app - LGTM! Didn't see any regression type behavior, tried various exit search scenarios

PR #254, opened 2016-07-11
Analytics: Add Telligent

KattMingMing:
##### Description

Adds Telligent SDK to event logger. Verified on staging4 that this will no longer cause problems and also verified data sent / received that everything is correct.


app/web_modules/sourcegraph/util/EventLogger.js
```diff
@@ -86,6 +91,13 @@ export class EventLogger {
 				}
 			}
 
+			this._telligent("newTracker", "sg", "sourcegraph-logging.telligentdata.com", {
+				appId: "SourcegraphWeb",
```
uforic: Should be be defensive here and wrap this in a try catch?

uforic:
LGTM. Not sure how hardened the telligent JS is, I'd say wrap in try/catch if high liklihood of exception being thrown (not sure how much damage it would cause in the EventLogger initialization), but other than that seems reasonable.

rothfels:
If their SDK throws exceptions it is a bad sign...

KattMingMing:
I haven't seen it throw any exceptions so it might be overkill to wrap it in a try catch.

KattMingMing:
Going to merge this without the try catch

PR #256, opened 2016-07-12
UX: Improved Global Search Experience

KattMingMing:
##### Description

- Added support for SGDesktop. Adds Golang as default search language until new settings have been added (clicking on another language or deselecting the current language) this is to reduce one step to allow a user to search without having to overcome an initial hurdle of selecting a language

- Escape now closes the search box whenever it is pressed

- Default show of SG Desktop now automatically focuses and selects search box

- Global search container no longer consumes the entire screen on small devices 

- Added a "reset" query button when the user is searching inside of the global nav on a smaller device.

##### Testing

Check staging4 to see improvements and to test

### DIFF
--- a/app/web_modules/sourcegraph/app/GlobalNav.js
+++ b/app/web_modules/sourcegraph/app/GlobalNav.js
@@ -212,7 +212,6 @@ class SearchForm extends React.Component {
 
 	_handleGlobalHotkey(ev: KeyboardEvent) {
 		if (ev.keyCode === 27 /* ESC */) {
-			console.log("Escape key pressed");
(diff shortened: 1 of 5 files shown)


uforic:
Verified that Escape now closes search box from wherever.
Verified that default show of SG Desktop now automatically focuses and selects search box.
By 'reset' button, you mean "X"? Vefified this appears on small devices, and works.
Verified on incognito window that if you click on search, it defaults to Golang as default.
Verified that the entire page is no longer taken up on small screens by search results.

uforic:
Looked over code - LGTM. I think this is good to merge.

PR #258, opened 2016-07-12
Fix vote styling for small screens

KattMingMing:
##### Description

Fixes the styling of the vote icons by giving them a width and left aligning them inside of the div when the usage example resizes itself on smaller devices.


slimsag:
@KattMingMing this causes some janky shifting on my MBP retina: gif

uforic:
Merging for now (only viewable by admins for forseeable future, and I want to get the team using it ASAP); also, Chelsea is coming back next week, and will likely apply her designing magic to shift things around a bunch. Will add it as a P2 bug for Chelsea to tackle (and in general, how these buttons might fit in better with a smaller viewing port).

PR #260, opened 2016-07-12
Update signup event action

KattMingMing:


KattMingMing:
Confirmed these events are updated and sending correctly

PR #261, opened 2016-07-12
github: Use an HTTP cache backed by Redis

keegancsmith:
Previously we used an in-memory cache for caching at the HTTP Transport
layer in the Github Client. This lead to large amount of uncached requests
to the Github API on startup (eg when deploying). This started to cause us
to get uncomfortably close to getting rate limited to no requests.

This change switches us from an in-memory cache to Redis. This will also
have the advantage of all instances sharing the same cache, further
reducing the hits we do against the GitHub API. A new type of Redis cache
is introduced called
`rcache.ByteCache` which matches the interface used by `httpcache`.

Note that `httputil.Cache` and `httputil.CachingClient` is currently only
used by GitHub.

Fixes https://app.asana.com/0/87040567695724/154090028181702

### DIFF
--- a/pkg/rcache/rcache.go
+++ b/pkg/rcache/rcache.go
@@ -57,6 +57,21 @@ func redisPool() (*pool.Pool, error) {
 	return connPool_, nil
 }
 
+// getConn returns a redis client from the pool. When you are done you must
+// call the cleanup function to return the connection to the pool.
+func getConn() (*redis.Client, func(), error) {
+	connPool, err := redisPool()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	conn, err := connPool.Get()
+	if err != nil {
+		return nil, nil, err
+	}
+	return conn, func() { connPool.Put(conn) }, nil
+}
+
 // Redis is a cache implemented on top of a Redis client. It is
 // designed to mimick the API of cache.Cache to make it easy to switch
 // instances of cache.Cache to Redis.
@@ -79,16 +94,11 @@ var ErrNotFound = errors.New("Redis key not found")
 func (r *Redis) Get(key string, dst interface{}) error {
 	rkey := fmt.Sprintf("%s:%s:%s", globalPrefix, r.keyPrefix, key)
 
-	connPool, err := redisPool()
+	conn, cleanup, err := getConn()
 	if err != nil {
 		return err
 	}
-
-	conn, err := connPool.Get()
-	if err != nil {
-		return err
-	}
-	defer connPool.Put(conn)
+	defer cleanup()
 
 	resp := conn.Cmd("GET", rkey)
 	if resp.IsType(redis.Nil) {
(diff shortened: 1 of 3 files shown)


neelance:
How does this new cache relate to the existing Redis cache that we had?

neelance:
Why are we using the https://github.com/mediocregopher/radix.v2 project with 157 stars when there is https://github.com/garyburd/redigo with 2,334 stars?

pkg/rcache/rcache.go
```diff
@@ -57,105 +57,181 @@ func redisPool() (*pool.Pool, error) {
...
+	if err != nil {
+		return nil, nil, err
+	}
+	return conn, func() { connPool.Put(conn) }, nil
+}
+
+// Redis exposes methods for speaking to the configured redis server. It
+// serves two main functions: Ensure we namespace keys and correct usage of
+// the connection pool.
 type Redis struct {
```
neelance: The multiple abstractions in this package seem quite bloated to me. I think it might be good to make `Cache` use `ByteCache` instead of `Redis` and then merge `Redis` into `ByteCache` (moving `keyPrefix` into `ByteCache` and inlining `Get`, etc.).

keegancsmith: This is an incremental improvement. Different underlying packages expect different interfaces for the cache (which is why we have `Cache` and `ByteCache`. We can probably have better names), but we now have a consistent way of storing and retrieving the values from Redis with our namespacing rules (the introduced `Redis` struct). I agree we can simplify our different uses of caches, but I'd rather follow up with that. Since that is fixing technical debt, while this is solving a real issue we are facing at the moment.

keegancsmith:
> How does this new cache relate to the existing Redis cache that we had?

This new cache is used by the httpcache library. The older cache worked on the `Repo` struct returned from the API. IE it uses different rules to decide if we can cache something. The `Repo` struct caching forces us to cache Public repos, disregarding the cache control headers from the GitHub API. Depending on the effectiveness on this change, we may be able to drop that cache.

> Why are we using the https://github.com/mediocregopher/radix.v2 project with 157 stars when there is https://github.com/garyburd/redigo with 2,334 stars?

I don't know. This was a decision made in the original implementation of the redis cache. Either way, this library has been in production and working for us well.

keegancsmith:
@neelance I'd really like to get this out. It is an improvement of our current state, and should solve the github pages that are constantly happening. I am very happy to discuss what the future should look like in this area and implement it as a followup.

pkg/rcache/rcache.go
neelance: The changes I suggested above keep both, `Cache` and `ByteCache`, so the outside API should stay the same.

I don't believe that tech debt with be tackled unless we do it *while* solving real issues.

keegancsmith: The contract of `Cache` is different. We silently ignore errors in `ByteCache` and treat it as a cache miss. In `Cache` we do not.

> I don't believe that tech debt with be tackled unless we do it while solving real issues.

Agreed, but at the same time I am tired of opsgenie waking me up at 4am. How about I promise to you I'll follow up straight away on this (ie right after merge), I just want to get this out into production to see if it works.

neelance:
> It is an improvement of our current state, and should solve the github pages that are constantly happening.

I see that, but you can probably do a similar argument for most of our changes. It is not a P0, so I'd rather support a culture of leaving code better than you found it. Still, I can see no obvious bugs, so LGTM if you really want to.

PR #270, opened 2016-07-13
app: link to the line of the ref when viewing an example

sqs:
Previously, the "Used in REPO" link just linked to the repo homepage, which
was not very helpful.

This change makes it so that it now says "From REPO" and it links to the file
and line number that the ref came from. This is not perfect, but it's a big
improvement.


sqs:
@rohanpai We can CR this together.

PR #273, opened 2016-07-13
app: fix repeated WantFile and WantAnnotations actions (second try)

neelance:
Previously the code was iterating over the list of definition results and issuing a WantFile and WantAnnotations for each of them, although they may all be in a few or even a single file. This means that the actions were issued multiple times for the same file. The new implementation iterates over the list of files instead.

Fixes https://app.asana.com/0/87040567695724/153775463609634

### DIFF
--- a/app/web_modules/sourcegraph/def/RefsContainer.js
+++ b/app/web_modules/sourcegraph/def/RefsContainer.js
@@ -184,12 +184,12 @@ export default class RefsContainer extends Container {
 		}
 
 		if (nextState.refs && !nextState.refs.Error && (nextState.refs !== prevState.refs || nextState.shownFiles !== prevState.shownFiles)) {
-			for (let ref of nextState.refs) {
-				let refRev = ref.Repo === nextState.repo ? nextState.commitID : ref.CommitID;
-				if (nextState.shownFiles.has(ref.File)) {
-					Dispatcher.Backends.dispatch(new BlobActions.WantFile(ref.Repo, refRev, ref.File));
-					Dispatcher.Backends.dispatch(new BlobActions.WantAnnotations(ref.Repo, ref.CommitID, ref.File));
-				}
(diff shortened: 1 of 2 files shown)


app/web_modules/sourcegraph/def/RefsContainer.js
```diff
@@ -184,12 +184,12 @@ export default class RefsContainer extends Container {
 		}
 
 		if (nextState.refs && !nextState.refs.Error && (nextState.refs !== prevState.refs || nextState.shownFiles !== prevState.shownFiles)) {
-			for (let ref of nextState.refs) {
-				let refRev = ref.Repo === nextState.repo ? nextState.commitID : ref.CommitID;
-				if (nextState.shownFiles.has(ref.File)) {
-					Dispatcher.Backends.dispatch(new BlobActions.WantFile(ref.Repo, refRev, ref.File));
-					Dispatcher.Backends.dispatch(new BlobActions.WantAnnotations(ref.Repo, ref.CommitID, ref.File));
-				}
+			let firstRef = nextState.refs[0]; // hack: assuming that all refs given to a RefsContainer are from the same repo and rev, thus using the first ref to determine which files we want to show
```
rothfels: Could refs be an empty array? We will throw an exception on the next line if so.

rothfels: I'd hope the parent component is smart enough to exclude that case, but worth checking.

rothfels:
One thing to verify, otherwise LGTM

app/web_modules/sourcegraph/def/RefsContainer.js
neelance: I don't think that a repo can be listed as an example if it has no ref. That wouldn't make any sense.

keegancsmith:
LGTM

PR #274, opened 2016-07-14
e2etest: send failure alerts to OpsGenie + minor fixes

slimsag:
This change effectively sends failure alerts (if any test has failed) top OpsGenie (only one alert will be created).

Additionally, a few minor bug fixes which make working with e2etest locally are also included.

##### Test plan

Tested manually. I will deploy directly to prod once merged.

### DIFF
--- a/test/e2e/e2etest.go
+++ b/test/e2e/e2etest.go
@@ -503,10 +503,16 @@ func (t *testRunner) runTests(logSuccess bool) bool {
 			t.slackLogBuffer.String(),
 		)
 
-		// emit the alert to monitoring-bot
-		err := t.sendAlert()
+		// Create an OpsGenie alert.
+		err := t.alertOpsGenie(fmt.Sprintf("[e2etest]: %v/%v tests failed against %v", failures, total, t.target.String()))
 		if err != nil {
-			t.log.Printf("[WARNING] error while sending alert to monitoring-bot %s", err)
+			t.log.Printf("[WARNING] error while sending alert to opsgenie %s\n", err)
+		}
+
+		// emit the alert to monitoring-bot
+		err = t.sendAlert()
+		if err != nil {
+			t.log.Printf("[WARNING] error while sending alert to monitoring-bot %s\n", err)
 		}
 	}
 	t.slackLogBuffer.Reset()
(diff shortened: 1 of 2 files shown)


test/e2e/e2etest.go
```diff
@@ -153,15 +153,19 @@ func (t *T) WaitForElement(by, value string, filters ...ElementFilter) selenium.
 			if err != nil {
 				return false
 			}
-			t.Logf("WaitForElement: %d matches for (%s, %q)", len(elements), by, value)
+			if *verboseFlag {
+				t.Logf("WaitForElement: %d matches for (%s, %q)", len(elements), by, value)
```
keegancsmith: should we use the same behaviour as go test, where this is only shown if the test fails or `-v` is set?

Also in the `_test.go` wrapper maybe set verboseFlag to the value of testing.Verbose

test/e2e/e2etest.go
```diff
@@ -451,12 +451,18 @@ func (t *testRunner) runTests(logSuccess bool) bool {
 						failuresMu.Unlock()
 					}
 
-					t.log.Printf("[failure] [%v] [%v]: %v\n", test.Name, unitTime, err)
+					if t.slack != nil {
```
keegancsmith: you have `!= nil` here, and the other line `== nil`. Intentional? I assume you just want `== nil`?

test/e2e/e2etest.go
```diff
@@ -503,10 +503,16 @@ func (t *testRunner) runTests(logSuccess bool) bool {
...
 
+		// Create an OpsGenie alert.
+		err := t.alertOpsGenie(fmt.Sprintf("[e2etest]: %v/%v tests failed against %v", failures, total, t.target.String()))
+		if err != nil {
+			t.log.Printf("[WARNING] error while sending alert to opsgenie %s\n", err)
+		}
+
 		// emit the alert to monitoring-bot
-		err := t.sendAlert()
+		err = t.sendAlert()
```
keegancsmith: do we still need the monitoring-bot alert? opsgenie is hooked up to monitoring-bot.

test/e2e/e2etest.go
```diff
@@ -538,6 +544,45 @@ func (t *testRunner) sendAlert() error {
...
+	payload := map[string]string{
+		"apiKey":  key,
+		"alias":   "e2etest", // using the same
+		"message": msg,
+	}
+	data, err := json.Marshal(payload)
+	if err != nil {
+		return err
+	}
+	req, err := http.NewRequest("POST", "https://api.opsgenie.com/v1/json/alert", bytes.NewReader(data))
```
keegancsmith: nit: I doubt this will be an issue, but just to be extra robust we probably want a timeout. In the case of opsgenie acting badly, we will likely trigger some sort of TCP timeout/failure. But it's nice to always set a timeout whenever using an HTTP client when it can block other operations.

keegancsmith:
LGTM. Some inline comments, but nothing that should block you.

test/e2e/e2etest.go
slimsag: > should we use the same behaviour as go test, where this is only shown if the test fails or -v is set?

I think that will require some extra buffering logic here, which isn't too hard but also maybe not _that useful_ for this one log. If we think it is important or helpful though, let's add it as a follow-up (I can do this too, just ping me about it).

slimsag: > Also in the _test.go wrapper maybe set verboseFlag to the value of testing.Verbose

Done.

test/e2e/e2etest.go
slimsag: Yeah, the structure of that code was a bit confusing the way I wrote it. I've clarified it in 7c2468586da340a40b11bd47a26712c70b4397e0

test/e2e/e2etest.go
slimsag: Ahh, I didn't know opsgenie was hooked up to monitoring-bot -- but that makes a lot of sense and simplifies a lot. Removed that code :) thanks for pointing it out

test/e2e/e2etest.go
slimsag: Good point, I guess I didn't think of it because it doesn't _normally_ happen. I'll try to keep this in mind in the future. `http.DefaultClient` should have a Timeout set IMO ¯\_(ツ)_/¯

Done.

PR #275, opened 2016-07-14
fixed rapid click behavior when clicking on top search examples

ggilmore:
Right now in prod, double clicking on any search example causes garbage text to be entered into the search box (although the correct term is still searched). 

You can see an example of this here: https://cl.ly/1J2j3122192v

Now, I've added checks that prevents more than one example from being typed at any given time. 

Example here: https://cl.ly/2b3u1I292v0Z


KattMingMing:
lgtm ++

PR #279, opened 2016-07-15
allow DB connections to stay idle

neelance:
Previously only 2 connections were allowed to stay idle. This caused a lot of connections to be closed and immediately reopened, which is bad for performance.


beyang:
Lgtm

PR #281, opened 2016-07-17
Runtime memory metrics into Prometheus and defaulting FreeOSMemory off

samlavery:
Adding memory and GC metrics to Prometheus and wrapping our FreeOSMemory call
in an environment variable (defaulted to off).  The Go 1.7 GC has much higher throughput,
so we should be able to remove that branch soon.


neelance:
Last week we looked at `go_memstats_heap_alloc_bytes` et al. and now you're adding duplicates of those?

samlavery:
Ah, you're right Richard, I forgot we already had the memory statistics in Prometheus, I couldn't find gc count though.  Thanks.

neelance:
GC count is `go_gc_duration_seconds_count`.

rothfels:
@samlavery can this PR be closed if both of those metrics already exist?

PR #283, opened 2016-07-18
Rig logout so the desktop app can detect it by looking at the URL

mkovacs:
Change the redirect URL after logout from `/` to `/#logout` so the desktop app can detect it.
Also redirect to `/` after login to clear any leftover `#logout` from a previous logout which would look confusing.

Tested locally. Deploying to `staging3`...

### DIFF
--- a/app/web_modules/sourcegraph/user/UserBackend.js
+++ b/app/web_modules/sourcegraph/user/UserBackend.js
@@ -102,7 +102,13 @@ const UserBackend = {
 					.then(checkStatus)
 					.then((resp) => resp.json())
 					.catch((err) => ({Error: err}))
-					.then((data) => Dispatcher.Stores.dispatch(new UserActions.LoginCompleted(data)))
+					.then((data) => {
+						Dispatcher.Stores.dispatch(new UserActions.LoginCompleted(data));
+						// Redirect on login.
+						if (data.Success) {
+							window.location.href = "/";
+						}
+					})
 			);
 			break;
 		case UserActions.SubmitLogout:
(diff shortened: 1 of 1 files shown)


app/web_modules/sourcegraph/user/UserBackend.js
```diff
@@ -102,7 +102,13 @@ const UserBackend = {
 					.then(checkStatus)
 					.then((resp) => resp.json())
 					.catch((err) => ({Error: err}))
-					.then((data) => Dispatcher.Stores.dispatch(new UserActions.LoginCompleted(data)))
+					.then((data) => {
+						Dispatcher.Stores.dispatch(new UserActions.LoginCompleted(data));
+						// Redirect on login.
+						if (data.Success) {
+							window.location.href = "/";
```
slimsag: I think we can redirect to the homepage without a page reload? I'm not 100% sure how this is done, but I'm almost certain we can. @rothfels ?

app/web_modules/sourcegraph/user/UserBackend.js
```diff
@@ -118,7 +124,7 @@ const UserBackend = {
 						Dispatcher.Stores.dispatch(new UserActions.LogoutCompleted(data));
 						// Redirect on logout.
 						if (data.Success) {
-							window.location.href = "/";
+							window.location.href = "/#logout";
```
slimsag: Maybe we should name it something more clear like `#logout-complete` or `#loggedout`? In its current state I would think that the user wants to log out (i.e. some modal should be displayed with a logout button).

slimsag:
Two minor comments, aside from that LGTM

mkovacs:
One more thing: signup seems to be broken on staging3.sourcegraph.com, where this branch is deployed. After clicking "Sign up" on the dialog, nothing happens. Can somebody please verify / offer thoughts what might be causing that?

app/web_modules/sourcegraph/user/UserBackend.js
mkovacs: After a few unsuccessful attempts at changing the URL without a redirect, we've decided to keep our momentum and merge as it is. I've added a task to our backlog to resolve it later.

PR #284, opened 2016-07-18
Language server support

attfarhan:
This change adds support for language servers.  In this change is a JSON-RPC implementation in `pkg/jsonrpc2`, language server protocol types in `pkg/lsp`, and a sample language server and Visual Studio Code extension for using the language server in `lang/server` and `lang/vscode-client` respectively.  

##### Reviewer tasks

### DIFF
--- a/lang/sample/README.md
+++ b/lang/sample/README.md
@@ -1,0 +1,20 @@
+# Running your own language server
+
+This describes how to run your own language server with the Visual Studio Code
+extension in `../vscode-client`.  For instructions on running the sample
+language server, see
+[`../vscode-client/README.md`](../vscode-client/README.md).
+
+To run your own language server:
+
+1.  In the extension code (`../vscode-client/src/extension.ts`), in the
+    `activate` function, change the `run` and `debug` field commands from
+    `sample_server` to `${your_server}`.
+1.  Open ../lang/vscode-client/package.json and under `"activationEvents"` add
+    `"onCommand:${your_language}"`
+1.  Build your server program in the `$PATH` that visual studio sees (e.g.
+    `go build -o /usr/local/bin/${your_server} ${your_server}.go`.)
+1.  Follow the instructions in the
+    [Node.js example language server tutorial](https://code.visualstudio.com/docs/extensions/example-language-server)
+    under "To test the language server do the following:" to build and run the
+    vscode-client sample extension.
--- a/lang/sample/sample_server.go
+++ b/lang/sample/sample_server.go
@@ -1,0 +1,136 @@
+package main
+
+import (
+	"encoding/json"
+	"flag"
+	"fmt"
+	"io"
+	"log"
+	"net"
+	"os"
+
+	"sourcegraph.com/sourcegraph/sourcegraph/pkg/jsonrpc2"
+	"sourcegraph.com/sourcegraph/sourcegraph/pkg/lsp"
+)
+
+var (
+	mode    = flag.String("mode", "stdio", "communication mode (stdio|tcp)")
+	addr    = flag.String("addr", ":2088", "server listen address (tcp)")
+	logfile = flag.String("log", "/tmp/sample_server.log", "write log output to this file (and stderr)")
+)
+
+func main() {
+	flag.Parse()
+	log.SetFlags(0)
+
+	if err := run(); err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
+	}
+}
+
+func run() error {
+	if *logfile != "" {
+		f, err := os.Create(*logfile)
+		if err != nil {
+			return err
+		}
+		defer f.Close()
+		log.SetOutput(io.MultiWriter(os.Stderr, f))
+	}
+
+	h := &jsonrpc2.LoggingHandler{handler{}}
+
+	switch *mode {
+	case "tcp":
+		lis, err := net.Listen("tcp", *addr)
+		if err != nil {
+			return err
+		}
+		defer lis.Close()
+		log.Println("listening on", *addr)
+		return jsonrpc2.Serve(lis, h)
+
+	case "stdio":
+		log.Println("reading on stdin, writing on stdout")
+		jsonrpc2.NewServerConn(os.Stdin, os.Stdout, h)
+		select {}
+
+	default:
+		return fmt.Errorf("invalid mode %q", *mode)
+	}
+}
+
+type handler struct{}
+
+func (handler) Handle(req *jsonrpc2.Request) (resp *jsonrpc2.Response) {
+	if !req.Notification {
+		resp = &jsonrpc2.Response{ID: req.ID}
+	}
+
+	switch req.Method {
+	case "initialize":
+		resp.SetResult(lsp.InitializeResult{
+			Capabilities: lsp.ServerCapabilities{
+				HoverProvider:      true,
+				DefinitionProvider: true,
+				ReferencesProvider: true,
+			},
+		})
+
+	case "shutdown":
+		// Result is undefined, per
+		// https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#shutdown-request.
+		resp.SetResult(true)
+
+	case "textDocument/hover":
+		var params lsp.TextDocumentPositionParams
+		if err := json.Unmarshal(*req.Params, &params); err != nil {
+			resp.Error = &jsonrpc2.Error{Code: 123, Message: "error!"}
+			return
+		}
+
+		pos := params.Position
+		resp.SetResult(lsp.Hover{
+			Contents: []lsp.MarkedString{{Language: "markdown", Value: "Hello over LSP!"}},
+			Range: lsp.Range{
+				Start: lsp.Position{Line: pos.Line, Character: pos.Character - 3},
+				End:   lsp.Position{Line: pos.Line, Character: pos.Character + 3},
+			},
+		})
+	case "textDocument/definition":
+		var params lsp.TextDocumentPositionParams
+		if err := json.Unmarshal(*req.Params, &params); err != nil {
+			resp.Error = &jsonrpc2.Error{Code: 123, Message: "error!"}
+			return
+		}
+		resp.SetResult(lsp.Location{
+			URI: params.TextDocument.URI,
+			Range: lsp.Range{
+				Start: lsp.Position{Line: 0, Character: 0},
+				End:   lsp.Position{Line: 0, Character: 0},
+			},
+		})
+	case "textDocument/references":
+		var params lsp.ReferenceParams
+		if err := json.Unmarshal(*req.Params, &params); err != nil {
+			resp.Error = &jsonrpc2.Error{Code: 123, Message: "error!"}
+			return
+		}
+		resp.SetResult([]lsp.Location{lsp.Location{
+			URI: params.TextDocument.URI,
+			Range: lsp.Range{
+				Start: lsp.Position{Line: 0, Character: 0},
+				End:   lsp.Position{Line: 0, Character: 3},
+			},
+		}, lsp.Location{
+			URI: params.TextDocument.URI,
+			Range: lsp.Range{
+				Start: lsp.Position{Line: 0, Character: 4},
+				End:   lsp.Position{Line: 0, Character: 6},
+			},
+		}})
+	}
+
+	return
+}
--- a/lang/vscode-client/README.md
+++ b/lang/vscode-client/README.md
@@ -1,0 +1,33 @@
+# Running the sample language server via a Visual Studio code extension
+
+This describes how to run the sample language server in `../sample`
+in Visual Studio Code via the Visual Studio Code extension in this directory.
+
+To run the sample language server and extension:
+
+1.  Run `npm install` in this directory (`vscode-client`).
+1.  Open this directory by itself in Visual Studio Code.
+1.  `cd ../sample`. Build the the `sample_server` program and
+    ensure it is in the `$PATH` that Visual Studio Code sees
+    (e.g., `go build -o /usr/local/bin/sample_server sample_server.go`).
+1.  Follow the instructions in the
+    [Node.js example language server tutorial](https://code.visualstudio.com/docs/extensions/example-language-server)
+    under "To test the language server do the following:" to build and
+    run the vscode-client sample extension.
+1.  When the new Visual Studio Code window opens with the extension,
+    create a new text file and type some text. Move your mouse cursor
+    over the text, and you should see "Hello over LSP!" in a tooltip.
+    Right click on any text and click "Go to Definition", and you'll
+    be taken to the beginning of the file. Right click on any text and
+    click "Find All References", and if your text is longer than 6
+    characters, you'll be shown two references in the first line of your
+    file.
+
+Debug tips if extension throws an error:
+
+1.  Open the debug pane (Cmd+Shift+D) to see the call stack.
+1.  Open the debug console (Cmd+Shift+Y) and press the green
+    "play"/triangle button to see the error logs.
+1.  Commonly seen error is the server is not in a `$PATH` that VSCode sees.
+
+To run your own language server with this extension, see [`../sample/README.md`](../sample/README.md).
--- a/lang/vscode-client/package.json
+++ b/lang/vscode-client/package.json
@@ -1,0 +1,31 @@
+{
+	"name": "language-client-example",
+	"description": "VSCode part of a language server",
+	"author": "Microsoft Corporation",
+	"license": "MIT",
+	"version": "0.0.1",
+	"publisher": "vscode",
+	"engines": {
+		"vscode": "^0.10.10"
+	},
+	"categories": [
+		"Other"
+	],
+	"activationEvents": [
+		"onLanguage:plaintext"
+	],
+	"main": "./out/src/extension",
+	"scripts": {
+		"vscode:prepublish": "node ./node_modules/vscode/bin/compile",
+		"compile": "node ./node_modules/vscode/bin/compile -watch -p ./",
+		"postinstall": "node ./node_modules/vscode/bin/install"
+
+	},
+	"devDependencies": {
+		"typescript": "^1.8.9",
+		"vscode": "^0.11.0"
+	},
+	"dependencies": {
+		"vscode-languageclient": "^2.2.1"
+	}
+}
--- a/lang/vscode-client/src/extension.ts
+++ b/lang/vscode-client/src/extension.ts
@@ -1,0 +1,43 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+'use strict';
+
+import * as path from 'path';
+
+import { workspace, Disposable, ExtensionContext } from 'vscode';
+import { LanguageClient, LanguageClientOptions, SettingMonitor, ServerOptions, ErrorAction, CloseAction, TransportKind } from 'vscode-languageclient';
+
+export function activate(context: ExtensionContext) {
+	let serverOptions: ServerOptions = {
+		run : { command: "sample_server" },
+		debug : { command: "sample_server" },
+	}
+
+	// Options to control the language client
+	let clientOptions: LanguageClientOptions = {
+		// Register the server for plain text documents
+		documentSelector: ['plaintext'],
+		synchronize: {
+			// Synchronize the setting section 'languageServerExample' to the server
+			configurationSection: 'languageServerExample',
+		},
+		errorHandler: {
+			error(error: Error, message, count: number): ErrorAction {
+				console.log("ERR", error, message, count);
+				return ErrorAction.Shutdown;
+			},
+    		closed(): CloseAction {
+				return CloseAction.Restart;
+			},
+		},
+	}
+
+	// Create the language client and start the client.
+	let disposable = new LanguageClient('Language Server Example', serverOptions, clientOptions).start();
+
+	// Push the disposable to the context's subscriptions so that the
+	// client can be deactivated on extension deactivation
+	context.subscriptions.push(disposable);
+}
--- a/lang/vscode-client/test/index.ts
+++ b/lang/vscode-client/test/index.ts
@@ -1,0 +1,22 @@
+//
+// PLEASE DO NOT MODIFY / DELETE UNLESS YOU KNOW WHAT YOU ARE DOING
+//
+// This file is providing the test runner to use when running extension tests.
+// By default the test runner in use is Mocha based.
+//
+// You can provide your own test runner if you want to override it by exporting
+// a function run(testRoot: string, clb: (error:Error) => void) that the extension
+// host can call to run the tests. The test runner is expected to use console.log
+// to report the results back to the caller. When the tests are finished, return
+// a possible error to the callback or null if none.
+
+var testRunner = require('vscode/lib/testrunner');
+
+// You can directly control Mocha options by uncommenting the following lines
+// See https://github.com/mochajs/mocha/wiki/Using-mocha-programmatically#set-options for more info
+testRunner.configure({
+	ui: 'tdd', 		// the TDD UI is being used in extension.test.ts (suite, test, etc.)
+	useColors: true // colored output from test results
+});
+
+module.exports = testRunner;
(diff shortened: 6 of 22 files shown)


pkg/lsp/structures.go
```diff
@@ -0,0 +1,157 @@
...
+	 * The text document's URI.
+	 */
+	URI string `json:"uri"`
+}
+
+type TextDocumentItem struct {
+	/**
+	 * The text document's URI.
+	 */
+	URI string `json:"url"`
```
sqs: Incorrect, the struct tag should be uri not url

pkg/lsp/structures.go
```diff
@@ -0,0 +1,157 @@
...
+
+	/**
+	 * The text document's language identifier.
+	 */
+	LanguageID string `json:"languageID"`
```
sqs: should be languageId not languageID: https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#textdocumentitem

sqs:
There are more missing LSP types. Here are some of them.

```go
type DidOpenTextDocumentParams struct {
	TextDocument lsp.TextDocumentItem `json:"textDocument"`
}

type DidChangeTextDocumentParams struct {
	TextDocument   lsp.VersionedTextDocumentIdentifier `json:"textDocument"`
	ContentChanges []TextDocumentContentChangeEvent    `json:"contentChanges"`
}

type TextDocumentContentChangeEvent struct {
	Range       *lsp.Range `json:"range,omitEmpty"`
	RangeLength uint       `json:"rangeLength,omitEmpty"`
	Text        string     `json:"text"`
}

type DidCloseTextDocumentParams struct {
	TextDocument lsp.TextDocumentIdentifier `json:"textDocument"`
}

```

lang/sample/README.md
```diff
@@ -0,0 +1,46 @@
+# sample language server
+
+This language server provides "hover" tooltips, go to definition, and find
+references to Visual Studio Code.
```
sqs: I actually think this doc would be better split into 2: move the `vscode-client`-specific parts to the vscode-client dir and have that be a standalone README for hooking up any lang server. Then have this README be just about this sample server (can be much shorter) and refer to the vscode-client doc.

lang/vscode-client/src/extension.ts
```diff
@@ -0,0 +1,47 @@
...
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+'use strict';
+
+import * as path from 'path';
+
+import { workspace, Disposable, ExtensionContext } from 'vscode';
+import { LanguageClient, LanguageClientOptions, SettingMonitor, ServerOptions, ErrorAction, CloseAction, TransportKind } from 'vscode-languageclient';
+
+export function activate(context: ExtensionContext) {
```
sqs: This file has Windows-style line endings. To be consistent with the rest of the files in our repo, now that we've adapted this vscode-client extension a lot from MSFT's code, you should make them Unix line endings.

pkg/jsonrpc2/jsonrpc2.go
```diff
@@ -0,0 +1,584 @@
...
+	"net"
+	"strconv"
+	"strings"
+	"sync"
+)
+
+const jsonrpcVersion = "2.0"
+
+// Request represents a JSON-RPC request (if ID is set) or
+// notification (if ID is not set). See
```
sqs: docstring needs to be updated now that the Notification field was added (would have needed updating for IDSet as well)

pkg/jsonrpc2/jsonrpc2.go
```diff
@@ -0,0 +1,584 @@
...
+// http://www.jsonrpc.org/specification#request_object and
+// http://www.jsonrpc.org/specification#notification.
+type Request struct {
+	Method       string           `json:"method"`
+	Params       *json.RawMessage `json:"params"` // slice, array, or map[string]T
```
sqs: remove comment - this was only accurate when Params was an `interface{}` type

lang/sample/sample_server.go
```diff
@@ -0,0 +1,137 @@
...
+
+	default:
+		return fmt.Errorf("invalid mode %q", *mode)
+	}
+}
+
+type handler struct{}
+
+func (handler) Handle(req jsonrpc2.Request) (resp *jsonrpc2.Response) {
+	if req.IDSet {
```
sqs: Needs to use Notification not IDSet

pkg/jsonrpc2/jsonrpc2.go
```diff
@@ -0,0 +1,584 @@
...
+func (sc *serverConn) handleNotification(notif Request) {
+	resp := sc.h.Handle(notif)
+	if resp != nil {
+		panic("jsonrpc2: Handle returned non-nil response for notification (must be nil)")
+	}
+}
+
+func (sc *serverConn) doHandleRequests(req ...Request) []Response {
+	var resps []Response
+	if bh, ok := sc.h.(BatchHandler); ok {
```
sqs: @attfarhan I see the problem re: compat with VS Code. We should only use the batch handler if the request was originally a batch.

sqs: @attfarhan Check the f9de3a1504b8e76c5cba3bb94bd9c603f86a61b8 fixup commit I just pushed. This should address the problem. The server will now only respond with a response batch (a JSON array) iff the original request was a batch. This is the correct behavior; the previous behavior was wrong (and VS Code correctly rejected it).

sqs:
@attfarhan Also check ff449cc441736286f2261d46233270e3236596ad, which removes the dep on the external repo (which wasn't added to the govendor dir, so it would fail CI/deploy; and the external repo dep is unnecessary anyway since those types already existed in the pkg/lsp package).

sqs:
@attfarhan Check fixup commit d5f8f6b3199a6280d8bd57c437668e97aee50684, which uses Notification and a one-line struct instantiation instead of IDSet and a more verbose 2-line pattern.

keegancsmith:
@sqs, @attfarhan Are we wanting to review and land this soon?

sqs:
@keegancsmith Yep.

lang/sample/README.md
```diff
@@ -0,0 +1,14 @@
+# Running your own language server
+
+This describes how to run your own language server with the Visual Studio code extension in `../vscode-client`.  For instructions on running the sample language server, see To run your own language server with this extension, see [`../vscode-client/README.md`](https://github.com/sourcegraph/sourcegraph/blob/master/lang/vscode-client/README.md).
+
+To run your own language server:
+
+1. In the extension code (`../vscode-client/extension.ts`), in the      `activate` function, change the `run` and `debug` field commands from `sample_server` to `${your_server}`.
```
sqs: There's a lot of whitespace right before the word "activate"

lang/sample/README.md
```diff
@@ -0,0 +1,14 @@
+# Running your own language server
+
+This describes how to run your own language server with the Visual Studio code extension in `../vscode-client`.  For instructions on running the sample language server, see To run your own language server with this extension, see [`../vscode-client/README.md`](https://github.com/sourcegraph/sourcegraph/blob/master/lang/vscode-client/README.md).
```
sqs: capitalize "code" in "Visual Studio code"

sqs: quote the "To run your own language server with this extension" section, since otherwise it looks like you just capitalized a word in the middle of a sentence

sqs: the URL to the vscode-client/README.md should not point to github and to master. You can use relative paths in Markdown files, like `[link text](../vscode-client/README.md)`.

lang/sample/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+To run your own language server:
+
+1. In the extension code (`../vscode-client/extension.ts`), in the      `activate` function, change the `run` and `debug` field commands from `sample_server` to `${your_server}`.
+1. Open ../lang/vscode-client/package.json and under `"activationEvents"`
+add `"onCommand:${YourLanguage}"`
```
sqs: why `YourLanguage` but `your_server`? be consistent

lang/sample/README.md
sqs: In this Markdown numbered list, either make it so each numbered item does NOT hard-wrap to new lines, or indent the wrapped lines so it is like:

```
1. foo
   bar
```

not

```
1. foo
bar
```

lang/vscode-client/README.md
```diff
@@ -0,0 +1,33 @@
...
+over the text, and you should see "Hello over LSP!" in a tooltip.
+Right click on any text and click "Go to Definition", and you'll
+be taken to the beginning of the file. Right click on any text and
+click "Find All References", and if your text is longer than 6
+characters, you'll be shown two references in the first line of your
+file.
+
+Debug tips if extension throws an error:
+
+1. Open the debug pane (Cmd+shift+D) to see the call stack.
```
sqs: capitalize Shift since you also capitalized Cmd (here and below)

lang/vscode-client/README.md
```diff
@@ -0,0 +1,33 @@
...
+1. Run `npm install` in this directory (`vscode-client`).
+1. Open this directory by itself in Visual Studio Code.
+1. `cd ../sample`. Build the the `sample_server` program and
+ensure it is in the `$PATH` that Visual Studio Code sees
+(e.g., `go build -o /usr/local/bin/sample_server sample_server.go`).
+1. Follow the instructions in the
+[Node.js example language server tutorial](https://code.visualstudio.com/docs/extensions/example-language-server)
+under "To test the language server do the following:" to build and
+run the vscode-client sample extension.
+1. When the new Visual Studio Code window opens with the extension,
```
sqs: Same note as above about not hard-wrapping numbered list items, or indent the lines below in the same list item

lang/vscode-client/package.json
```diff
@@ -0,0 +1,44 @@
...
+	"activationEvents": [
+		"onLanguage:plaintext"
+	],
+	"main": "./out/src/extension",
+	"contributes": {
+		"configuration": {
+			"type": "object",
+			"title": "Example configuration",
+			"properties": {
+				"languageServerExample.maxNumberOfProblems": {
```
sqs: This config is used by their sample language server but not ours, so get rid of it.

lang/vscode-client/src/extension.ts
```diff
@@ -0,0 +1,47 @@
...
+	//let serverProgram = context.asAbsolutePath(path.join('server', 'server.js'));
+	//const args = ["run", "../sample/sample_server.go", "-mode", "stdio"];
```
sqs: Remove commented-out code

lang/vscode-client/src/extension.ts
```diff
@@ -0,0 +1,47 @@
...
+	
+	// Options to control the language client
+	let clientOptions: LanguageClientOptions = {
+		// Register the server for plain text documents
+		documentSelector: ['plaintext'],
+		synchronize: {
+			// Synchronize the setting section 'languageServerExample' to the server
+			configurationSection: 'languageServerExample',
+			// Notify the server about file changes to '.clientrc files contain in the workspace
+			fileEvents: workspace.createFileSystemWatcher('**/.clientrc')
```
sqs: We do not use .clientrc (it was part of their demo tutorial but we don't need it). Remove it.

lang/vscode-client/test/extension.test.ts
```diff
@@ -0,0 +1,22 @@
+// 
+// Note: This example test is leveraging the Mocha test framework.
```
sqs: Remove this file

lang/vscode-client/test/index.ts
```diff
@@ -0,0 +1,22 @@
...
+// By default the test runner in use is Mocha based.
+// 
+// You can provide your own test runner if you want to override it by exporting
+// a function run(testRoot: string, clb: (error:Error) => void) that the extension
+// host can call to run the tests. The test runner is expected to use console.log
+// to report the results back to the caller. When the tests are finished, return
+// a possible error to the callback or null if none.
+
+var testRunner = require('vscode/lib/testrunner');
+
```
sqs: Remove this file

pkg/jsonrpc2/jsonrpc2.go
```diff
@@ -0,0 +1,658 @@
...
+		return err
+	}
+	*r = Request{
+		Method:  m.Method,
+		Params:  m.Params,
+		JSONRPC: m.JSONRPC,
+	}
+	if m.ID != nil {
+		(*r).ID = *m.ID
+		(*r).Notification = false
```
sqs: unnecessary to set this to false. We create the struct right above and the zero (default) value for a bool field is false.

pkg/jsonrpc2/jsonrpc2.go
```diff
@@ -0,0 +1,658 @@
...
+		line, err := r.ReadString('\r')
+		if err != nil {
+			return 0, err
+		}
+		b, err := r.ReadByte()
+		if err != nil {
+			return 0, err
+		}
+		if b != '\n' {
+			return 0, fmt.Errorf(`jsonproto2: line endings must be \r\n`)
```
sqs: Typo, should be "jsonrpc2" not "jsonproto2" (occurs in some other places, too)

attfarhan:
Have updated this to address @sqs comments.  Will wait for @keegancsmith's review before going through with the fixups and merging

pkg/jsonrpc2/jsonrpc2.go
```diff
@@ -0,0 +1,658 @@
...
+	// If the client sent a batch request, and the handler doesn't
+	// also implement BatchHandler, Handle will be called once for
+	// each element of the batch, in the same order the requests
+	// appeared in the batch.
+	Handle(Request) *Response
+}
+
+// BatchHandler is like Handler, but it also operates on entire
+// batches at once.
+type BatchHandler interface {
```
keegancsmith: why don't we require the `Handler` interface to also implement `HandleBatch`? What is the use case of a server we implement only being a `Handler`?

pkg/jsonrpc2/jsonrpc2.go
```diff
@@ -0,0 +1,657 @@
...
+	// Handle handles a request (if !Request.Notification) or notification
+	// (if Request.Notification). If the argument is a request, it must
+	// return a non-nil response. If the argument is a notification,
+	// it must return a nil response.
+	//
+	// If the client sent a batch request, and the handler doesn't
+	// also implement BatchHandler, Handle will be called once for
+	// each element of the batch, in the same order the requests
+	// appeared in the batch.
+	Handle(Request) *Response
```
keegancsmith: inconsistency. We return a pointer, but do not take in a pointer. I'd prefer taking in a pointer.

pkg/jsonrpc2/jsonrpc2.go
```diff
@@ -0,0 +1,657 @@
...
+}
+
+// BatchHandler is like Handler, but it also operates on entire
+// batches at once.
+type BatchHandler interface {
+	Handler
+	HandleBatch([]Request) []Response
```
keegancsmith: inconsistency: We return a *Response in the non-batch interface, but do not for this. Which also means we can't do notifications in Batch.

lang/sample/README.md
```diff
@@ -0,0 +1,16 @@
+# Running your own language server
+
+This describes how to run your own language server with the Visual Studio Code extension in `../vscode-client`.  For instructions on running the sample language server, see [`../vscode-client/README.md`](../vscode-client/README.md).
+
+To run your own language server:
+
+1.  In the extension code (`../vscode-client/extension.ts`), in the
```
keegancsmith: path is missing the `src` component.

pkg/jsonrpc2/jsonrpc2.go
sqs: Why can't we do notifications in batch? We just don't return an element for a notification. So if there were 2 requests but 1 was a notification, HandleBatch would return 1 response.

keegancsmith: Why can't the signature be
```
HandleBatch([]*Request) []*Response
```
then you can have nil responses in the slice.

sqs:
On the pointer-vs.-non-pointer stuff, whatever @keegancsmith thinks is good. I just cleared up one semantic question (does HandleBatch include elements in the return value for notifications? nope - and we should update the docs to explain that).

sqs:
Also, one shortcoming of this jsonrpc2 package is that it doesn't support the server sending notifications or requests (such as https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#publishdiagnostics-notification). We don't need that, but it is something for you guys to keep in the backs of your minds.

pkg/lsp/jsonrpc.go
```diff
@@ -0,0 +1,37 @@
+package lsp
+
+import "fmt"
+
+type ResponseError struct {
+	/**
+	 * A number indicating the error type that occurred.
+	 */
+	Code int
```
slimsag: This package doesn't seem to be auto-generated, but the documentation doesn't appear to be in Go style. Not important short term, but if it's not too hard we should fix.

sqs: @attfarhan copied the docs from https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#publishdiagnostics-notification. I think it's fine to omit the docs and tell people to refer to the exact same fields in that protocol.md doc. That might be easier than reformatting all the doc comments.

pkg/lsp/jsonrpc.go
```diff
@@ -0,0 +1,37 @@
...
+	InvalidRequest             = -32600
+	MethodNotFound             = -32601
+	InvalidParams              = -32602
+	InternalError              = -32603
+	serverErrorStart           = -3209
+	serverErrorEnd             = -32000
+)
+
+func (e *ResponseError) Error() string {
+	return fmt.Sprintf("%+v", e)
```
slimsag:
```
runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow
```

That's gonna call `e.Error()` a lot :)

pkg/jsonrpc2/jsonrpc2.go
sqs: Yep, we can do that, but I'm not sure what that gains. We probably don't want to REQUIRE HandleBatch to return responses in the same order as the requests, since that would require BatchHandlers to do extra work. Allowing emitting nils for notifications (and we then would have to map `[nil, nil, nil, nil, ...]` to no response in the server per the JSON-RPC spec, as opposed to mapping `[]Response(nil)` to no response. (If a request just consists of notifications, then the server should not send anything back, not even an empty batch response.)

Again, I don't care either way, just explaining why it's a certain way. :)

pkg/jsonrpc2/jsonrpc2.go
```diff
@@ -0,0 +1,657 @@
...
+// Client is a JSON-RPC client.
+type Client struct {
+	conn io.ReadWriteCloser
+
+	mu    sync.Mutex
+	resps chan Response // only set if someone is in a RequestAndWaitForResponse call
+}
+
+// NewClient creates a new JSON-RPC client.
+func NewClient(conn io.ReadWriteCloser) *Client {
```
slimsag: Confusing aspect here: Who is responsible for closing the Client connection? Initially because `*Client` has no `Close` method you would assume you have to `Close` it yourself, and in fact that is what the test cases do too. But then why does `NewClient` take an io.ReadWrite**Closer**? Because it is closing it internally [in some situations](https://github.com/sourcegraph/sourcegraph/pull/284/files#diff-8e66a1d3a28d345b10b5fd8d17637ac8R168) ❓ 🎊 

Then a double-Close will end up failing on the user's end.. which is no good.

pkg/lsp/service.go
```diff
@@ -0,0 +1,759 @@
...
+	SignatureHelpRequest(params *TextDocumentPositionParams, result *SignatureHelp) error
+	GoToDefinition(params *TextDocumentPositionParams, result *[]Location) error
+	FindReferences(params *ReferenceParams, result *[]Location) error
+	DocumentHighlights(params *ReferenceParams, result *DocumentHighlight) error
+	DocumentSymbols(params *DocumentSymbolParams, result *[]SymbolInformation) error
+	WorkspaceSymbols(params *WorkspaceSymbolParams, result *[]SymbolInformation) error
+	CodeAction(params *CodeActionParams, result *[]Command) error
+	CodeLensRequest(params *CodeLensParams, result *[]Command) error
+	CodeLensResolve(params *CodeLens, result *CodeLens) error
+	DocumentRangeFormattingParams(params *DocumentFormattingParams, result *[]TextEdit) error
```
slimsag: @attfarhan Is this a typo, or not? Because `DocumentRangeFormatting` is directly below and has almost the same signature!

pkg/lsp/service.go
```diff
@@ -0,0 +1,759 @@
+package lsp
+
+type LangSvc interface {
+	Initialize(params *InitializeParams, result *InitializeResult) error
+	Completion(params *TextDocumentPositionParams, result *CompletionList) error
```
slimsag: Who is going to implement this interface and why is it useful to pass results in a parameter like this? I wanted to try using this for a Client interface to the protocol but it produces not-so-nice type signatures IMO.

Maybe it's because of how we plan to relay it into gRPC?

slimsag: (i.e. why not move the result parameter in all of these signatures into a second return value?)

pkg/lsp/service.go
attfarhan: Yep, good catch

pkg/lsp/service.go
slimsag: I've found out this matches the pattern used by the language server implementations like here from Beyang https://sourcegraph.com/github.com/sourcegraph/tag-server/-/blob/ctags/lsp.go#L20:27-20:45

and we're not using this interface currently; but it does serve as an OK summary / reference to which methods are available generally.

pkg/lsp/service.go
```diff
@@ -0,0 +1,759 @@
+package lsp
+
+type LangSvc interface {
```
beyang: FYI: there is no need to actually use this interface right now. Might be useful if you want to copy and paste the methods into a struct that implements these methods, but nowhere is this actually used.

slimsag:
Pushed a few minor fixes to the client code.

pkg/lsp/service.go
```diff
@@ -2,25 +2,6 @@ package lsp
...
-	DocumentSymbols(params *DocumentSymbolParams, result *[]SymbolInformation) error
-	WorkspaceSymbols(params *WorkspaceSymbolParams, result *[]SymbolInformation) error
-	CodeAction(params *CodeActionParams, result *[]Command) error
-	CodeLensRequest(params *CodeLensParams, result *[]Command) error
-	CodeLensResolve(params *CodeLens, result *CodeLens) error
-	DocumentFormatting(params *DocumentFormattingParams, result *[]TextEdit) error
-	DocumentOnTypeFormatting(params *DocumentFormattingParams, result *[]TextEdit) error
-	Rename(params *RenameParams, result *WorkspaceEdit) error
-}
-
```
slimsag: Note: I removed this for a few reasons:

1. We don't currently use this interface _anywhere_ (not even in `tag-server`).
2. After discussion with @sqs it sounded like these method stubs aren't a 1:1 representation of what needs to be implemented by a server, so looking at an demo server is better anyway.
3. It's likely to become outdated (in fact, consider the fact that there were two duplicate method names accidently added here before and not noticed by anyone).

keegancsmith:
I'm going to squash and merge this in. I've taken a look over it and there are still things that should change, but we can change them as users of the interfaces emerge.

For example, given our use case of LSP I would like us to just support the `HandleBatch` interface. We can't actually do the `Handle` interface, excluding of course us introducing a proxy which manages "sessions".

keegancsmith:
One other thing to potentially follow up on: the client code for jsonrpc seems very complicated.

slimsag:
@keegancsmith thanks! I plan to do a follow-up review and submit PRs as needed.

> One other thing to potentially follow up on: the client code for jsonrpc seems very complicated.

Are you talking about from a user perspective, or the implementation? I just wanted to say I am COMPLETELY aware of the complexity from a user perspective; and I have some cool ideas on how we can fix it. Let's chat if you're interested, otherwise I'll send a PR sometime.

keegancsmith:
@slimsag from both perspectives. IE implementation and API. I assume once it has a simpler API, that will lead to a simpler implementation.

Thanks for following up! I may also follow-up at some point when I run into the pain points.

PR #285, opened 2016-07-18
e2etest: run on CircleCI

slimsag:
- Based against https://github.com/sourcegraph/sourcegraph/pull/276 for simplicity. (That PR must be reviewed first!)
- Run e2etest on CircleCI, in specific run all standard Go/JS tests on the ~7 CircleCI nodes, and dedicate the first CircleCI node to creating a production binary + running e2etest.
- At this time, three major / critical tests do have to be disabled (see commit message for details).
- From what I understand, this will increase our CI times by two factors:
  1. ~30s for installing Selenium via Docker (I have not found a way to cache this) on each node.
  2. Loss of one CircleCI build node (I think we can add one more, for a fee?).

##### Test plan

N/A


dev/circle-build-docker-image.sh
```diff
@@ -3,7 +3,9 @@ set -ex
 
 VERSION=$1
 
-make dist PACKAGEFLAGS="--os=linux $VERSION"
+# HACK: We don't need to do this here because it's already done for every build
+# by circle-ci-run-e2etests.sh
+#make dist PACKAGEFLAGS="--os=linux $VERSION"
```
keegancsmith: I need to read the underlying code (I can't remember what we do), but why is this not just essentially a no-op? ie go is pretty smart at not rebuilding things.

dev/circle-ci-run-e2etests.sh
```diff
@@ -0,0 +1,22 @@
...
+# Note: process is basically guaranteed to start here because it's already been tested
+#       so extensively prior to this script even running, hence we don't do any
+#       complex error checking.
+./src pgsql create -c --db=app
+./src pgsql create -c --db=graph
+./src serve &
+
+# Sleep for a few seconds to give the server process time to start up, or else
+# e2e tests will fail.
+sleep 5
```
keegancsmith: is there not a way we can test if the server is up? generic sleeps are a surefire way to get flakiness.

dev/circle-ci-run-e2etests.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# Grab the network address for 'localhost' because 'localhost' isn't valid from
+# within the Selenium docker container.
+hostip=($(hostname -I | awk '{print $3}'))
+WRITE_SCREENSHOTS=$CIRCLE_ARTIFACTS TARGET=http://$hostip:3080 go test ./test/e2e -v -parallel=50
```
keegancsmith: nice. should we not have it in a subdir within circle_artifacts? Don't bother if it makes code more complicated due to directory creation.

test/e2e/def_flow.go
```diff
@@ -11,6 +15,13 @@ func init() {
 }
 
 func testDefFlow(t *T) error {
+	if ci := os.Getenv("CI"); ci != "" {
+		// TODO(slimsag): Use gRPC to add the repositories that we expect to
+		// have indexed.
+		t.Logf("TODO: def_flow test does not work on CI (expects existing repos)")
```
keegancsmith: can we add a `t.Skip` to our testing interface?

test/e2e/def_flow.go
```diff
@@ -11,6 +15,13 @@ func init() {
 }
 
 func testDefFlow(t *T) error {
+	if ci := os.Getenv("CI"); ci != "" {
+		// TODO(slimsag): Use gRPC to add the repositories that we expect to
```
keegancsmith: this is likely a bit more work than that. they depend on computed data. We probably want some sort of fixture containing a db dump?

keegancsmith:
LGTM. Very cool btw. Do you have any idea on the impact of CI performance? I assume you ran it a few times. How long does the CircleCI node for e2e take to run roughly? How does that compare to the other builds? Maybe push this to a staging node a few times so you get an idea of how long it takes when we have to run all the unittests as well.

dev/ci/install-selenium.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -ex
+
+# TODO(slimsag): If we can find a way to effectively cache this, we can save
+# ~30s off every CircleCI build.
+docker run -d -p 4444:4444 selenium/standalone-chrome
```
keegancsmith: we only need to run this on the first node

PR #288, opened 2016-07-20
Desktop: Autoupdate from GCloud instead of GitHub

KattMingMing:
##### Description

Moves the endpoint and logic to check gcloud instead of GitHub. This URL will also be hosted on the GitHub repo for download access.


uforic:
LGTM, love the hack. Let's merge this first thing in the morning. Thx!

PR #291, opened 2016-07-20
Fixing JSON-RPC implementation

alexsaveliev:
- request/response's `id` attribute may be `string` as specified by http://www.jsonrpc.org/specification#request_object. Java binding of LSP (https://github.com/TypeFox/ls-api/) operates with string `id` as well..


slimsag:
Thanks for this @alexsaveliev -- good to know someone is keeping a close eye 😄 A big 👍 to the first `id` `uint32` -> `string` change, I like the idea of following the spec more here.

> reverted changes made in sourcegraph/sourcegraph@f920eb5 because I think they break communication with JSON-RPC server which expects lowercase attribute names

I think this is not accurate. Because `*Request` implements the [json.Marshaler](https://golang.org/pkg/encoding/json/#Marshaler) and [json.Unmarshaler](https://golang.org/pkg/encoding/json/#Unmarshaler) interfaces, the struct fields are never consulted at all. Their field tags are useless, because the `MarshalJSON` and `UnmarshalJSON` methods of `*Request` will always be invoked to perform the conversion to JSON. And those methods are implemented to use a map of lowercase field names and an anonymous struct with the proper JSON field tags, respectively.

If I'm wrong here please let me know! I just want simpler code 😄 Glad to chat more about it too

pkg/jsonrpc2/jsonrpc2.go
```diff
@@ -25,12 +25,12 @@ const jsonrpcVersion = "2.0"
...
 	Method       string
 	Params       *json.RawMessage
-	ID           uint32
+	ID           string
 	Notification bool
 	JSONRPC      string // "2.0"
 
 	// SPEC NOTE: The spec allows string, number, or null for "id",
-	// but we're pickier. We require numeric IDs.
+	// but we're pickier. We require not null IDs
```
slimsag: This entire comment can be removed; we follow the spec now.

pkg/jsonrpc2/jsonrpc2.go
```diff
@@ -95,7 +95,7 @@ type Response struct {
 	// in the Request object.
 
 	// SPEC NOTE: The spec allows string, number, or null for "id",
-	// but we're pickier.
+	// but we're pickier. We require not null IDs
```
slimsag: Same as above, just remove this comment.

slimsag:
LGTM; thanks for this @alexsaveliev! Let's merge asap :)

/cc @beyang @attfarhan @keegancsmith for those whom this change might require small updates to LSP server implementations.

PR #293, opened 2016-07-20
fix DefInfo not loading

neelance:
For some reason the backend is returning inconsistent data, listing a file as an example but then returning no reference in that file. This change makes the UI not fail in that case.

Fixes https://app.asana.com/0/87040567695724/156775931921647


app/web_modules/sourcegraph/def/RefsContainer.js
```diff
@@ -184,7 +184,7 @@ export default class RefsContainer extends Container {
 			Dispatcher.Backends.dispatch(new DefActions.WantDef(repo, rev, def));
 		}
 
-		if (nextState.refs && !nextState.refs.Error && (nextState.refs !== prevState.refs || nextState.shownFiles !== prevState.shownFiles)) {
+		if (nextState.refs && nextState.refs.length > 0 && !nextState.refs.Error && (nextState.refs !== prevState.refs || nextState.shownFiles !== prevState.shownFiles)) {
```
rothfels: Do we prefer the style `&& nextState.refs.length > 0 &&` or the `&& nextState.refs.length &&`?

app/web_modules/sourcegraph/def/DefBackend.js
```diff
@@ -118,8 +118,10 @@ const DefBackend = {
 							.then(checkStatus)
 							.then((resp) => resp.json())
 							.catch((err) => ({Error: err}))
-							.then(convNotFound)
 							.then((data) => {
+								if (get(data, "Error.response.status") === 404) {
+									data = [];
```
rothfels: :( bummer

rothfels:
Sad to introduce more debt, but fine to merge.

app/web_modules/sourcegraph/def/RefsContainer.js
neelance: I personally prefer to make explicit that it is an integer, not a boolean. Implicit conversions are a source of pain. ;-)

PR #295, opened 2016-07-21
allow global search when filters unchecked

beyang:
This allows global search results to appear when all search filter boxes are deselected. This is not an end state (the end state is to remove the filter boxes entire and have a more seamless search experience), but this will let us test global search performance so we can get a better estimate of how long the entire undertaking will take before taking a big global step in this space.


rothfels:
LGTM

PR #296, opened 2016-07-21
golang: Initial LSP server implementation

keegancsmith:
The implementation is very naive, and only resolves symbols contained in
the current file. However, this is a good starting point for further changes.


keegancsmith:
Plan going forward:
1. Integrate the VFS solution we end up with.
1. Re-use existing tools that answer these questions (godef, guru). Do this by vendoring + modifying for VFS in such a way that we can maintain the vendored files.
1. Support stretch goals (external references, etc)

cc @rothfels @slimsag

lang/golang/hover.go
```diff
@@ -0,0 +1,111 @@
...
+	conf.CreateFromFiles(importPath, f)
+	prog, err := conf.Load()
+	if err != nil {
+		log.Printf("typechecking %s: %s", params.TextDocument.URI, err)
+		if prog == nil {
+			return nil, err
+		}
+	}
+
+	// We already performed the PathEnclosingInterval operation above,
```
sqs: This comment can be removed. It was for when the implementation called `prog.PathEnclosingInterval(...)`.

lang/golang/references.go
```diff
@@ -0,0 +1,139 @@
...
+	conf.CreateFromFiles(importPath, f)
+	prog, err := conf.Load()
+	if err != nil {
+		log.Printf("typechecking %s: %s", params.TextDocument.URI, err)
+		if prog == nil {
+			return nil, err
+		}
+	}
+
+	// We already performed the PathEnclosingInterval operation above,
```
sqs: As mentioned above, this comment can be removed. We no longer call PathEnclosingInterval again.

lang/golang/references.go
```diff
@@ -0,0 +1,139 @@
...
+	// performing it again.
+	pkg := prog.InitialPackages()[0]
+	if pkg == nil {
+		return nil, errors.New("no package found")
+	}
+	o := pkg.ObjectOf(node)
+
+	var locs []lsp.Location
+	for astIdent, obj := range pkg.Info.Uses {
+		if sameObj(o, obj) {
```
sqs: Probably worth adding a comment mentioning that this doesn't yet handle includeDeclaration from https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#find-references.

sqs:
LGTM after a couple of minor comments.

keegancsmith:
@sqs: Thanks for the feedback, I'll fix. Just a note: I'll be ripping all this out tomorrow more than likely and replacing it with something far more robust. Just wanted to get a base into master to iterate on.

PR #298, opened 2016-07-22
reenable defs2 ref_ct update

beyang:
The update of `defs2.ref_ct` has been disabled since July 1st because it was leading to deadlocks in the database that resulted in p0 performance degradation of search. Without the ref count update, search rankings are based on text similarity only, which means oftentimes results from popular libraries don't show up in results. We've worked around this with the search filter buttons, but these are cumbersome to use and severely limit the power and usefulness of search. This change re-enables the ref count update while ensuring that the row-level update locks are acquired in a consistent order to prevent deadlocks.

For explanation of why this works, see:
- https://www.postgresql.org/docs/current/static/sql-select.html#SQL-FOR-UPDATE-SHARE
- http://dba.stackexchange.com/questions/68388/optimizing-concurrent-updates-in-postgres


keegancsmith:
@beyang LGTM, but monitor https://grafana.sgdev.org/dashboard/db/build-metrics
Currently the async queue size as well as number of errors is always close to zero. IE the async queue works very well at the moment. I'm concerned this change might either make updates a lot slower (so the queue size grows), or introduce errors (from lock acquisition). But best way to find out is in prod :P

FYI I can finish off a change where we ensure only one update to a repo happens at a time. I'll timebox 15min to see if I can finish it off. That should mean we can remove the concurrent lock hacks.

PR #300, opened 2016-07-22
golang: Use godef

keegancsmith:
See the `definition-*` test cases in testdata. I have added tests jumping to definitions in the same file, golang and external to workspace. Currently we skip the external to workspace since godef fails on them. We may actually want to not use godef, and instead write our own custom implementation which does the "right thing" for external workspaces. What the "right thing" is, is still up for debate.

In my test data I have made the "right thing" return a URI like `golang://github.com/prometheus/client_golang/prometheus.CounterOpts`, and no range. I am suggesting our LSP proxy implementation then does a "where is symbol defined" sort of query against the external workspace. But how we deal with that case is still up for debate.


keegancsmith:
@neelance See PR description. I have mentioned this issue randomly on quip and slack. Not sure what others think about a solution.

sqs:
BTW, check godefinfo (github.com/sqs/godefinfo) for something that prints out this kind of Go URI that uses go/types and not godef (which is not based on the latest go/types typechecker). That could be repurposed to be the custom tool you refer to.

sqs:
Anything in particular you want a review of? @keegancsmith

keegancsmith:
@sqs Nothing in particular, just looking for an LGTM.

lang/golang/testdata/mux/mux.go
```diff
@@ -0,0 +1,495 @@
...
+	return contextSet(r, varsKey, val)
+}
+
+func setCurrentRoute(r *http.Request, val interface{}) *http.Request {
+	return contextSet(r, routeKey, val)
+}
+
+// ----------------------------------------------------------------------------
+// Helpers
+// ----------------------------------------------------------------------------
```
slimsag: nit: I'm not a fan of these sections within Go documents, personally, I might suggest moving them to separate smaller files? `helpers.go`, etc.

slimsag:
It looks like `TestFixtures` is failing on CI.

I see some minor issues with how we're creating `workspace://` URIs which we'll need to resolve eventually.. but I don't have concrete enough details to comment yet. Let's merge first and address as a follow-up.

LGTM

lang/golang/testdata/mux/mux.go
keegancsmith: this is testdata, taken from the gorilla/mux repo.

keegancsmith:
whoops, will fix CI failure now :)

PR #301, opened 2016-07-22
hover tooltips backed by on-demand request to backend

neelance:
The old srclib based annotations fetched for the whole document are still there, but the hover tooltip is not using them any more. Instead when you hover some identifier, the `hover-info` endpoint is used to map a position of the document to the definition information shown in the tooltip. This is a first step towards our new language server approach. The `hover-info` endpoint is currently still backed by the srclib data, but it can be easily changed to do a request to a language server.
Known regression: No highlighting of the identifier/word in other lines when you hover it.


app/web_modules/sourcegraph/def/DefActions.js
```diff
@@ -104,15 +104,31 @@ export class SelectDef {
...
+	pos: ?BlobPos;
+	info: string;
 
-	constructor(url: ?string, language: ?string) {
-		this.url = url;
-		this.language = language;
-		this.eventName = "HighlightDef";
+	constructor(pos: ?BlobPos, info: string) {
+		this.pos = pos;
+		this.info = info;
```
slimsag: I see hover info in this context is just a `string` -- did we want to handle the fact that hover info is actually a list of [language, value] pairs on the backend?

app/web_modules/sourcegraph/def/DefBackend.js
```diff
@@ -128,6 +128,27 @@ const DefBackend = {
...
+				if (info === null) {
+					let url = `/.api/repos/${action.pos.repo}@${action.pos.commit}/-/hover-info?file=${action.pos.file}&line=${action.pos.line}&character=${action.pos.character}`;
+					trackPromise(
+						DefBackend.fetch(url)
+							.then(checkStatus)
+							.then((resp) => resp.json())
+							.catch((err) => ({Error: err}))
+							.then((data) => {
+								if (get(data, "Error.response.status") === 404) {
+									return; // TODO we may want to indicate error in the UI
```
slimsag: Can we at least `console.log` the error here? Or better maybe `console.error` since that would be caught by Sentry?

app/web_modules/sourcegraph/def/DefTooltip.js
```diff
@@ -1,7 +1,12 @@
+// @flow weak
```
slimsag: 👏

app/web_modules/sourcegraph/def/DefActions.js
neelance: Oops, that type annotation is not correct. Right now it is a `{def: <sourcegraph.Def>}`. Changing that to the pairs is a subsequent step.

app/web_modules/sourcegraph/def/DefTooltip.js
```diff
@@ -47,9 +63,19 @@ class DefTooltip extends Component {
...
 
 	render() {
-		let def = this.state.def;
+		if (this.state.hoverPos === null) {
+			return null;
+		}
+
+		let info = this.state.hoverInfos.get(this.state.hoverPos);
+		if (info === null) {
+			// TODO show loading indicator
```
slimsag: What affect does this have in practice? The tooltip just doesn't show? Or it shows the tooltip box with no text?

app/web_modules/sourcegraph/util/EventLogger.js
```diff
@@ -304,17 +304,6 @@ export class EventLogger {
 			}
 			break;
 
-		case DefActions.HighlightDef:
```
slimsag: Do we still get analytics events for hovering elsewhere? Might be important, not sure.

services/backend/annotations.go
```diff
@@ -204,3 +208,83 @@ func computeLineStartBytes(data []byte) []uint32 {
...
 	return pos
 }
+
+func (s *annotations) GetDefAtPos(ctx context.Context, opt *sourcegraph.AnnotationsGetDefAtPosOptions) (*sourcegraph.DefSpec, error) {
+	if err := accesscontrol.VerifyUserHasReadAccess(ctx, "Annotations.GetRef", opt.Entry.RepoRev.Repo); err != nil {
+		return nil, err
+	}
+
+	if opt.Entry.Path == "" {
+		return nil, fmt.Errorf("GetRef: no file path specified for file in %v", opt.Entry.RepoRev)
```
slimsag: s/`GetRef:`/`GetDefAtPos`/ ?

app/web_modules/sourcegraph/def/DefBackend.js
neelance: Will do.

services/backend/annotations.go
```diff
@@ -204,3 +208,83 @@ func computeLineStartBytes(data []byte) []uint32 {
 	}
 	return pos
 }
+
+func (s *annotations) GetDefAtPos(ctx context.Context, opt *sourcegraph.AnnotationsGetDefAtPosOptions) (*sourcegraph.DefSpec, error) {
+	if err := accesscontrol.VerifyUserHasReadAccess(ctx, "Annotations.GetRef", opt.Entry.RepoRev.Repo); err != nil {
```
slimsag: Same here, s/`GetRef`/`GetDefAtPos`/

app/web_modules/sourcegraph/def/DefTooltip.js
neelance: It does not show. The returned `null` is not rendered at all.

app/web_modules/sourcegraph/util/EventLogger.js
neelance: I'll check that.

services/backend/annotations.go
```diff
@@ -204,3 +208,83 @@ func computeLineStartBytes(data []byte) []uint32 {
...
+	offset += opt.Character
+
+	filters := []srcstore.RefFilter{
+		srcstore.ByRepos(repo.URI),
+		srcstore.ByCommitIDs(opt.Entry.RepoRev.CommitID),
+		srcstore.ByFiles(true, opt.Entry.Path),
+		srcstore.RefFilterFunc(func(ref *graph.Ref) bool {
+			return ref.Start <= offset && ref.End > offset
+		}),
+	}
```
slimsag: 😢 I think these filters might panic if any of the input strings are empty(from what I remember). Maybe be careful since the input could be an empty string? If that's not the case then ignore.

app/web_modules/sourcegraph/def/DefBackend.js
neelance: The JS console will already show an error.

services/backend/annotations.go
neelance: :+1:

services/backend/annotations.go
```diff
@@ -204,3 +208,83 @@ func computeLineStartBytes(data []byte) []uint32 {
...
+		})
+		if err != nil {
+			return nil, err
+		}
+		defCommitID = defaultRev.CommitID
+	}
+
+	return &sourcegraph.DefSpec{
+		Repo:     defRepo.ID,
+		CommitID: defCommitID, // TODO this is only correct for local defs
```
slimsag: Does this TODO mean that hovering over a def which comes from another repository won't work? If so that seems like a second regression?

services/backend/annotations.go
neelance: I think there is no way to have an empty value here.

services/backend/annotations.go
neelance: Oops, scratch that. I already fixed it.

services/httpapi/hover_info.go
```diff
@@ -0,0 +1,60 @@
...
+	})
+	if err != nil {
+		return err
+	}
+
+	return writeJSON(w, struct {
+		Def *sourcegraph.Def `json:"def"`
+	}{
+		Def: def,
+	})
```
slimsag: Is the wrapping struct useful for extending in the future? Or could we just return a `*sourcegraph.Def` directly?

slimsag:
LGTM

slimsag:
I guess maybe double-check with someone else about the regression just to be sure it's OK? Not sure. I think it's fine though.

services/httpapi/hover_info.go
neelance: We could, but I wanted to keep this a bit separated. We're returning a "hover info" here, which just right now is equal to `*sourcegraph.Def`. Will probably change in the future.

neelance:
Already checked with John.

neelance:
Addressed all comments. :-)

PR #302, opened 2016-07-22
{lang/golang, pkg/lputil}: add langprocessor-go which implements the intermediate REST protocol

slimsag:
This change adds a langprocessor-go binary which implements the intermediate Language Processor REST protocol (via the new `pkg/lputil`). It simply translates from LSP into the REST protocol in an effort to serve hover requests.

It can be tested via:

```
curl -H "Content-Type: application/json" -X POST -d '{"Repo": "github.com/gorilla/mux", "Commit": "d391bea3118c9fc17a88d62c9189bb791255e0ef", "File": "mux.go", "Line": 17, "Character": 5}' http://localhost:4141/hover
```

After starting it:

```
go install ./lang/golang/cmd/langprocessor-go/ && langprocessor-go
```

And the LSP server which it connects to:

```
go install ./lang/golang/cmd/langserver-go/ && langserver-go
```

Which ultimately produces the response:

```
{"Contents":[{"Type":"","Value":""},{"Type":"go","Value":"func github.com/gorilla/mux.NewRouter() *github.com/gorilla/mux.Router"}]}
```

### DIFF
--- a/lang/golang/cmd/langprocessor-go/Dockerfile
+++ b/lang/golang/cmd/langprocessor-go/Dockerfile
@@ -1,0 +1,5 @@
+FROM alpine:3.4
+
+ADD langprocessor-go /langprocessor-go
+
+ENTRYPOINT ["/langprocessor-go"]
--- a/lang/golang/cmd/langprocessor-go/langprocessor-go.go
+++ b/lang/golang/cmd/langprocessor-go/langprocessor-go.go
@@ -1,0 +1,32 @@
+package main
+
+import (
+	"flag"
+	"log"
+	"net/http"
+	"os"
+	"path/filepath"
+
+	"sourcegraph.com/sourcegraph/sourcegraph/pkg/lputil"
+)
+
+var (
+	httpAddr = flag.String("http", ":4141", "HTTP address to listen on")
+	lspAddr  = flag.String("lsp", ":2088", "LSP server address")
+)
+
+func main() {
+	flag.Parse()
+
+	gopath := os.Getenv("GOPATH")
+	rootPath := func(repo, commit string) string {
+		return filepath.Join(gopath, "src", repo)
+	}
+
+	log.Println("Translating HTTP", *httpAddr, "to LSP", *lspAddr)
+	http.Handle("/", &lputil.Translator{
+		Addr:     *lspAddr,
+		RootPath: rootPath,
+	})
+	http.ListenAndServe(*httpAddr, nil)
+}
--- a/pkg/lputil/lputil.go
+++ b/pkg/lputil/lputil.go
@@ -1,0 +1,141 @@
+// Package lputil implements Language Processor utilities.
+package lputil
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+	"net"
+	"net/http"
+
+	"sourcegraph.com/sourcegraph/sourcegraph/pkg/jsonrpc2"
+	"sourcegraph.com/sourcegraph/sourcegraph/pkg/lsp"
+)
+
+// Translator is an HTTP handler which translates from the Language Processor
+// REST API (defined in proto.go) directly to Sourcegraph LSP batch requests.
+type Translator struct {
+	// Addr is the address of the LSP server which translation should occur
+	// against.
+	Addr string
+
+	// RootPath is invoked to determine the workspace directly at which all
+	// file requests are relative to.
+	RootPath func(repo, commit string) string
+}
+
+// ServeHTTP implements the http.Handler interface.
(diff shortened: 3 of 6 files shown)


pkg/lputil/lputil.go
```diff
@@ -0,0 +1,144 @@
...
+		})
+		if err2 != nil {
+			// TODO: configurable logging
+			log.Println(err2)
+		}
+	}
+}
+
+func (t *Translator) serveHover(w http.ResponseWriter, r *http.Request) error {
+	if r.Method != "POST" || r.URL.Path != "/hover" || len(r.URL.Query()) > 0 {
```
neelance: The hover request is not modifying anything. That would be `GET` for me.

pkg/lputil/lputil.go
```diff
@@ -0,0 +1,144 @@
...
+		w.WriteHeader(http.StatusNotFound)
+		return nil
+	}
+
+	// TODO(slimsag): We don't need to create a new JSON RPC 2 connection every
+	// time, but we will need a connection pool and reconnection logic to avoid
+	// it.
+	conn, err := net.Dial("tcp", t.Addr)
```
neelance: I think we can assume exactly 1 language server per language processor. They will be colocated in a pod. It would make sense to connect on startup and then reuse that single connection, since request concurrency is allowed.

pkg/lputil/lputil.go
slimsag: Right, I thought that too. But it does mean that we can't read from `r.Body` for the parameters (at least, not according to HTTP spec). And then we need to pass repo/commit/file/line/character parameters all via the URL query. I thought keeping it in the body as a JSON object would be quite nice?

slimsag: I'm of course happy to change it if you feel strongly^^

pkg/lputil/lputil.go
slimsag: Yes, agreed. Though that is not why there is one connection per request here. It is because `jsonrpc2.NewClient` isn't thread safe due to the way it tries to map request IDs to responses poorly.

pkg/lputil/lputil.go
neelance: Makes sense. Let's keep it a `POST` then, I care more about having a shared Go type that we simply serialize.

pkg/lputil/lputil.go
slimsag: (i.e. request concurrency is handled by LSP servers according to spec... but our client is dumb and doesn't handle concurrency, so we need a pool of them)

neelance: Okay. For now, let's maybe change the comment to say exactly that?

slimsag: Done in 156a5e0516958932b0586724c8b072c51f43b217

lang/golang/cmd/langprocessor-go/langprocessor-go.go
```diff
@@ -0,0 +1,32 @@
...
+	lspAddr  = flag.String("lsp", ":2088", "LSP server address")
+)
+
+func main() {
+	flag.Parse()
+
+	gopath := os.Getenv("GOPATH")
+	rootPath := func(repo, commit string) string {
+		return filepath.Join(gopath, "src", repo)
+	}
```
slimsag: @rothfels @samlavery FYI this is where we will have to pick where we are serving workspaces from (this is the language processor sitting in front of the LSP server). For the initial deployment I'm sure `$GOPATH` won't be set up, so maybe we need to hard-code something here temporarily for today/etc.

pkg/lputil/lputil.go
slimsag: Agreed. Maybe you would be a good person to look over the REST spec I defined in https://sourcegraph.quip.com/OEKWAMvbfvF6 as well and tell me if it's sane? since I defined everything there as POST for this exact reason.

slimsag:
@neelance TODO comment updated per our in-person discussion; and build script added (and image built/pushed) based on your commit https://github.com/sourcegraph/sourcegraph/commit/e7a432fcd9ecf2720e621c749bf2208578329f74

lang/golang/cmd/langprocessor-go/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM alpine:3.4
+
+ADD langprocessor-go /langprocessor-go
+
+ENTRYPOINT ["/langserver-go"]
```
neelance: oO

neelance:
LGTM

slimsag:
Squashed Dockerfile ENTRYPOINT commit; merging.

PR #304, opened 2016-07-25
grpc: Log unsanitized error

keegancsmith:
Viewing logs at the moment is not very useful since nearly everything is
sanitized. This modifies are behaviour such that we still return sanitized
errors, but we log the real error to aid debugging.


beyang:
LGTM

PR #305, opened 2016-07-25
golang: Add debug HTTP endpoints to cmds

keegancsmith:
Factored out the code we use in `src serve` so we can re-use it.

### DIFF
--- a/lang/golang/cmd/langprocessor-go/langprocessor-go.go
+++ b/lang/golang/cmd/langprocessor-go/langprocessor-go.go
@@ -7,16 +7,22 @@ import (
 	"os"
 	"path/filepath"
 
+	"sourcegraph.com/sourcegraph/sourcegraph/pkg/debugserver"
 	"sourcegraph.com/sourcegraph/sourcegraph/pkg/lputil"
 )
 
 var (
 	httpAddr = flag.String("http", ":4141", "HTTP address to listen on")
 	lspAddr  = flag.String("lsp", ":2088", "LSP server address")
+	profbind = flag.String("prof-http", "", "net/http/pprof http bind address")
 )
 
 func main() {
 	flag.Parse()
+
+	if *profbind != "" {
+		go debugserver.Start(*profbind)
+	}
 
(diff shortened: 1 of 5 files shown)


pkg/debugserver/debug.go
```diff
@@ -29,20 +23,18 @@ func Start(addr string) {
...
-		pp.Handle("/metrics", prometheus.Handler())
-		log15.Warn("could not start debug HTTP server:", "error", http.ListenAndServe(addr, pp))
-	}()
-	log15.Debug("Profiler available", "on", fmt.Sprintf("%s/pprof", addr))
+	})
+	pp.Handle("/", index)
+	pp.Handle("/debug", index)
+	pp.Handle("/vars", http.HandlerFunc(expvarutil.ExpvarHandler))
+	pp.Handle("/gc", http.HandlerFunc(expvarutil.GCHandler))
+	pp.Handle("/freeosmemory", http.HandlerFunc(expvarutil.FreeOSMemoryHandler))
```
slimsag: Didn't know we had this.. cool!

slimsag:
LGTM but why not have a default set for always-on debug servers? It seems like it will always be useful

lang/golang/cmd/langprocessor-go/langprocessor-go.go
```diff
@@ -7,17 +7,23 @@ import (
...
 	"path/filepath"
 
+	"sourcegraph.com/sourcegraph/sourcegraph/pkg/debugserver"
 	"sourcegraph.com/sourcegraph/sourcegraph/pkg/lputil"
 )
 
 var (
 	httpAddr = flag.String("http", ":4141", "HTTP address to listen on")
 	lspAddr  = flag.String("lsp", ":2088", "LSP server address")
+	profbind = flag.String("prof-http", "", "net/http/pprof http bind address")
```
slimsag: e.g. maybe make the default for here and the other program `":6060"` ?

slimsag:
@keegancsmith I am going to merge and address my one comment, if you disagree let's do a follow-up or revert.

